PR #12, opened 2015-01-22
Add headless build env var to servo-linux-2 builder.

glennw:


glennw:
I haven't tested this, but it seems a simple enough change. Is it possible to test buildbot changes before pushing them?

Ideally this could be done simply by passing --features=headless to the ./mach build command. However, this is a chicken and egg problem, since cargo will then fail until it gets the PR which includes the new headless feature.

Doing it this way should mean we can update saltfs, and then add a PR which has the headless feature, and reads the env var inside mach to set up the headless build.

Once this all settles, we could change to using the method above if we wanted to.

r? @larsbergstrom

PR #27, opened 2015-04-16
Stop running content tests.

Ms2ger:
They have all been moved into test-wpt in
<https://github.com/servo/servo/pull/5700>.


saneyuki:
r+

PR #28, opened 2015-05-06
Add --dev flag to mach build commands

mbrubeck:
This is a no-op right now, but it will be necessary if servo/servo#5964 lands. r? @larsbergstrom

### DIFF
--- a/buildbot/master/master.cfg
+++ b/buildbot/master/master.cfg
@@ -115,7 +115,7 @@ mac_test_env = dict({
 linux1_factory = BuildFactory()
 linux1_factory.addStep(Git(repourl=SERVO_REPO, mode="full", method="clobber"))
 linux1_factory.addStep(ShellCommand(command=["./mach", "test-tidy"], env=linux_headless_env))
-linux1_factory.addStep(Compile(command=["./mach", "build"], env=linux_headless_env))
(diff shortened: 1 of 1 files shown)


Larsberg:
I think you've got the wrong guy, maybe you want @larsbergstrom https://github.com/larsbergstrom?

larsbergstrom:
Hrm, this brings up the fact that we never *did* turn on the release builders. Maybe we should be building all of these with `--release` instead? I'm a little scared of the fact that we're not testing the build we should be shipping.

mbrubeck:
Oops, sorry @Larsberg!  Feel free to unsubscribe from this issue.  @larsbergstrom uses that username on some other forums, which is why I got mixed up.

@larsbergstrom: Yes, we could change at least one of these to `--release`, if we have the time or capacity to add separate release builders. But we'd lose coverage of `debug_assert` and overflow checking during any tests that are no longer run on debug builds.

Note: I updated this PR to use `--dev` instead of `--debug`, to reflect changes in #5964.  But this means it's no longer a no-op on Servo master, so we shouldn't deploy this yet.

mbrubeck:
This is ready to merge now that servo/servo#6023 has landed.

PR #35, opened 2015-06-02
Remove bors

Manishearth:
We've been running that cronjob all this while.

Let's wait for the current salt changes (csstests) to simmer before landing this though

r? @larsbergstrom


metajack:
r=me when you want to land this.

PR #38, opened 2015-06-09
Manage homu service with Salt.

metajack:
r? @edunham @larsbergstrom


edunham:
r+ Nothing here sticks out at me as wrong, although I'm still leveling up my Salt skills.

larsbergstrom:
Where will the logs go when we switch to this method of running homu? I'm really excited to get us off of the "run it in a screen session in root" but want to make sure we can still see the output easily.

metajack:
Everything upstart manages gets logged in /var/log/upststart/$name.log and rotated.

PR #39, opened 2015-06-09
Minor fixes for homu service.

metajack:


edunham:
r+, and would you be interested in something like https://github.com/saltstack/salt/issues/802#issuecomment-33642334 to catch typos like that colon on future PRs?

metajack:
Indeed I am interested in that!

PR #40, opened 2015-06-09
Add html5ever repo support to homu.

metajack:
To enable repos you have to:

1) Add them in the homu/cfg.toml by adding to the list as ('$owner', '$repo')
2) Add a webhook for http://build.servo.org:54856/github to the repo webhooks for push,pull request, and issue comments with the secret key (see private data pillar for the key)
3) Add a webhook in `.travis.yml` in the repo pointing at http://build.servo.org:54856/travis

Additionally you may need to reset the travis token and associated user in the travis service in the github admin panel for the repo. (use bors-servo as the user).

### DIFF
--- a/homu/cfg.toml
+++ b/homu/cfg.toml
@@ -11,49 +11,37 @@ app_client_id = "{{ pillar["homu"]["app-client-id"] }}"
 app_client_secret = "{{ pillar["homu"]["app-client-secret"] }}"
 
 [web]
-
-# The port homu listens on
 port = 54856
 
-# An example configuration for repository (there can be many of these)
-[repo.servo]
+{% set repos = [('servo', 'html5ever')] %}
+{% set reviewers = ["glennw","jdm","kmcallister","larsbergstrom","Manishearth","metajack","mbrubeck","Ms2ger","pcwalton","SimonSapin","mrobinson","brunoabinader","saneyuki","nox"] %}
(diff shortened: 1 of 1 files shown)


edunham:
If you've already set the value for `reviewers` in the pillar, I think it should be good. I often trust syntax highlighting to point out errors in files like this, but github's clearly unaware that it's actually jinja-templated. html5ever already has the hook that you described (https://github.com/servo/html5ever/blob/master/.travis.yml). Will you do the repo webhooks, or would you like me to?

metajack:
There's no reviewers value in the pillar. The value is initially set here in the template.
I already did the mentioned steps for html5ever, but we'll have to repeat those steps for the other repos we enable. i wasn't sure where to put such documentation, so I figured I'd at least get it into the PR description.

Suggestions on where docs should go?

edunham:
How does https://github.com/servo/saltfs/wiki/Adding-a-repo-to-Homu look for now, and we can replace it with a link to the new location if we find a better place? 

I'll also update them if the process doesn't work right for this PR.

metajack:
Excellent!

metajack:
Any other concerns, or is this ready to merge and deploy?

edunham:
If there are any syntax errors, I'm not seeing them. Let's merge and deploy :+1:

PR #43, opened 2015-06-10
Add rust-geom and rust-layers to homu repos.

metajack:
Changes have already been made for rust-geom and rust-layers except for the travis.yml updates which I am submitting now.

r? @edunham


edunham:
r+, and r+'d the others as well. 

sorry about the build stuff; I'm currently figuring out why the script from that blog post isn't working right. In retrospect, it would have been less intrusive to workflow here if I'd tested it all in my fork...

PR #47, opened 2015-06-11
Gate on CSS tests on OSX

jgraham:


larsbergstrom:
This needs to also edit homu to add tracking for the "mac3" builder:
https://github.com/servo/saltfs/blob/master/homu/cfg.toml#L31-L32

PR #52, opened 2015-06-16
add azure to homu

Manishearth:


metajack:
@nox Homu checks this repo already, so next time use the normal r+ method.

PR #83, opened 2015-07-24
Run WPT against the release binaries, so that IPC servo can land

larsbergstrom:
r? @edunham 

This change makes our web platform tests run against a release build of servo instead of debug, so that @pcwalton's IPC servo changes can land.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/83)

### DIFF
--- a/buildbot/master/master.cfg
+++ b/buildbot/master/master.cfg
@@ -141,15 +141,15 @@ linux1_factory.addStep(steps.ShellCommand(command=["./mach", "test-ref", "--kind
 
 linux2_factory = util.BuildFactory()
 linux2_factory.addStep(steps.Git(repourl=SERVO_REPO, mode="full", method="clobber"))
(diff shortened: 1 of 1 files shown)


edunham:
Travis failed because a system package is not available to it; not your fault

Were we getting any benefit from compiling with --dev before?

larsbergstrom:
@edunham The main benefit from `--dev` is that it tests for integer overflow. We plan to add it back so that we're testing both release (what we actually ship) and dev (the only way to get the overflow checking) once a Cargo feature lands that allows us to compile just some critical libraries in release during a dev build so that we aren't so slow we hit all the timeouts.

edunham:
r+ thanks for the explanation! I'll trust your judgement on whether adding it back needs to be captured in an issue or not :)

larsbergstrom:
Will do, but in the main Servo repo! I'll hit the big "merge" button since Travis is borken :-(

Thanks!

PR #84, opened 2015-07-29
Shuffle jobs between builders to hopefully reduce turnaround time.

Ms2ger:
[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/84)

PR #86, opened 2015-07-29
Install Salt from ppa rather than curl|sh

edunham:
Old way was fragile, and broke, as noted most recently in https://github.com/servo/saltfs/pull/84#issuecomment-125971687 

It'd be even better to use Travis's containers, but that's blocked on https://github.com/travis-ci/apt-source-whitelist/issues/82 . Once Saltstack is an approved PPA, it'll be trivial to refactor the `.travis.yml` to stop using sudo as described in http://docs.travis-ci.com/user/migrating-from-legacy/ .

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/86)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,31 +3,30 @@ python:
 - '2.7'
 
 before_install:
-  # Acquire the dependencies which can be installed normally.
+  - sudo add-apt-repository -y ppa:saltstack/salt
   - sudo apt-get update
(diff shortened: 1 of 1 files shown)


metajack:
@bors-servo r+

Thanks for tracking this down!

---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/86)

PR #106, opened 2015-08-29
Rename the builders.

Ms2ger:
[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/106)


Manishearth:
Can we take this moment to prefix all homu-run builders with `auto-` so that they get grouped nicely in the waterfall?

metajack:
The only non-homu run builder is android-nightly. So I'm not sure this solves anything.

PR #107, opened 2015-08-31
Run WPT tests in debug mode

nox:
This lets us use debug_assert!() in Servo. Tests should be run fast enough now that Servo improved.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/107)

### DIFF
--- a/buildbot/master/master.cfg
+++ b/buildbot/master/master.cfg
@@ -105,7 +105,7 @@ c['schedulers'] = []
 c['schedulers'].append(schedulers.AnyBranchScheduler(
     name="servo-auto",
     treeStableTimer=None,
-    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-ref-unit", "mac-rel-css", "android", "gonk"],
+    builderNames=["linux-dev-ref-unit", "linux-dev-wpt", "mac-dev-wpt", "mac-dev-ref-unit", "mac-dev-css", "android", "gonk"],
     change_filter=util.ChangeFilter(filter_fn=servo_auto_try_filter),
 ))
 c['schedulers'].append(schedulers.SingleBranchScheduler(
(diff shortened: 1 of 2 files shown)


Manishearth:
You're adding steps to the release builders; they'll build Servo in release, and then build it again in debug to make wpt work. Not what we want :smile:

nox:
Oh, I need to build debug too I guess?

nox:
Isn't the configuration a mess? I think linux_dev_factory and mac_dev_ref_unit_factory are defined twice.

metajack:
If you want to switch to dev builds, then you should also rename the factories to match. Calling it linux_rel_factory when it builds dev is not the best.

Also, running wpt in debug mode when everything is compiled in release mode will about double our compile times (since it will have to do it twice).

Probably we should switch to running everything in debug. Later we can add new slaves and expand to cover release mode again.

nox:
Rebased and amended, I also renamed the factories and builders accordingly.

Ms2ger:
How about the CSS tests? Did they improve as well?

nox:
@Ms2ger I have absolutely no idea. In doubt, I didn't change it.

Ms2ger:
Except you did on Linux

nox:
I have no idea what you mean. Where did I change test-css?

nox:
Oh right, it's right in the middle of it!

nox:
So I guess I need to either change the other tests to stay consistent, or to move CSS tests on Linux in a separate factory?

metajack:
Reviewed 2 of 2 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/107)

metajack:
You should change the mac-rel-css factory to mac-dev-css. That will make everything consistent. r=me with that change.

PR #108, opened 2015-09-01
iptables rules for servo-master host

edunham:
This opens all of the ports currently available on servo-master, which is both saltmaster and buildmaster. 

```
22 (&TCP6): SSH. Whitelist to bastion if we get one, otherwise allow from anywhere
80: nginx for Buildbot status page. Allow from anywhere.
4505, 4506: Salt master; allow from minions only
54856: Homu (python3)
9001, 8010: Buildbot
UDP & UDP6
123: NTP
```

The only other ports currently in use on the master are for dhclient, which is unnecessary since it has a static IP address. 

It seems like Salt should have a better way to get minion IPs for line 112, but some searching revealed conflicting advice, so I figured it would be more efficient to just consult the PR's reviewer. I got the list from https://github.com/servo/servo/wiki/Buildbot-administration . 

I checked the iptables state [source](https://github.com/saltstack/salt/blob/develop/salt/states/iptables.py) and it looks like rules are only saved if they do not match an existing rule, so they won't get rewritten on each run.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/108)

### DIFF
--- a/buildbot/master.sls
+++ b/buildbot/master.sls
@@ -60,5 +60,77 @@ remove-old-build-logs:
   cron.present:
     - name: 'find /home/servo/buildbot/master/*/*.bz2 -mtime +5 -delete'
     - user: root
-    - minute: 1 
+    - minute: 1
     - hour: 0
+
+iptables:
+  service.running:
+    - enable: True
+    - reload: True
+
+# Open TCP ports for nginx, Homu, Buildbot, ssh
+{% for port in '54856','9001','9010','ssh' %}
+iptables-open-port-{{ port }}:
+  iptables.append:
+    - table: filter
+    - chain: INPUT
+    - jump: ACCEPT
+    - dport: ssh
+    - proto: tcp
+    - save: True
+{% endfor %}
+
+# Enable SSH IPv6 connections
+iptables-enable-ipv6:
+  iptables.append:
+    - table: filter
+    - chain: INPUT
+    - jump: ACCEPT
+    - dport: ssh
+    - proto: tcp
+    - save: True
+    - family: ipv6
+
+# Open ports for ntp
+{% for family in 'ipv4','ipv6' %}
+iptables-enable-ntp-{{ family }}:
+  iptables.append:
+    - table: filter
+    - chain: INPUT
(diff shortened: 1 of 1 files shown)


metajack:
We'll need to set up static IPs for the EC2 slaves I guess and we'll have to use those IPs here.

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---

<sup>**[buildbot-master.sls, line 112 \[r1\]](https://reviewable.io:443/reviews/servo/saltfs/108#-Jy9O3sk4cpD-FZ62Sy-)** ([raw file](https://github.com/servo/saltfs/blob/afa3cc21571cb4bde5d4232a158e4709852a950b/buildbot-master.sls#L112)):</sup>
This is not going to work. The EC2 slaves do not have static IPs currently.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/108)

edunham:
@metajack Compared to leaving the ports open to the world or opening the ports to anyone in [our EC2 region](https://ip-ranges.amazonaws.com/ip-ranges.json), setting up [elastic IPs](http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/elastic-ip-addresses-eip.html) does look like the best option.

buildbot-master.sls
```diff
@@ -65,3 +65,74 @@ buildbot-github-listener:
...
+    - save: True
+    - family: {{ family }}
+{% endfor %}
+
+# Open the ports Salt needs, but only to the minions.
+# Minion IPs should probably come from Pillar? Spelling out the list here
+# doesn't seem like the right way.
+{% for minion in '208.52.161.130', '208.52.161.128', '63.135.170.19',
+    '208.52.170.250', '66.228.48.56', '173.255.201.95', '45.79.167.177',
+    '72.14.176.110','96.126.114.185', '0.0.0.0'  %}
```
claudijd: Just an FYI here, but having '0.0.0.0' in the list I would suspect has an *ANY* effect, which makes the other IPs redundant.

PR #116, opened 2015-09-15
Revert "Run WPT tests in debug mode"

metajack:
This reverts commit 0e985389723a06b9cf51ed502af32e93f467289a.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/116)


metajack:
Just a note that this revert is because of serde being too slow still in debug mode. We'd like to take this patch in general, but until that is fixed we're going to have to stick to release mode.

metajack:
Deployed.

PR #117, opened 2015-09-16
Use correct package name for mscorefonts

metajack:
Salt has some bug with virtual package names, so it failed mysteriously
when it was msttcorefonts.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/117)


metajack:
Tested in production already.

PR #120, opened 2015-09-21
Revert "Check if the lockfile changed"

metajack:
This reverts commit f2ea93aad059894d370252614913071757735eb8.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/120)


Manishearth:
Ah, I should have put that check in a bash script, python doesn't directly run shell builtins

PR #122, opened 2015-10-01
Add a dependency on autoconf-2.13

michaelwu:
This is necessary for https://github.com/servo/mozjs/pull/56 which runs autoconf at build time instead of checking in configure.

No idea if this will work on osx..

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/122)


metajack:
@bors-servo r+


---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/122)

PR #123, opened 2015-10-01
Kill Servo processes left around

metajack:
After running wpt tests, we sometimes end up with zombie servo processes
at 100% CPU. This kills those off to reset machine state.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/123)


metajack:
Already tested in production.

Issue #142, opened 2015-10-30
buildbot-master needs python3


Manishearth:
Homu only works on a python3 toolchain. Our salt configuration does not autoinstall this.

Homu worked so far because it somehow had a python3 virtualenv with everything it needed, without a global install of any of the python3 tools.

Specifically, we need:

```
sudo apt-get install python3.4 python3-pip
pip3 install virtualenv
```

We can do the former with a `- require: - pkg: ` key in `homu.sls` somewhere.

I'm not sure how to indicate a pip3-specific dependency, though. Perhaps something with `bin_env`?

We also need to make sure that the pip/python/virtualenv setup does not make 3.4 the default for the `pip`/`python`/`virtualenv` binaries. Installing them as shown above seemed to preserve the default, but I'm not sure if that's something we can rely on.

cc @edunham @larsbergstrom

edunham:
We might run into https://github.com/saltstack/salt/issues/5494 with Salt. One workaround would be to create the virtualenv, with the flags telling it to use python3, through a shell command.

Manishearth:
Huh. It worked for us, the current salt configuration just says "virtualenv.managed, but use the binary called `virtualenv3.4`" and it works

PR #144, opened 2015-11-01
Enable usage of the Trusty beta on Travis

aneeshusa:
Travis now has new Trusty based instances in Beta. They offer much
faster boot times, twice as much memory, and a more reliable core count.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/144)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,22 +3,33 @@
 language: cpp
 
 # NOTE: Make sure the matrix covers all node types in top.sls
+# sudo: required and dist: trusty enable usage of Trusty
 matrix:
   include:
-    - os: linux
-      env: SALT_NODE_ID=servo-master
-    - os: linux
-      env: SALT_NODE_ID=servo-linux1
-    - os: osx
-      env: SALT_NODE_ID=servo-mac1
-    - os: osx
-      env: SALT_NODE_ID=servo-macpro1
-    - os: linux
-      env: SALT_NODE_ID=linux1
-    - os: linux
-      env: SALT_NODE_ID=servo-linux-android1
-    - os: linux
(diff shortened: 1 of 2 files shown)


aneeshusa:
Also see http://docs.travis-ci.com/user/ci-environment/#Virtualization-environments for a summary of the differences between the standard (Precise) and beta (Trusty) environments.

Manishearth:
@bors-servo r+

Should we be turning this on for other repos that require sudo?

Manishearth:
We also have Trusty on our linux boxes so this works out well!

aneeshusa:
Which other repos? I was looking through the Travis docs (actually looking into #142) and noticed they now have a beta for Trusty environments. There only seem to be advantages to switching, so I'd recommend yes after testing to make sure it doesn't cause any regressions.

aneeshusa:
Ooh, the increased parity also sounds like a good idea, +1 for keeping all environments in sync as much as possible.

Manishearth:
Well, `sudo: false` is still faster than the Trusty thing right now.

The repos that need `sudo: required` are actually using `sudo: 9000`, which is a whole different thing and probably won't work with this Trusty stuff.

aneeshusa:
Yeah, the significantly faster speed of the container infrastructure makes sense for anything that doesn't need sudo.

What's sudo:9000? A quick search seems to say it's Ubuntu 14.04 on GCE, which AFAICT is the same as the Trusty beta.

Manishearth:
Unpublicized faster builds on GCE (experimental feature), not sure if it's exactly the same as trusty, @larsbergstrom would know.

larsbergstrom:
They're the same thing.

PR #155, opened 2015-11-05
Wrap shell scripts for safety

aneeshusa:
Wrap the entirety of the install_salt and configure_salt shell scripts
in a function to prevent problems if interrupted mid-download, as
preparation to update the installation guide in the wiki to use these
scripts via curl | sudo sh.

Also make scripts portable and do a bit of cleanup.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/155)

### DIFF
--- a/.travis/configure_salt
+++ b/.travis/configure_salt
@@ -1,12 +1,15 @@
-#/usr/bin/env bash
+#!/usr/bin/env sh
 
-set -e
-set -u
-set -o pipefail
+configure_salt () {
+    set -o errexit
+    set -o nounset
 
-sudo mkdir -p /etc/salt
-sudo cp .travis/minion /etc/salt/minion
+    sudo mkdir -p /etc/salt
+    sudo cp .travis/minion /etc/salt/minion
 
-sudo mkdir -p /srv/salt
-sudo cp -r . /srv/salt/states
-sudo cp -r .travis/test_pillars /srv/salt/pillars
+    sudo mkdir -p /srv/salt
+    sudo cp -r . /srv/salt/states
+    sudo cp -r .travis/test_pillars /srv/salt/pillars
+}
+
+configure_salt "$@"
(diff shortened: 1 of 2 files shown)


aneeshusa:
The install instructions on the wiki don't pin the salt version, which will cause problems if run afresh, so this will allow reusing the install script from the repo for the install instructions to keep the versions in sync. (Right now, installing salt from homebrew on OS X will break everything as 2015.8.0 is bad, and 2015.8.1 isn't packaged in homebrew yet.)

metajack:
Question below, otherwise looks great.


---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[.travis/configure_salt, line 5 \[r1\]](https://reviewable.io:443/reviews/servo/saltfs/155#-K2briBemSJ7HWnz_NTq)** ([raw file](https://github.com/servo/saltfs/blob/d6f5e390d733d52ca351e60b1e40ca7e29821ec6/.travis/configure_salt#L5)):</sup>
Is pipefail not needed anymore?


---

<sup>**[.travis/install_salt, line 5 \[r1\]](https://reviewable.io:443/reviews/servo/saltfs/155#-K2brkYyJgeGHlD-XnT9)** ([raw file](https://github.com/servo/saltfs/blob/d6f5e390d733d52ca351e60b1e40ca7e29821ec6/.travis/install_salt#L5)):</sup>
Ditto


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/155)

aneeshusa:
Unfortunately, raw `sh` doesn't support `set -o pipefail`, only `bash` and `zsh` do.

aneeshusa:
Thanks for getting this merged! I updated the Linux installation instructions on the wiki, but I don't have a Mac and am not confident about updating the current Mac instructions to use the install script as well.  Can I get some help with that, or can someone with more Mac experience update those instructions to be more in line with the Linux ones?

PR #167, opened 2015-11-25
Added ajeffrey's ssh key.

asajeffrey:
Added my ssh public key to common/ssh/ and added my name to common/map.jinja.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/167)


common/map.jinja
```diff
@@ -13,6 +13,7 @@
           'servo-head': '96.126.114.185'
         },
         'ssh_users': [
+	  'ajeffrey',
```
Manishearth: nit: ws

asajeffrey:
*[common/map.jinja, line 16 \[r1\]](https://reviewable.io:443/reviews/servo/saltfs/167#-K3zNmU0mIMW_Qa3WNdU-r1-16:-K3zRl7vMuwzrj-vdGnD:-1152851933) ([raw file](https://github.com/servo/saltfs/blob/15daae04c036bef3d95924bbf1eb9422b598c0a0/common/map.jinja#L16)):*
Grr, emacs tabs.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/167)*

Issue #165, opened 2015-11-21
Workflow for upgrading Salt


aneeshusa:
Homebrew has packaged Salt 2015.8.1 so I'll put in a PR to update the pinned salt version soon. However, upgrading Salt requires some extra care for two reasons:
  1. It's best to restart the service when the package is updated to keep the installed and running versions in sync, but restarting a master or minion service will interrupt an ongoing highstate.
  2. [Masters need to be updated before minions](https://docs.saltstack.com/en/latest/faq.html#can-i-run-different-versions-of-salt-on-my-master-and-minion), but `salt '*' state.highstate` cannot enforce ordering - Salt's [orchestration](https://docs.saltstack.com/en/latest/topics/tutorials/states_pt5.html) can help with this.

I want to make sure there is a concrete plan for this and future Salt version upgrades to prevent breakage. Do you currently have a workflow for updating Salt? If not, I think manual updates are fine for now since there are so few machines: first on the salt master, then on the minions. I'll look into adding some automation to this.

aneeshusa:
Blocked on https://github.com/saltstack/salt/issues/29098.

edunham:
@aneeshusa for the "document the workflow" part of this issue, I've moved the Salt docs into their own wiki page at https://github.com/servo/servo/wiki/SaltStack-Administration. 

It looks like the Salt issue that this was previously blocked on is now resolved.

aneeshusa:
Great, closing since we've agreed on manual updates for now.

PR #177, opened 2015-12-15
[DO NOT MERGE] Update to Salt 2015.8.3

aneeshusa:
This update also removes usage of deprecated options, so these states will not work properly on 2015.5. (The run will attempt to succeed with an exit code of 0, but with a warning that reflects the broken state.) Hence, it is important that the the Salt installations in production are updated when these states are deployed.

Blocked on #165.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/177)


aneeshusa:
Whoops, forgot about OS X, see #178.

PR #180, opened 2015-12-22
Make Salt install/configure Vagrant 1.8 compatible

aneeshusa:
The original Vagrantfile was meant to be used with Vagrant 1.7.4, but
required 2 patches to Vagrant for proper functionality, which I had
submitted as pull requests. Vagrant has since released a new version,
1.8.0, which has updates the Salt provisioner. PR status:
 - [One PR](https://github.com/mitchellh/vagrant/pull/6474) was merged
 - [Another PR](https://github.com/mitchellh/vagrant/pull/6473) was not

The second PR was rejected "in favor" of [a third PR](https://github.com/mitchellh/vagrant/pull/6073) which simply
removed the config_dir option entirely instead of using sudo to be able
to write in a privileged directory. [An unrelated PR](https://github.com/mitchellh/vagrant/pull/6382/files) removed a second
option we had been using: install_command.

These changes in Vagrant were due to incompatibilities with the official
Salt bootstrap script, so this commit brings the install_salt script's
behavior closer to that of the official Salt bootstrap script as well.
Namely, it can now copy over configuration files as part of
installation, using the same flags as the official script.

The configure_salt script is now used just to setup the salt and pillar
roots on Travis (this is accomplished via shared folders in Vagrant),
hence the renaming.

With this commit and the new Vagrant version, the included Vagrantfile
now works with a vanilla Vagrant 1.8.0 install, no patching needed.

Note to self: update the wiki when this is merged!

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/180)


aneeshusa:
This is fairly low-importance, but help from someone with a Mac would be appreciated on this one! I have no idea how OS X got '.travis/install_salt' for $OS_NAME (via $1) in the error message...

johntron:
The title is misleading - I thought you were trying to undo the changes we waited so long to get. This is great though, thanks!

The build seems to be failing, because you've [changed the parameter order of the call to `install_salt`](https://github.com/servo/saltfs/pull/180/files#diff-354f30a63fb0907d4ad57269548329e3R35), but the `install_salt` script wasn't updated to use the correct parameter index (maybe `$5` instead of `$1`). This means that instead of receiving the value of `${TRAVIS_OS_NAME}`, instead it gets something that's probably considered an invalid value when used with `getopt` - this seems to be what's causing the error message. 

You'll also need to update [the other instance of `$1`](https://github.com/servo/saltfs/pull/180/files#diff-55324b37a48fe5bcd6f115a60506a041R62) (possibly others if I missed one). I could be wrong, but might be able to test by using the same call to `install_salt` as Travis: `.travis/install_salt -F -c .travis -- "osx"` (kill the command before it does anything).

aneeshusa:
@johntron Sorry, I didn't realize the title was misleading. Feel free to change it to something that's more understandable.

Looking at your comments now.

aneeshusa:
Sorry, I should have mentioned this before - it looks like the problem is that getopt behaves differently on OS X vs. Linux. (The builds are successful on Linux hosts but not OS X hosts). I couldn't figure out how it differs from the [man page](https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/getopt.1.html), so I'd appreciate help from someone with a Mac that can investigate what getopt is doing.

I changed the parameter order to mirror the flags used by the official script, and am using getopt to parse them portably. Here's the behavior I want, which is what's happening on Linux: the [arg-parsing loop](https://github.com/aneeshusa/servo-saltfs/blob/49faf6610814e03b57ee577a000fa128ad60fbe9/.travis/install_salt#L39) should call ```shift``` after each arg it handles to bring the next one up to the first spot in "$1", and stop after it sees the "--". At this point, the only arg left should be the positional parameter (either "linux" or "osx"), which is why I'm reading "$1" into OS_NAME.

aneeshusa:
What's especially confusing to me is that ".travis/install_salt" should be ```$0```, so I have no idea how it got into the ```$1``` slot to go into ```OS_NAME```.

aneeshusa:
I figured out how to make this work on OS X! It would be super helpful for me if this and #181 got merged in (ideally this one first).

aneeshusa:
Also, a note to myself: when this gets merged, remember to update the installation notes on the wiki.

aneeshusa:
Thanks for getting this merged! Being able to use Vagrant locally is really useful in testing out changes (and prevents a ton of amended commits over small typos caught by Travis :smile:). I've updated the Vagrant instructions on the wiki; it turns out the installation instructions don't need changing after all.

larsbergstrom:
Thank you for all your help! Sorry this one lingered a bit in the review queue.

PR #184, opened 2016-01-03
So I can try to figure servo#8641 out.

notriddle:
[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/184)


jdm:
@bors-servo: r+
`<larsberg> Let me (or jack) know if he has trouble accessing the machine! Now that our linux builders are EC2 instances, the IPs change every time they get up/down'd`

PR #191, opened 2016-01-08
Add an RSA SSH key for me

aneeshusa:
[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/191)


larsbergstrom:
the key should be named `common/ssh/aneeshusa.pub`

aneeshusa:
Whoops, thanks for catching that.

PR #195, opened 2016-01-12
Hack to add more latent builders in case of ill-timed restart

larsbergstrom:
We should never have more than two active, but when one builder "dies" for w/e reason, for the duration of the buildbot run, the machine will not be restarted. This adds 3, getting us to 5, an arbitrary number that I hope will keep us at our 2 live instances at any point in time until we can either pull the trigger on reserved instances or get better about Graceful Shutdown, etc.

Also, removed `servo-linux1`, which is no longer around but was still on the waterfall page.

r? @edunham @metajack @Manishearth

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/195)

### DIFF
--- a/buildbot/master/master.cfg
+++ b/buildbot/master/master.cfg
@@ -11,8 +11,8 @@ from passwords import MINION_PUBLIC_KEY, MINION_PRIVATE_KEY
 
 SERVO_REPO = "https://github.com/servo/servo"
 HEAD_SLAVES = ["servo-head"]
-LINUX_SLAVES = ["linux1", "linux2"]
-LINUX_LINODE_SLAVES = ["servo-linux1"]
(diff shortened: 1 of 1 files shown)


larsbergstrom:
Aha, this will create more than 2 live instances in the case where there is both a real build and a try build going at the same time.

metajack:
I'm not a huge fan of this, since it will make the real problem harder to debug and easier to ignore.

metajack:
@bors-servo r+

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/195#-:-K7qcpNkPm1haZ5XEP__:1236647940)*

PR #200, opened 2016-01-19
Remove 5-day-old zipped logs from buildslave dirs every night

edunham:
Low disk space command from https://github.com/servo/servo/wiki/Buildbot-administration#handling-low-disk-space . We should no longer need to escape the semicolon, since Bash eats semicolons but Salt doesn't (right?)

Salt cron syntax from https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cron.html

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/200)

### DIFF
--- a/buildbot/master.sls
+++ b/buildbot/master.sls
@@ -51,3 +51,10 @@ buildbot-github-listener:
     - watch:
       - file: /usr/local/bin/github_buildbot.py
       - file: /etc/init/buildbot-github-listener.conf
+
+remove-old-build-logs:
+  cron.present:
+    - name: 'find /home/servo/buildbot/master/*/*.bz2 -mtime +5 -delete'
+    - user: root
+    - minute: 1 
+    - hour: 0


buildbot/master.sls
```diff
@@ -58,3 +58,9 @@ buildbot-github-listener:
     - mode: 644
     - watch_in:
       - service: buildbot-github-listener
+
+find /home/servo/buildbot/master/*/*.bz2 -mtime +5 -exec rm {} ;:
```
ecoal95: I *think* you can do `-delete` instead of `-exec rm {}` (which may or may not need the semicolon) :P

edunham: Yep, that works. Fixed; thanks for catching that.

aneeshusa:
I'd suggest using a more suggestive id and putting the actual command as the ```name``` parameter, like so:

```salt
remove-old-build-logs:
  cron.present:
    - name: 'find /home/servo/buildbot/master/*/*.bz2 -mtime +5 -delete'
    - user: root
    - minute: 1 
    - hour: 0
```

I also recommend quoting complicated commands like these to make it clear that they are strings and avoid ambiguity in parsing, but strictly speaking the quotes are optional.

This makes it easier to see what the state is for and easier to reference it from another state. Here's the Salt documentation about [names](https://docs.saltstack.com/en/latest/ref/states/highstate.html#name-declaration). Additionally, for the cron states in particular, the state id will become an ```identifier argument``` to the state, and using a stable identifier will allow for more robust updates in the future. The link @edunham posted to the Salt docs has more information about this behavior.

larsbergstrom:
@edunham Can you address this when you get a chance? We hit another out of disk space on buildbot master today :-(

edunham:
Overwrote old branch with @aneeshusa 's recommended change applied to latest master

PR #211, opened 2016-02-02
Link autoconf after installing autoconf213 on OS X

aneeshusa:
This needs to be tested manually on one of the Mac builders before merging.

Use --overwrite because autoconf and autoconf213 are installed, and
either one may have been linked beforehand. They overlap in the files
they provide, and Homebrew will refuse to destructively overwrite
any symlinks without an explicit flag.

There's no built-in Salt module for brew link, so for now this is a
quick fix. This provides some idempotence for repeated Salt highstate
runs by checking the directory where Homebrew keeps track of linked
packages (and linking autoconf if it is not linked). It's unknown
whether this approach properly handles updates to one or both packages.

A better approach would be to write a custom Salt state (perhaps
homebrew.linked); even if this state works properly, a custom state
would be more declarative and could be upstreamed to Salt proper.

Closes servo/servo#9504.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/211)

### DIFF
--- a/servo-dependencies.sls
+++ b/servo-dependencies.sls
@@ -30,6 +30,20 @@ servo-darwin-homebrew-versions-dependencies:
       - homebrew/versions
     - require_in:
       - pkg: servo-dependencies
+
+homebrew-link-autoconf:
+  cmd.run:
+    - name: 'brew link --overwrite autoconf'
+    - user: administrator
+      # Warning: Only checks that some autoconf Homebrew package is linked,
+      # not necessarily the version installed above.
+      # Whether this handles updating autoconf properly is an open question.
+      # This state should be replaced by a custom Salt state.
+    - creates: /usr/local/Library/LinkedKegs/autoconf
+    - require:
+      - module: servo-darwin-homebrew-versions-dependencies
+    - require_in:
+      - pkg: servo-dependencies
 {% endif %}
 
 servo-dependencies:


aneeshusa:
cc @larsbergstrom

aneeshusa:
I was worried about handling version upgrades properly, so I tried testing things out manually. However, autoconf213 is already pinned to a version, and autoconf has been at 2.69 for the last ~3-4 years; homebrew won't install autoconf 2.68 anymore because of the MD5 checksum. Based on some quick searching, it doesn't seem like autoconf 2.70 will be released anytime soon. I'm comfortable punting on the question of if this handles version upgrades properly for now, as there's no good way to test it, and we can handle it if autoconf 2.70 is ever released.

aneeshusa:
For future reference:
Here are the manual test cases that I'm going to be testing. Starting from each of these states, I'll run the Salt highstate, then attempt to build servo. 
- [ ] Start without autoconf or autoconf213 installed
- [ ] Only autoconf is already installed
- [ ] Only autoconf213 is already installed
- [ ] Both autoconf and autoconf213 are already installed
Let me know if you can think of any other cases to test by hand.

larsbergstrom:
That's the full set of cases! Thanks again for doing all of this work :-) And I apologize for the slow builds - we're just getting `CCACHE` support enabled right now, and it really only helps with one part of our build (SpiderMonkey, the JS engine).

aneeshusa:
These builds are only taking ~5-6 minutes, which is pretty fast to build a whole browser :)

Once I'm done testing these cases, this should be ready for review.

larsbergstrom:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/211)*

aneeshusa:
OK, all done testing.

larsbergstrom:
Thanks again for all your work on this - much appreciated!

aneeshusa:
Cool, I'll see about applying this to all the Mac builders. But first, I'm off to watch Spectre!

aneeshusa:
I just checked and it looks like all the Mac builders already have autoconf linked.

larsbergstrom:
@aneeshusa Sorry, I missed the earlier comment! Yes, there are a few things where we've manually "fixed" the builders (for better or worse), so we really only notice problems like this when I spin up a new builder.

I hope you enjoyed Spectre!

PR #220, opened 2016-02-11
Add an operators group

larsbergstrom:
r?  @Manishearth 

Manish, do you have any idea how I can test this without dropping things on their head? The goal here is to have a "role" for operators, separate from that of reviewers (even if it's the same thing under the hood) to let people self-retry without waiting for delegate permissions.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/220)

### DIFF
--- a/homu/cfg.toml
+++ b/homu/cfg.toml
@@ -85,9 +85,12 @@ secret = "{{ pillar["homu"]["web-secret"] }}"
     "zmike",
 ] %}
 
+{% set operators = [
+    "bholley",
+] %}
+
 {% set try = [
     "askalski",
-    "bholley",
     "chandler",
     "dzbarsky",
     "gsnedders",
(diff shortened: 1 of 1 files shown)


KiChjang:
Doesn't the try group already allow members to retry?

larsbergstrom:
Aha, I'm wrong - it's for delegated `r+`, not `retry`. Thanks, @KiChjang!

Manishearth:
Try allows for retry, but not for re-r+ given a change.

We don't need to test this, we can deploy it and check that the generated cfg.toml is as expected.

aneeshusa:
"We don't need to test this" makes me worried :worried: 

I agree that checking the generated cfg.toml is a sufficient test for this; I'd suggest [doing that in a Vagrant VM](https://github.com/servo/servo/wiki/Buildbot-administration#recommended) before deploying to production. You can also run a highstate with `test=True` to see the expected changes (although in this case it looks like the generated file will be the same afterwards).

Manishearth:
Oh, we always do `test=True` before deploying :smile: 

I meant that we don't need to test this by opening a dummy PR and having Bobby review it. We'll do the regular salt tests of course.

larsbergstrom:
I confirmed with the nifty Vagrant wizardry that this works and generates a good cfg.toml for homu.

r? @Manishearth

PR #227, opened 2016-02-27
Use Salt to manage increased maxfiles on OS X

aneeshusa:
These settings are necessary to handle running the WPT (web platform
tests); taken out of the wiki to reduce the need for manual
interventions.

Note, /etc/launchd.conf is not included because OS X no longer uses
this file and it has no effect. Whether a replacement for this
functionality is required is TBD.

Also, note that this hasn't been tested since there is no OS X Vagrant box.

cc @larsbergstrom since you added the maxfiles section to the wiki

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/227)

### DIFF
--- a/osx/files/profile
+++ b/osx/files/profile
@@ -1,0 +1,11 @@
+# System-wide .profile for sh(1)
+
+if [ -x /usr/libexec/path_helper ]; then
+        eval `/usr/libexec/path_helper -s`
+fi
+
+if [ "${BASH-no}" != "no" ]; then
+        [ -r /etc/bashrc ] && . /etc/bashrc
+fi
+
+ulimit -n 10240
--- a/osx/files/sysctl.conf
+++ b/osx/files/sysctl.conf
@@ -1,0 +1,3 @@
+kern.maxfiles=65536
+kern.maxfilesperproc=30000
+kern.ipc.somaxconn=2048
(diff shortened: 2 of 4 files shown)


larsbergstrom:
Reviewed 4 of 4 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/227)*

larsbergstrom:
@bors-servo r+

Thanks a ton for automating this! I've been doing it by hand, but this will really help. We're looking to add a bunch more OSX builders, and I probably would have forgotten this otherwise :-)

aneeshusa:
Wait, do we need a replacement for ```/etc/launchd.conf```? OS X no longer honors the file, but I presume we added it for a reason. Are we no longer running into those limits with the WPTs?

I'm looking to do some testing on servo-macpro1 in the next few days and try to make the big block of commands needed to install on OS X much smaller :)

larsbergstrom:
I landed because I'm pretty sure we don't need it if it's being ignored anyway :-) All of our OSX builders are on 10.10.x. We haven't been running into those limits with the WPTs, and I'm sure we would otherwise, because the defaults are *super* low.

Thanks again for helping automate all of this stuff!

PR #228, opened 2016-02-29
Pin virtualenv pip package to pin installed binary

aneeshusa:
I noticed homu broke (sometime today I guess) while working on another PR.

A recent update of the virtualenv package changed the installed
binaries from `virtualenv` and `virtualenv-3.4`, to `virtualenv` and
`virtualenv-3.5`, breaking the homu installation which needs a python3
version of virtualenv. Pin the package to the newest version and update
the homu installation to use the virtualenv-3.5 binary.

Note: the python3 apt package still installs python3.4, but the version
difference should be ok since virtualenv is just used to install homu,
not run it.

@Manishearth as I said in the commit message I don't think combining
python3.4 and virtualenv-3.5 should cause problems, but how can I test
homu manually to make sure it runs properly?

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/228)

### DIFF
--- a/common/init.sls
+++ b/common/init.sls
@@ -22,11 +22,12 @@ pip:
       {% endif %}
     - reload_modules: True
 
-# Virtualenv package creates virtualenv and virtualenv-3.4 executables
+# virtualenv == 14.0.6 package creates virtualenv and virtualenv-3.5 executables
+# note that the version of the second may change between virtualenv versions
 virtualenv:
   pip.installed:
     - pkgs:
-      - virtualenv
+      - virtualenv == 14.0.6
     - require:
       - pkg: pip
 
--- a/homu/init.sls
+++ b/homu/init.sls
@@ -11,7 +11,7 @@ homu:
     - user: servo
   virtualenv.managed:
     - name: /home/servo/homu/_venv
-    - venv_bin: virtualenv-3.4
+    - venv_bin: virtualenv-3.5
     - python: python3
     - system_site_packages: False
     - require:


Manishearth:
Pop some dummy values into the config (stick a valid github login and repo URL in it), and ensure `service start homu` outputs the following:

```
Bottle v0.12.9 server starting up (using WaitressServer())...
Listening on http://:54856/
Hit Ctrl-C to quit.

serving on http://0.0.0.0:54856
```

in `/var/log/upstart/homu.log`

Manishearth:
This is what's currently on the server, btw:

```
root@servo-master:~# pip show virtualenv
---
Metadata-Version: 2.0
Name: virtualenv
Version: 1.11.6
Summary: Virtual Python Environment builder
Home-page: https://virtualenv.pypa.io/
Author: Jannis Leidel, Carl Meyer and Brian Rosner
Author-email: python-virtualenv@groups.google.com
License: MIT
Location: /usr/local/lib/python2.7/dist-packages
Requires: 
You are using pip version 7.1.2, however version 8.0.2 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
root@servo-master:~# pip3 show virtualenv
---
Name: virtualenv
Version: 13.1.2
Location: /usr/local/lib/python3.4/dist-packages
Requires: 
```

aneeshusa:
OK, tested it locally and homu ran successfully. (It only needed an access token with the user:email scope.)

Manishearth:
@bors-servo r+

Oh, yeah, that thing shouldn't actually be necessary (we don't use that setting); I should remove it.

aneeshusa:
I check the homu source, and it looks like it wants access to the user:email data in order to have an email to use when making commits via git. For example, bors-servo commits with the email 'lbergstrom+bors@mozilla.com'. I guess this could be configured as a separate config option, but I do think it's useful to have an email associated with homu.

Manishearth:
Oh ah. I was thinking about a different optional key which seems to now actually be optional.

Manishearth:
Deployed, works. There's a virtualenv-3.4 still hanging around, but it's not breaking anything as far as I can tell.

aneeshusa:
That's weird, I'm going to look into it. I don't want it causing any problems in the future (like #203).

Manishearth:
That's probably because the salt-redeploy doesn't uninstall things?

aneeshusa:
Well, the virtualenv pip package was already installed, so it should just update it to a newer version via pip. I'm guessing this is a pip behavior, but it could be due to Salt or the virtualenv package itself.

PR #229, opened 2016-02-29
Disable service autostart on package install

aneeshusa:
Debian and its derivatives will autostart services when installing
packages. However, often we need to configure installed software
before it is ready to run, so it's preferable to use Salt states and
requisites to handle package installation or upgrades, configuration
changes and service [re]starting.

Adding a policy-rc.d file with exit code 101 sets a policy which denies
this default behavior. Making this change in the install script also
prevents new builders from attempting to contact the Salt master before
they have been configured properly.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/229)


aneeshusa:
@larsbergstrom this should prevent some of those spurious entries in ```salt-key -L``` we were discussing in #222.

Reminder to self: update the wiki installation instruction after this lands.

larsbergstrom:
@bors-servo r+

Thanks - this looks great to me! I think the wiki instruction changes will also be key for OSX, where we have to do a much more manual installation of salt.

aneeshusa:
@larsbergstrom This PR actually only affects the Linux builders, but I've gone some more PRs in the pipeline:)

BTW, I'm testing some things on servo-macpro1 and it seems to be having some networking issues (I'm testing out installing some things through brew). I believe it's not hooked up to buildbot yet and no one else seems to be using it at the moment - is it ok if I try a reboot?

larsbergstrom:
@aneeshusa Go right ahead and reboot it!

@mbrubeck was the only other one looking at it, and I don't believe he's actively using it right now.

PR #232, opened 2016-03-05
Add appveyor support for servo's homu

Manishearth:
Don't merge yet, needs https://github.com/servo/homu/pull/13 to merge and then a bump in the homu SHA
r? @larsbergstrom

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/232)

### DIFF
--- a/homu/cfg.toml
+++ b/homu/cfg.toml
@@ -120,6 +120,9 @@ try_builders = ["linux-dev", "linux-rel", "android", "mac-dev-unit", "mac-rel-wp
 username = "{{ pillar["homu"]["buildbot-http-user"] }}"
 password = "{{ pillar["homu"]["buildbot-http-pass"] }}"
 
+[repo.servo.status.appveyor]
+context = 'continuous-integration/appveyor/branch'
+
 # Standard per-repo configs (generated)
 
 {% for repo in travis_repos %}
--- a/homu/init.sls
+++ b/homu/init.sls
@@ -6,7 +6,7 @@ python3:
 homu:
   git.latest:
     - name: https://github.com/servo/homu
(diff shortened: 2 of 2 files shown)


larsbergstrom:
r=me after SHA bump

larsbergstrom:
```
          ID: android-dependencies

    Function: pkg.installed

      Result: False

     Comment: An error was encountered while installing package(s): W: Failed to fetch http://dl.google.com/linux/chrome/deb/dists/stable/Release  Unable to find expected entry 'main/binary-i386/Packages' in Release file (Wrong sources.list entry or malformed file)

              

              E: Some index files failed to download. They have been ignored, or old ones used instead.

     Started: 16:53:16.581770

    Duration: 3715.043 ms

     Changes:   
```

aneeshusa:
Google has dropped support for 32-bit Chrome (planned for March hence the sudden breakage): http://www.omgubuntu.co.uk/2016/03/fix-failed-to-fetch-google-chrome-apt-error-ubuntu

Is there any reason we need Chrome? If not, I'd rather remove it entirely than trying to modify the appropriate `.list` file to only try to get the 64-bit builds.

Manishearth:
We don't explicitly mention Chrome anywhere. I don't know why it tries to download chrome.

aneeshusa:
I checked via Salt on the builders and you're right, none of our machines have chrome in their apt sources.
The error only happens on the machines with multi-arch enabled, which is the 2 android related machines. My hunch is that the Travis builders have chrome in their sources. I'll try it out in Vagrant to see if that's the case.

aneeshusa:
Yup, looks ok in Vagrant. I'll put in a test PR to check Travis.

aneeshusa:
See #233 for the Travis test PR

PR #233, opened 2016-03-05
Test commit to check Travis's apt sources

aneeshusa:
Google has just dropped support for 32-bit Chrome, and it seems that
Travis has the Chrome repository enabled, causing build failures on our
multi-arch machines. This is a commit to check if Travis actually does
have Chrome in its sources.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/233)


aneeshusa:
Yup, looks like Travis has the Chrome apt repository configured. I'll put in a PR to get around it.

PR #234, opened 2016-03-05
Remove non-Salted external apt sources

aneeshusa:
Workaround https://github.com/saltstack/salt/issues/26605
by cleaning the sources.list.d folder first, and running
pkgrepo states that add repositories to the folder afterwards
using require.

This unbreaks #232: Travis has the Chrome apt repo configured, but
Google has dropped 32-bit support (March 2016), causing apt to fail
on multiarch hosts. By removing all external repos we will only use
repositories we configure ourselves.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/234)

### DIFF
--- a/longview/init.sls
+++ b/longview/init.sls
@@ -1,0 +1,8 @@
+longview:
+  pkgrepo.managed:
+    - name: 'deb http://apt-longview.linode.com/ trusty main'
+    - file: /etc/apt/sources.list.d/longview.list
+    - key_url: https://apt-longview.linode.com/linode.gpg
+    - require:
+      - file: /etc/apt/sources.list.d
+
--- a/top.sls
+++ b/top.sls
@@ -33,3 +33,4 @@ base:
     - buildbot.master
     - homu
     - nginx
+    - longview
--- a/ubuntu/init.sls
+++ b/ubuntu/init.sls
@@ -6,3 +6,18 @@
     - group: root
     - mode: 755
     - source: salt://{{ tpldir }}/files/policy-rc.d
+
+# Workaround for https://github.com/saltstack/salt/issues/26605
+# Clean the directory first, and require it in all pkgrepo states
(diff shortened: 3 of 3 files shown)


aneeshusa:
Heads up: I checked the builders and `servo-master` has a `longview.list` file in the sources.list.d folder, which seems to be for the [Linode monitoring agent](https://github.com/linode/longview) package. If this is something we need to keep around, let me know and I'll add a commit to ensure that package stays around; this PR will remove it otherwise when applied!

larsbergstrom:
@edunham Do you know about the longview.list thing? Can we just allow it to be removed since we'll be moving it to EC2 at some point, or do we have to turn off the monitoring in the linode manager tool first?

aneeshusa:
I can also tell Salt to keep it around if it's important, so not a problem if we're using it (sounds like it might be good to keep it around even if we're not using it right now, just in case).

larsbergstrom:
Yeah, I think that might be best - then we can remove it once we actually make the move to EC2.

aneeshusa:
Do you know if there's a way to tell from inside a VM whether it's a Linode VM? Alternately, is `servo-master` the only VM left on Linode?

larsbergstrom:
I do not know if there's an easy way to tell. The VMs on EC2 are `servo-linux?`. The `servo-mac*` are all on macstadium. The reset (`servo-head`, `servo-master`, and `servo-linux-android1`) are all currently on linode.

aneeshusa:
Interesting, only `servo-master` has the longview.list file, so it may not be necessary after all. (It's also the only machine which has the longview package installed.)

larsbergstrom:
Makes sense - I bet that's the only machine where we have the extra monitoring or linode turned on.

aneeshusa:
OK, that commit will re-add `longview.list` after all external sources are removed. It doesn't specify that the package be installed, however.

PR #239, opened 2016-03-08
Add support for ARM32/ARM64 to our build infrastructure

larsbergstrom:
r? @aneeshusa 

I tested these changes inside of the *awesome* vagrant setup, and they seem to work locally, so I think that the next step is to review my miserable salt work.

There's some pretty hack-y stuff - assumptions of target triples differ between the Rust compiler, installed gcc binaries, and configure scripts, so there is some `PATH` and symlink hackery going on that could hopefully be reduced over time.

I've also renamed `servo-linux-androidN` to `servo-linux-crossN`, to better reflect the change. I'm assuming that we will need two cross builders online for {android, gonk, arm32, arm64} builds. Possibly three, as arm32/arm64 release builds are *slow*.

CC @edunham @Manishearth @metajack @mmatyas

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/239)

### DIFF
--- a/arm-dependencies.sls
+++ b/arm-dependencies.sls
@@ -1,0 +1,111 @@
+{% from 'common/map.jinja' import config %}
+
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set rootfs_repo = 'https://servo-rust.s3.amazonaws.com/ARM' %}
+{% set targets = [{
+                    'name': 'arm-linux-gnueabihf',
+                    'symlink_name': 'arm-unknown-linux-gnueabihf',
+                    'version': 'v1',
+                    'local_name': 'armhf-trusty-libs.tgz',
+                    'hash': 'd9a31ed488e4f848efcd07f71aa167fc73252da2a2c3b53ba8216100e2b4302b5ccd273b27c434ad189650652a1877607d328ff6b8e1edb5aa68a8927c617b49',
+                  },
+                  {
+                    'name': 'aarch64-linux-gnu',
+                    'symlink_name': 'aarch64-unknown-linux-gnu',
+                    'version': 'v1',
+                    'local_name': 'arm64-trusty-libs.tgz',
+                    'hash': '6c86097188b70940835b2fc936fe70f01890fae45ba4ef79dcccc6a552ad319dcba23e21d6c849fd9d396e0c2f4476a21c93f9b3d4abb4f34d69f22d18017b1b',
+                  }] %}
+
+{% set binaries = [
+    'elfedit',
+    'gcov',
+    'nm',
+    'addr2line',
+    'g++',
+    'objcopy',
+    'ar',
+    'objdump',
+    'as',
+    'gcc',
+    'ranlib',
+    'c++filt',
+    'gprof',
+    'readelf',
+    'cpp',
+    'ld',
+    'size',
+    'ld.bfd',
+    'strings',
+    'strip'
+] %}
+
+{% for target in targets %}
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: {{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }} # Directory to extract into
+    - source: {{ rootfs_repo }}/{{ target.version }}/{{ target.local_name }}
+    - source_hash: sha512={{ target.hash }}
+    - archive_format: tar
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+
+{% for binary in binaries %}
+{{ config.servo_home }}/bin/{{ target.symlink_name }}-{{ binary }}:
+  file.symlink:
+    - target: /usr/bin/{{ target.name }}-{{ binary }}
+    - makedirs: True
+    - require:
+      - archive: libs-{{ target.name }}
+{% endfor %}
+
+{{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }}:
+  file.directory:
+    - user: servo
+    - group: servo
+    - dir_mode: 755
+    - file_mode: 644
+    - makedirs: True
+    - require:
+      - archive: libs-{{ target.name }}
+
+{{ config.servo_home }}/rootfs-trusty-{{ target.name }}:
+  file.directory:
+    - user: servo
+    - group: servo
+    - dir_mode: 755
+    - file_mode: 644
+    - makedirs: True
+    - clean: True
+    - require:
+      - file: {{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }}
+
+{% for root in ['usr/include', 'usr/lib', 'lib'] %}
+/{{ root }}/{{ target.name }}:
+  file.symlink:
+    - target: {{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }}/{{ root }}/{{ target.name }}
+    - require:
+      - archive: libs-{{ target.name }}
+{% endfor %}
+
+{% endfor %}
+
+{{ config.servo_home }}/bin:
+  file.directory:
+    - user: servo
+    - group: servo
+    - dir_mode: 755
+    - file_mode: 644
+    - makedirs: True
+    - clean: True
+    - require:
+      {% for target in targets %}
+      {% for binary in binaries %}
+      - file: {{ config.servo_home }}/bin/{{ target.symlink_name }}-{{ binary }}
+      {% endfor %}
+      {% endfor %}
--- a/buildbot/master/master.cfg
+++ b/buildbot/master/master.cfg
@@ -11,7 +11,7 @@ SERVO_REPO = "https://github.com/servo/servo"
 HEAD_SLAVES = ["servo-head"]
 LINUX_RESERVED_SLAVES = ["servo-linux1", "servo-linux2"]
 MAC_SLAVES = ["servo-mac1", "servo-mac2", "servo-mac3"]
-ANDROID_SLAVES = ["servo-linux-android1"]
+CROSS_SLAVES = ["servo-linux-cross1"]
 
 # These will lock a project to a single platform
 MAC_ONLY_PROJECTS = ['servo/io-surface-rs']
@@ -30,7 +30,7 @@ c['caches'] = {
 ####### BUILDSLAVES
 
 c['slaves'] = []
-for s in MAC_SLAVES + ANDROID_SLAVES + HEAD_SLAVES + LINUX_RESERVED_SLAVES:
+for s in MAC_SLAVES + CROSS_SLAVES + HEAD_SLAVES + LINUX_RESERVED_SLAVES:
     c['slaves'].append(buildslave.BuildSlave(s, SLAVE_PASSWORD, max_builds=1))
 
 ####### CHANGESOURCES
@@ -72,7 +72,7 @@ c['schedulers'] = []
 c['schedulers'].append(schedulers.AnyBranchScheduler(
     name="servo-auto",
     treeStableTimer=None,
-    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk"],
+    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "arm32", "arm64"],
     change_filter=util.ChangeFilter(filter_fn=servo_auto_try_filter),
 ))
 c['schedulers'].append(schedulers.SingleBranchScheduler(
@@ -83,7 +83,7 @@ c['schedulers'].append(schedulers.SingleBranchScheduler(
 ))
 c['schedulers'].append(schedulers.ForceScheduler(
     name="force",
-    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "android-nightly"]
+    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "arm32", "arm64", "android-nightly"]
 ))
 c['schedulers'].append(schedulers.Nightly(
     name="Nightly",
@@ -242,6 +242,34 @@ gonk_factory = create_servo_factory([
     steps.ShellCommand(command=["bash", "./etc/ci/manifest_changed.sh"], env=gonk_compile_env),
 ])
 
+arm32_compile_env = dict({'PATH': '/home/servo/bin:$PATH',
+                          'BUILD_TARGET': 'arm-unknown-linux-gnueabihf',
+                          'PKG_CONFIG_ALLOW_CROSS': '1',
+                          'PKG_CONFIG_PATH': '/usr/lib/arm-linux-gnueabihf/pkgconfig',
+                          'EXPAT_NO_PKG_CONFIG': '1',
+                          'FREETYPE2_NO_PKG_CONFIG': '1',
+                          'FONTCONFIG_NO_PKG_CONFIG': '1'
+                          'CC': 'arm-linux-gnueabihf-gcc',
+                          'CXX': 'arm-linux-gnueabihf-g++'},
+                        **linux_test_env)
+arm32_factory = create_servo_factory([
+    steps.Compile(command=["./mach", "build", "--rel", "--target=$BUILD_TARGET"], env=arm32_compile_env),
+])
+
+arm64_compile_env = dict({'PATH': '/home/servo/bin:$PATH',
+                          'BUILD_TARGET': 'aarch64-unknown-linux-gnu',
+                          'PKG_CONFIG_ALLOW_CROSS': '1',
+                          'PKG_CONFIG_PATH': '/usr/lib/aarch64-linux-gnu/pkgconfig',
+                          'EXPAT_NO_PKG_CONFIG': '1',
+                          'FREETYPE2_NO_PKG_CONFIG': '1',
+                          'FONTCONFIG_NO_PKG_CONFIG': '1'
+                          'CC': 'aarch64-linux-gnu-gcc',
+                          'CXX': 'aarch64-linux-gnu-g++'},
+                        **linux_test_env)
+arm64_factory = create_servo_factory([
+    steps.Compile(command=["./mach", "build", "--rel", "--target=$BUILD_TARGET"], env=arm64_compile_env),
+])
+
 doc_factory = create_servo_factory([
     steps.ShellCommand(command=["etc/ci/upload_docs.sh"],
                        env={'TOKEN': GITHUB_DOC_TOKEN,
@@ -308,16 +336,29 @@ c['builders'].append(util.BuilderConfig(
 ))
 c['builders'].append(util.BuilderConfig(
     name="android",
-    slavenames=ANDROID_SLAVES,
+    slavenames=CROSS_SLAVES,
     factory=android_factory,
     nextBuild=branch_priority,
     category="auto",
 ))
-# NOTE: gonk currently shares the android builder.
 c['builders'].append(util.BuilderConfig(
     name="gonk",
-    slavenames=ANDROID_SLAVES,
+    slavenames=CROSS_SLAVES,
     factory=gonk_factory,
+    nextBuild=branch_priority,
+    category="auto",
+))
+c['builders'].append(util.BuilderConfig(
+    name="arm32",
+    slavenames=CROSS_SLAVES,
+    factory=arm32_factory,
+    nextBuild=branch_priority,
+    category="auto",
+))
+c['builders'].append(util.BuilderConfig(
+    name="arm64",
+    slavenames=CROSS_SLAVES,
+    factory=arm64_factory,
     nextBuild=branch_priority,
     category="auto",
 ))
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ matrix:
       os: linux
       sudo: required
       dist: trusty
-    - env: SALT_NODE_ID=servo-linux-android1
+    - env: SALT_NODE_ID=servo-linux-cross1
       os: linux
       sudo: required
       dist: trusty
--- a/.travis/test_pillars/top.sls
+++ b/.travis/test_pillars/top.sls
@@ -15,7 +15,7 @@ base:
     - match: pcre
     - buildbot.slave
 
-  'servo-linux-android\d+':
+  'servo-linux-cross\d+':
     - match: pcre
     - buildbot.slave
 
--- a/top.sls
+++ b/top.sls
@@ -13,11 +13,12 @@ base:
     - buildbot.slave
     - android-dependencies
 
-  'servo-linux-android\d+':
+  'servo-linux-cross\d+':
     - match: pcre
     - buildbot.slave
     - android-dependencies
     - gonk-dependencies
+    - arm-dependencies
 
   'servo-(mac|macpro)\d+':
     - match: pcre


aneeshusa:
I don't have an opinion on renaming the android builders, but you should be aware that Salt matches minion keys by ID, so if you rename the minion you also need to do some key shuffling. The ideal way to do this is spin up a completely new machine with the `servo-linux-crossN` setup (i.e. it will generate a new key) and then decom the old `servo-linux-android` machine (and also unaccept/delete its key); this gives zero downtime and is more robust.

You can also just rename the machine (i.e., by setting the `id` key in the `/etc/salt/minon` config file on the machine) and restart the minion,  but when it comes back up it won't be able to connect. To fix this, you'll also need to log into `servo-master`, cd into the `/etc/salt/pki/master/minions/` directory, and rename the `servo-linux-android1` file to `servo-linux-cross1`. This updates the Salt key database and the minion should automatically reconnect.

aneeshusa:
It looks like the `arm32` and `arm64` setups are very similar, so this should use [Jinja templating](http://jinja.pocoo.org/docs/dev/templates/) (probably some kind of loop like https://github.com/servo/saltfs/blob/master/common/init.sls#L40) in order to generate both configs.

arm-dependencies.sls
```diff
@@ -0,0 +1,78 @@
...
+      - g++-arm-linux-gnueabihf
+
+/home/servo/bin:
+  file.directory:
+    - user: servo
+    - group: servo
+    - dir_mode: 755
+    - file_mode: 644
+    
+arm64_links:
```
aneeshusa: style nit: I tend to prefer dashes (`-`) to underscores (`_`) in state IDs.

arm-dependencies.sls
```diff
@@ -0,0 +1,78 @@
...
+  cmd.wait:
```
aneeshusa: These commands should use the `creates` parameter to avoid being re-executed on subsequent runs. Docs are here: https://docs.saltstack.com/en/2015.5/ref/states/all/salt.states.cmd.html#salt.states.cmd.wait.

aneeshusa: Actually, we should be use [file.symlink](https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html#salt.states.file.symlink) inside a jinja loop for this logic instead. It's easier to see the intention and ensures we know exactly which symlinks are being created (and we get the `creates` behavior for free).

aneeshusa: Also, these states should require the `/home/servo/bin` state. They also shouldn't need to require the arm-dependencies state at all (you can symlink to files that don't exist yet), although this should be double-checked.

buildbot/master/master.cfg
```diff
@@ -242,6 +242,36 @@ gonk_factory = create_servo_factory([
     steps.ShellCommand(command=["bash", "./etc/ci/manifest_changed.sh"], env=gonk_compile_env),
 ])
 
+arm32_compile_env = dict({'PATH': '/home/servo/bin:$PATH',
+                          'BUILD_TARGET': 'arm-unknown-linux-gnueabihf',
+                          'TRIPLET': 'arm-linux-gnueabihf',
```
aneeshusa: Is the `TRIPLET` variable required at run time or is it only used to set `CC` and `CXX`? If the latter, I'd prefer to use python string formatting to bake the right value into those variables and avoid polluting the environment with the `TRIPLET` variable.

arm-dependencies.sls
aneeshusa: This should probably go into the style guide.

arm-dependencies.sls
```diff
@@ -0,0 +1,78 @@
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+/home/servo/bin:
```
aneeshusa: I think we should add some subdirectories inside this folder to keep the arm32 and arm64 binaries cleanly separated.

arm-dependencies.sls
larsbergstrom: Do you have an example of such a jinja loop? The complication here is that there's also a file rename going on with the symlink (to change `-linux` to `-unknown-linux`). Sorry I'm such a noob at it, but I couldn't find a good example in the salt docs.

buildbot/master/master.cfg
larsbergstrom: It's not - will remove!

arm-dependencies.sls
larsbergstrom: Will do!

larsbergstrom:
I'm happy to move the `servo-linux-android` machines to `servo-linux-cross*` and shut down the old ones. We (@edunham and I) have been planning to move that machine, which is at linode, over to EC2, and this seems like a great opportunity to do so.

arm-dependencies.sls
larsbergstrom: Hrm, your earlier template document on jinja is better; I'll look at that in the morning.

larsbergstrom:
@aneeshusa OK, I think that I have *finally* managed to address everything except for trying to template the `for ` loop over the symlinks, which I wasn't easily able to figure out. Does this look closer?

larsbergstrom:
Failed in travis. Probably:
```
[0;31m----------[0m
    [0;31m      ID: /usr/include/arm-linux-gnueabihf[0m
    [0;31mFunction: file.symlink[0m
    [0;31m  Result: False[0m
    [0;31m Comment: The following requisites were not found:
                                 require:
                                     archive: libs-arm-linux-gnueab[0m
    [0;31m Started: [0m
    [0;31mDuration: [0m
[0;31m     Changes:   [0m
[0;31m----------[0m
    [0;31m      ID: /usr/include/aarch64-linux-gnu[0m
    [0;31mFunction: file.symlink[0m
    [0;31m  Result: False[0m
    [0;31m Comment: The following requisites were not found:
                                 require:
                                     archive: libs-arm-linux-gnueab[0m
    [0;31m Started: [0m
    [0;31mDuration: [0m
[0;31m     Changes:   [0m

**Comment by larsbergstrom** [comment]
Two more:
```
[0;31m----------[0m
    [0;31m      ID: /home/servo/buildbot/slave[0m
    [0;31mFunction: file.recurse[0m
    [0;31m  Result: False[0m
    [0;31m Comment: #### /home/servo/buildbot/slave/buildbot.tac ####
              Unable to manage file: Jinja variable 'collections.OrderedDict object' has no attribute 'master'[0m
    [0;31m Started: 15:15:02.288078[0m
    [0;31mDuration: 6088.391 ms[0m
[0;31m----------[0m
    [0;31m      ID: buildbot-slave[0m
    [0;31mFunction: service.running[0m
    [0;31m  Result: False[0m
    [0;31m Comment: One or more requisite failed: buildbot.slave./home/servo/buildbot/slave[0m
    [0;31m Started: [0m
    [0;31mDuration: [0m
```

**Comment by larsbergstrom** [comment]
@aneeshusa Can you please take a look at this? I'd love to get ARM32+ARM64 automation online, if this is closer to what you're thinking.

It's now passing Travis and seems to work in my local vagrant box.

I do still need some help with the more wizardly jinja rules for symlinks in place of my wonky `cmd.run` with the shell replacement scripts, but hopefully we could land without and fix that in a follow-up, if that works for you :-)

File: arm-dependencies.sls (hunk 1 above)
Comment (aneeshusa) [inline arm-dependencies.sls]: I'd still like this to get added to the style guide, but not a blocker.

File: arm-dependencies.sls [hunk 5]
```diff
@@ -0,0 +1,63 @@
...
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: /home/servo/rootfs-trusty-{{ target.name }} # Directory to extract into
+    - source: {{ target.file }}
+    - source_hash: sha512={{ target.hash }}
+    - archive_format: tar
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+    - if_missing: /home/servo/{{ target.local_name }}
+
+/usr/include/{{ target.name }}:
```
aneeshusa: We should loop over the /usr/include, /usr/lib/ and /lib states with Jinja.

arm-dependencies.sls
```diff
@@ -0,0 +1,63 @@
...
+    - name: 'for f in /usr/bin/{{ target.name }}*; do f2=$(basename $f); ln -s $f /home/servo/bin/{{ target.name }}/${f2/-linux/-unknown-linux}; done'
+    - user: servo
+    - group: servo
+    - require:
+      - pkg: arm-dependencies
+      - file: /home/servo/bin/{{ target.name }}
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: /home/servo/rootfs-trusty-{{ target.name }} # Directory to extract into
```
aneeshusa: These directories should also have versions in their path (like the Android NDK), so that Salt doesn't simply skip this state when we update the rootfs because it sees there is a previous version already extracted.

aneeshusa:
For the symlinks:
If you look inside the `/home/servo/bin/{aarch64-linux-gnu,arm-linux-gnueabihf}` folders, you'll see 29 symlinks being created inside each folder. The symlinks have the form `$PREFIX-$BINARY`, where `$PREFIX` is `aarch64-unknown-linux-gnu` and `arm-unknown-linux-gnueabihf`, respectively, and `$BINARY` is things like `as`, `gcc`, and `ld`. Inside the Jinja block where you set `targets`, you can add the respective `$PREFIX` for each target to the set of variables, and you can use another Jinja `set` command to make a list of just the binaries we need (hopefully not all 29! we should be able to skip the versioned links at least, like `gcc-4.8`). Then, you can loop over the contents of this new Jinja variable of the binaries to generate all the symlink states.

Less important but nice to have:
Also, the symlink states should use ` - makedirs: True`, and we should move the state that currently make the subfolders inside `/home/servo/bin` after the symlink states. That state should be amended to `require` all of the symlink states (using another Jinja loop to generate the `require`s), and we should add ` - clean: True` to it.

Let me know if something is unclear here! IMO, the original bash script will be harder to read because of the substitutions inside the f2 variable - most people aren't very familiar with these. For that reason, I would like to get this fixed before merging if possible, but doing it in a follow-up wouldn't be out of the question.

arm-dependencies.sls
```diff
@@ -0,0 +1,63 @@
...
+    - source: {{ target.file }}
+    - source_hash: sha512={{ target.hash }}
+    - archive_format: tar
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+    - if_missing: /home/servo/{{ target.local_name }}
```
aneeshusa: `archive.extracted` actually downloads the archive to a temporary location and extracts it directly to the location indicated by `name`, so it does not leave the tarball in this location. Please remove this line to prevent spurious re-downloads (without `if_missing`, it will check for the `name` argument instead, which is what we want.)

larsbergstrom:
hrm...
```
[0;31m----------[0m
    [0;31m      ID: /usr/includeaarch64-linux-gnu[0m
    [0;31mFunction: file.symlink[0m
    [0;31m  Result: False[0m
    [0;31m Comment: The following requisites were not found:
                                 require:
                                     archive: libs-/usr/include-aarch64-linux-gnu[0m
    [0;31m Started: [0m
    [0;31mDuration: [0m
[0;31m     Changes:   [0m
[0;31m----------[0m
    [0;31m      ID: /usr/libaarch64-linux-gnu[0m
    [0;31mFunction: file.symlink[0m
    [0;31m  Result: False[0m
    [0;31m Comment: The following requisites were not found:
                                 require:
                                     archive: libs-/usr/lib-aarch64-linux-gnu[0m
    [0;31m Started: [0m
    [0;31mDuration: [0m
[0;31m     Changes:   [0m
[0;31m----------[0m
    [0;31m      ID: /libaarch64-linux-gnu[0m
    [0;31mFunction: file.symlink[0m
    [0;31m  Result: False[0m
    [0;31m Comment: The following requisites were not found:
                                 require:
                                     archive: libs-/lib-aarch64-linux-gnu[0m
    [0;31m Started: [0m
    [0;31mDuration: [0m
[0;31m     Changes:   [0m
[0;36m
```

arm-dependencies.sls
```diff
@@ -0,0 +1,52 @@
...
+    - name: 'for f in /usr/bin/{{ target.name }}*; do f2=$(basename $f); ln -s $f /home/servo/bin/{{ target.name }}/${f2/-linux/-unknown-linux}; done'
+    - user: servo
+    - group: servo
+    - require:
+      - pkg: arm-dependencies
+      - file: /home/servo/bin/{{ target.name }}
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: /home/servo/v1/rootfs-trusty-{{ target.name }} # Directory to extract into
```
aneeshusa: Please pull the version (i.e. the `1`) out into a variable (or two probably - we should be able to handle different versions for the rootfs files). I think a good place might be inside the `targets` variable.

arm-dependencies.sls
```diff
@@ -0,0 +1,52 @@
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set targets = [{ 'name': 'arm-linux-gnueabihf', 
+                    'file': 'https://servo-rust.s3.amazonaws.com/ARM/v1/armhf-trusty-libs.tgz',
```
aneeshusa: Use the (to-be-added) `version` and `local_name` variables to create these URLs where it's needed in the archive.extracted state, instead of setting them here.

arm-dependencies.sls
```diff
@@ -0,0 +1,52 @@
...
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: /home/servo/v1/rootfs-trusty-{{ target.name }} # Directory to extract into
+    - source: {{ target.file }}
+    - source_hash: sha512={{ target.hash }}
+    - archive_format: tar
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+
+{% for root in ['/usr/include', '/usr/lib', '/lib'] %}
+{{ root }}{{ target.name }}:
```
aneeshusa: I think you may be missing a slash between these two.

arm-dependencies.sls
```diff
@@ -0,0 +1,52 @@
...
+  file.symlink:
+    - target: /home/servo/v1/rootfs-trusty-{{ target.name }}{{ root }}{{ target.name }}
```
aneeshusa: Same here, I think we need a slash between root and target.name.

larsbergstrom:
OK, I *think* I've made all of the requested changes and travis passes :-)

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
...
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set path = 'https://servo-rust.s3.amazonaws.com/ARM/' %}
+{% set targets = [{ 'name': 'arm-linux-gnueabihf', 
+                    'unknown_name': 'arm-unknown-linux-gnueabihf',
+                    'version': 'v1',
+                    'file': 'armhf-trusty-libs.tgz',
```
aneeshusa: Looks like the `file` variables are redundant here with `local_name`. I think `local_name` is clearer, so let's keep those ones.

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
...
+                'strings',
+                'dwp',
+                'ld.gold',
+                'strip'] %}
+
+{% for target in targets %}
+
+{% for file in binaries %}
+
+/home/servo/bin/{{ target.unknown_name }}-{{ file }}:
```
aneeshusa: style nit: this state is pretty short (6 lines), so I'd remove the empty newlines above and below it so the for loop lines are right next to the state. I find this makes it easier to see exactly where a short loop starts and ends, especially since we have nested loops.

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set path = 'https://servo-rust.s3.amazonaws.com/ARM/' %}
+{% set targets = [{ 'name': 'arm-linux-gnueabihf', 
+                    'unknown_name': 'arm-unknown-linux-gnueabihf',
```
aneeshusa: `unknown_name` sounds weird to me - maybe `symlink_name` instead?

larsbergstrom:
OK, hopefully the last time! Thanks for sticking with this review through all of this :-)

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
...
+/home/servo/bin/{{ target.name }}:
+  file.directory:
+    - user: servo
+    - group: servo
+    - dir_mode: 755
+    - file_mode: 644
+    - makedirs: True
+    - clean: True
+    - require:
+{% for file in binaries %}
```
aneeshusa: style nit: When I have for loops in require statements, I tend to keep the for loop at the same level of indentation so it doesn't break the block shape. Here, I would indent this line and the corresponding `endfor` over by six spaces.

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
...
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+
+{% for root in ['/usr/include', '/usr/lib', '/lib'] %}
+{{ root }}/{{ target.name }}:
+  file.symlink:
+    - target: /home/servo/{{ target.version }}/rootfs-trusty-{{ target.name }}{{ root }}/{{ target.name }}
+    - require:
+      - archive: libs-{{ target.name }}
+
+{% endfor %}
```
aneeshusa: style nit: swap this `endfor` and the empty line above it, similar reasoning as above

aneeshusa:
Getting close! Let me take a look inside Vagrant with the new changes.

Also, some of the style nits I'm making should probably get added to the style guide.

arm-dependencies.sls
```diff
@@ -0,0 +1,85 @@
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set path = 'https://servo-rust.s3.amazonaws.com/ARM/' %}
+{% set targets = [{ 'name': 'arm-linux-gnueabihf',
```
aneeshusa: whitespace: I'm seeing an extra space at the end of this line.

arm-dependencies.sls
```diff
@@ -0,0 +1,85 @@
...
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set path = 'https://servo-rust.s3.amazonaws.com/ARM/' %}
+{% set targets = [{ 'name': 'arm-linux-gnueabihf', 
+                    'symlink_name': 'arm-unknown-linux-gnueabihf',
+                    'version': 'v1',
+                    'local_name': 'armhf-trusty-libs.tgz',
+                    'hash': 'd9a31ed488e4f848efcd07f71aa167fc73252da2a2c3b53ba8216100e2b4302b5ccd273b27c434ad189650652a1877607d328ff6b8e1edb5aa68a8927c617b49',
+                   },
```
aneeshusa: style nit: I tend to like my close brackets lining up vertically with my open brackets (this one and the next one are one space too far over to the right).

arm-dependencies.sls
```diff
@@ -0,0 +1,85 @@
...
+                  { 'name': 'aarch64-linux-gnu',
+                    'symlink_name': 'aarch64-unknown-linux-gnu',
+                    'version': 'v1',
+                    'local_name': 'arm64-trusty-libs.tgz',
+                    'hash': '6c86097188b70940835b2fc936fe70f01890fae45ba4ef79dcccc6a552ad319dcba23e21d6c849fd9d396e0c2f4476a21c93f9b3d4abb4f34d69f22d18017b1b',
+                   }] %}
+
+{% set binaries = ['elfedit',
```
aneeshusa: style nit: I think I'd put `elfedit` on the next line, only use 4 spaces of indent for the whole list, and add a newline right after strip but before the closing brackets and end of tag.

arm-dependencies.sls
```diff
@@ -0,0 +1,85 @@
...
+                'dwp',
+                'ld.gold',
+                'strip'] %}
+
+{% for target in targets %}
+
+{% for file in binaries %}
+/home/servo/bin/{{ target.symlink_name }}-{{ file }}:
+  file.symlink:
+    - target: /usr/bin/{{ target.name }}-{{ file }}
```
aneeshusa: Just `file` is a little ambiguous - I'd probably do `for binary in binaries` in this loop and below.

arm-dependencies.sls
```diff
@@ -0,0 +1,85 @@
...
+    - clean: True
+    - require:
+      {% for file in binaries %}
+      - file: /home/servo/bin/{{ target.symlink_name }}-{{ file }}
+      {% endfor %}
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: /home/servo/{{ target.version }}/rootfs-trusty-{{ target.name }} # Directory to extract into
+    - source: {{ path }}{{ target.version }}/{{ target.local_name }}
```
aneeshusa: style nit: Directly concatenating variables in paths or URLs always looks weird to me, so I'd remove the slash at the end of the `path` variable and put it in here.

arm-dependencies.sls
```diff
@@ -0,0 +1,85 @@
...
+    - makedirs: True
+    - clean: True
+    - require:
+      {% for file in binaries %}
+      - file: /home/servo/bin/{{ target.symlink_name }}-{{ file }}
+      {% endfor %}
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: /home/servo/{{ target.version }}/rootfs-trusty-{{ target.name }} # Directory to extract into
```
aneeshusa: We should import from `common/map.jinja` and use the servo_home variable from there.

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
+{% from 'common/map.jinja' import config with context %}
```
aneeshusa: I've been meaning to update this, but we don't need `with context`.

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
...
+    'readelf',
+    'cpp',
+    'ld',
+    'size',
+    'ld.bfd',
+    'strings',
+    'dwp',
+    'ld.gold',
+    'strip'
+    ] %}
```
aneeshusa: style nit: This line doesn't need any indentation.

arm-dependencies.sls
```diff
@@ -0,0 +1,89 @@
+{% from 'common/map.jinja' import config with context %}
+
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set path = 'https://servo-rust.s3.amazonaws.com/ARM' %}
+{% set targets = [{ 'name': 'arm-linux-gnueabihf',
```
aneeshusa: I'd put 'name' on the next line instead of the same line as the opening brace. (same below)

larsbergstrom: Only indented 4 or still where it is?

arm-dependencies.sls
larsbergstrom: Then should I also undent the similar line at the end of the definition of `targets` above?

aneeshusa:
Thanks for your patience! I don't have any other style comments right now, so I'm going to test this out in Vagrant - this will probably take a while (I'm guessing at least half an hour?). You might also want to test that the cross builders still work!

larsbergstrom:
Sounds good! I'll also spin up a Vagrant instance, but it takes that long on my old laptop, too, so I may be asleep again by then :-) I'll give it a quick build-test, too (e.g., to make sure that the instructions in buildbot/master.cfg are still accurate after all of the directory changes).

arm-dependencies.sls
```diff
@@ -0,0 +1,91 @@
...
+    - makedirs: True
+    - clean: True
+    - require:
+      {% for binary in binaries %}
+      - file: {{ config.servo_home }}/bin/{{ target.symlink_name }}-{{ binary }}
+      {% endfor %}
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: {{ config.servo_home }}/{{ target.version }}/rootfs-trusty-{{ target.name }} # Directory to extract into
```
aneeshusa: We should switch the ordering here and put the `target.version` directory inside the `rootfs-trusty-target.name` directory.

aneeshusa: Also, we should add a `file.directory` state for the `{{ config.servo_home }}/rootfs-trusty-{{ target.name }}` folder with a `clean:  True` parameter and require this `libs-{{ target.name }}` state in it. This should automatically clean out old versions, preventing issues like #203 by failing fast.

larsbergstrom:
Do we care about whether the symlinks are in a separate directory under `/home/servo/bin` right now? There's a bug that they are all symlinked into `bin`, but then the next rule creates two (empty) directories for the two target types immediately afterwards.

arm-dependencies.sls
```diff
@@ -0,0 +1,91 @@
+{% from 'common/map.jinja' import config %}
+
+arm-dependencies:
+  pkg.installed:
+    - pkgs:
+      - g++-aarch64-linux-gnu
+      - g++-arm-linux-gnueabihf
+
+{% set path = 'https://servo-rust.s3.amazonaws.com/ARM' %}
```
aneeshusa: `path` is a little ambiguous. How about `rootfs_repo` (feel free to put something else if you can think of something better)?

arm-dependencies.sls
```diff
@@ -0,0 +1,91 @@
...
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: {{ config.servo_home }}/{{ target.version }}/rootfs-trusty-{{ target.name }} # Directory to extract into
+    - source: {{ path }}/{{ target.version }}/{{ target.local_name }}
+    - source_hash: sha512={{ target.hash }}
+    - archive_format: tar
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+
+{% for root in ['/usr/include', '/usr/lib', '/lib'] %}
```
aneeshusa: style nit: I would remove the slashes at the beginning of each of these, add a slash to the beginning of the state ID, and add a space between `{{ target.name }}` and `{{ root }}` in the `- target` argument.

larsbergstrom:
I confirmed that at least before commit 1043622 the vagrant install works and I can ssh in and start a build of Servo (I can't finish one - the default vagrant VMs have WAY too little RAM configured <grin>).

aneeshusa:
I think what happened with the `bin` symlinks is I asked you to create subdirectories under `bin` for the symlinks when we still had the `cmd.run` to try to and keep them separate, but now they're empty since we moved to `file.symlink`. Since Salt has more first-hand knowledge of the symlink names and can make sure they don't clash, I don't think we need the subdirectories anymore.

aneeshusa:
How much RAM are you using in the VM? The Vagrantfile should give them each 1 GB, and `free -m` shows that I'm using about `850 MB` right now. It's close for me, but there should be some breathing room. If it isn't enough, I'll put in a PR to bump it up (probably just for the cross builder) - just let me know how much is needed.

aneeshusa:
Err, my numbers are only for running the highstate, I have no idea how much RAM it takes to actually build Servo, so bump it up in the Vagrantfile and let me know how much we need.

aneeshusa:
Actually, we should replace the state that makes the subdirectories with a file.directory state that works on the entire `/home/servo/bin` folder, still with the `- clean: True` argument.

larsbergstrom:
I'm using 1GB, but we need between 3--4GB to build Servo (I'm not sure exactly - cross builds seem to need more). I can do that separately. This is enough to clone Servo and do enough of the build to know that everything is working :-)

larsbergstrom:
OK, everything is addressed, it runs and creates a builder that works for cross-compiles (there was one path thing that needed reflection into the buildbot config for the env). @aneeshusa  r+?

arm-dependencies.sls
```diff
@@ -0,0 +1,104 @@
...
+
+libs-{{ target.name }}:
+  archive.extracted:
+    - name: {{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }} # Directory to extract into
+    - source: {{ rootfs_repo }}/{{ target.version }}/{{ target.local_name }}
+    - source_hash: sha512={{ target.hash }}
+    - archive_format: tar
+    - archive_user: servo # 2015.8 moves these to the standard user and group parameters
+
+{{ config.servo_home }}/rootfs-trusty-{{ target.name }}:
```
aneeshusa: OK, this one is my fault, but we have to fix this: right now this state will blow away the archives we just extracted, due to https://github.com/saltstack/salt/issues/26605, making the whole thing useless. To fix this, we should add a `file.directory` state that ensures the directory created by `archive.extracted` exists (so basically a no-op), then require that state instead of the `libs-` state in this state that has the `- clean: True`. (The bug is that the `clean` parameter only recognizes other `file` states.)

arm-dependencies.sls
```diff
@@ -0,0 +1,104 @@
...
+    'strings',
+    'dwp',
+    'ld.gold',
+    'strip'
+] %}
+
+{% for target in targets %}
+
+{% for binary in binaries %}
+{{ config.servo_home }}/bin/{{ target.symlink_name }}-{{ binary }}:
```
aneeshusa: It would be nice to put this state after the `libs-` state that it requires.

arm-dependencies.sls
```diff
@@ -0,0 +1,104 @@
...
+                  },
+                  {
+                    'name': 'aarch64-linux-gnu',
+                    'symlink_name': 'aarch64-unknown-linux-gnu',
+                    'version': 'v1',
+                    'local_name': 'arm64-trusty-libs.tgz',
+                    'hash': '6c86097188b70940835b2fc936fe70f01890fae45ba4ef79dcccc6a552ad319dcba23e21d6c849fd9d396e0c2f4476a21c93f9b3d4abb4f34d69f22d18017b1b',
+                  }] %}
+
+{% set binaries = [
```
aneeshusa: We're getting some broken symlinks: `gcov-tool` for arm32 and arm64, as well as `dwp` and `ld.gold` on arm64. If we don't need these, we should remove them from the list.

larsbergstrom:
OK, addressed the symlinks and the erroneous clean that needed an extra step.

arm-dependencies.sls
```diff
@@ -0,0 +1,111 @@
...
+    - file_mode: 644
+    - makedirs: True
+    - clean: True
+    - require:
+      - file: {{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }}
+
+{% for root in ['usr/include', 'usr/lib', 'lib'] %}
+/{{ root }}/{{ target.name }}:
+  file.symlink:
+    - target: {{ config.servo_home }}/rootfs-trusty-{{ target.name }}/{{ root }}/{{ target.version }}/{{ target.name }}
```
aneeshusa: So so very close, this should be the last thing: we need to swap `root` and `version` to make these symlinks correct.

aneeshusa:
Once we fix this last bug, I think this should be ready to go after a squash.

Before you deploy this, I saw we have an Android bug that I'd like to fix before you bring up the new `-cross` builder.

aneeshusa:
Please squash so I can approve this.

larsbergstrom:
Great, squashed!

larsbergstrom:
@aneeshusa There isn't, but generally what we do is say "r=me after squash" to people with review priviledges or "@bors-servo delegate" to people who do not, and then they can "@bors-servo r=aneeshusa" when it's ready.

larsbergstrom:
And speaking of, given that homu didn't notice it, I think that I need to open a new PR :-)

larsbergstrom:
@aneeshusa What is the Android bug that you'd like me to fix before we bring up the new -cross builder?

I think that @edunham and I will probably be looking into that next week, as part of the Linode=>EC2 migration (the easy bit <grin>).

aneeshusa:
There's actually a bunch of intertwined Android fixes that are a bit involved (a lot of overlapping changes similar to ones we did here), so expect to see a PR from me in the next few days.

Let me know if you need any help with the migration.

Issue #238, opened 2016-03-08
Vagrant on Linux failing to load due to permission error with the minon file?


larsbergstrom:
@aneeshusa Do you have any idea why this might be failing when I attempt to `vagrant up` a new machine and how to fix it?

```
Copying salt minion config to /etc/salt
Failed to upload a file to the guest VM via SCP due to a permissions
error. This is normally because the SSH user doesn't have permission
to write to the destination location. Alternately, the user running
Vagrant on the host machine may not have permission to read the file.

Source: /home/larsberg/saltfs/.travis/minion
Dest: /etc/salt/minion
[larsberg@thor saltfs]$ 
```

larsbergstrom:
Aha! Looks like this is a problem with newer Vagrants:
https://github.com/mitchellh/vagrant/issues/5973

larsbergstrom:
s/newer/older. Ubuntu installs 1.7.4 Vagrant, but #180 moved us to 1.8.0. So this is user error on my part!

aneeshusa:
Thanks for updating the wiki, I totally meant to add that after the PR landed but forgot! This bug was definitely a thorn in the side for a while.

larsbergstrom:
No problem! It was actually somewhat fortunate, as I wasn't in a huge rush today and it forced me to learn a lot about how Vagrant works :-)

PR #240, opened 2016-03-08
Ensure Vagrant version is at least 1.8.0

aneeshusa:
Use the Vagrant.require_version helper documented here:
https://www.vagrantup.com/docs/vagrantfile/vagrant_version.html

Eliminates issues like #238.

Heads up: I didn't test this with a Vagrant older than 1.8.0, so I have no idea if this actually works.
@larsbergstrom can you test this if possible since you ran into this?
I want to make sure this is actually a useful change.

[Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/240)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,6 +7,8 @@ def extract_id(env)
   id[0]
 end
 
+Vagrant.require_version '>= 1.8.0'
+
 Vagrant.configure(2) do |config|
 
   if Vagrant.has_plugin?('vagrant-cachier')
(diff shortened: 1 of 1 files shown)


edunham:
Build broke due to timeout issues so I've retried it through Travis

edunham:
When I try this on Vagrant 1.4.3, I get the error: 

```
Vagrant failed to initialize at a very early stage:

This Vagrant environment has specified that it requires the Vagrant
version to satisfy the following version requirements:

  >= 1.8.0

You are running Vagrant 1.4.3, which does not satisify
these requirements. Please change your Vagrant version or update
the Vagrantfile to allow this Vagrant version. However, be warned
that if the Vagrantfile has specified another version, it probably has
good reason to do so, and changing that may cause the environment to
not function properly.
```

In my opinion, forbidding Vagrant versions which sporadically break the build is a better use of our time than fixing each quirk that shows up only in old Vagrants. 

r? @larsbergstrom

PR #246, opened 2016-03-12
Enable (re-)running vagrant provision

aneeshusa:
Vagrant will pass the -C flag to the bootstrap script when Salt is
already installed (and only needs to be configured). Implement this
flag for our bootstrap script so that running vagrant provision after
vagrant up, or vagrant provision repeatedly, works correctly.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/246)


larsbergstrom:
@bors-servo r+

Thanks! I had been wondering why `vagrant provision` didn't work :-)

aneeshusa:
I updated the wiki to mention this as a useful command.

PR #248, opened 2016-03-14
Use Vagrant to set the VM hostname

aneeshusa:
This has the benefit that Salt will automatically pick up on the
hostname and use it as the minion id, removing the need to manually
pass the minion ID in the Vagrantfile and when debugging inside the VM.

This is blocked on https://github.com/mitchellh/vagrant/pull/7137 landing
and getting released in a new Vagrant version.

Reminder to self: update the wiki when this lands.

---

This change is [Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/248)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -7,7 +7,7 @@ def extract_id(env)
   id[0]
 end
 
-Vagrant.require_version '>= 1.8.0'
+Vagrant.require_version '>= 1.8.5'
 
 Vagrant.configure(2) do |config|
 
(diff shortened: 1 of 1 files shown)


edunham:
Reviewed 1 of 1 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/248)*

aneeshusa:
Vagrant 1.8.5 will pass `--retcode-passthrough` to Salt, even in masterless mode. r? @edunham

aneeshusa:
I need to make another vagrant PR to add `file_root` and `pillar_root` options for the Salt provisioner, now that we're using those after #350, so this is blocked again.

aneeshusa:
Happily, this is no longer blocked as the Vagrant upstream feature this depended on has been merged and released as part of Vagrant 2.0: https://github.com/mitchellh/vagrant/blob/a124f6635844848f4369e620e1607701d22ddcb9/CHANGELOG.md.

I will try to update this PR soon.

Issue #186, opened 2016-01-04
Problems with setting up the msttfonts on linux


larsbergstrom:
We moved EC2 builders this past week, and had problems with the MS TT fonts installer again. The relevant install lines appear to be here:
https://github.com/servo/saltfs/blob/master/servo-dependencies.sls#L10-L19

What seems to have happened (from looking at the failing machine) is that the mscorefonts-eula was set to `False` rather than true before the installation of the package happened, and so the package was showing as installed but none of the fonts were available. I had to reset the eula flag then forcibly uninstall/reinstall the fonts.

Any ideas? CC: @metajack @aneeshusa @edunham

aneeshusa:
How can I tell if the fonts are properly installed? I checked it out in a Vagrant VM and it seems to work OK: the fonts are getting downloaded and placed into /usr/share/fonts/truetype/msttcorefonts, and it looks like the debconf selection is getting set properly.

metajack:
When it fails nothing is created in /usr/share/fonts/truetype but no errors are reported and the debconf selections are all fine. You can look at the salt log in AWS boot or on the build machines for its details, but not much is reported there other than it successfully installed the package.

aneeshusa:
Hmm, interesting. I thinking looking at the logs will be important to figure this one out, but I don't want to just put them in a Gist or Pastebin. Would you be comfortable giving me SSH access to these machines? Also, which machines (hostnames) had this problem?

larsbergstrom:
Sure! I'd be happy to give you access to ssh in. You can open a PR like https://github.com/servo/saltfs/pull/184 to try it out.

IIRC, it happened to our EC2 "linux2" builder. I can give you that IP after we have your ssh key installed. Alternatively, I can also log in there and retrieve any logs that you need.

larsbergstrom:
Also, if it matters, I've noticed that if I `highstate` that box, I sometimes get:
```
----------
          ID: FIX enable multiverse
    Function: pkgrepo.absent
        Name: deb http://archive.ubuntu.com/ubuntu trusty multiverse
      Result: True
     Comment: Removed package repo deb http://archive.ubuntu.com/ubuntu trusty multiverse
     Started: 13:09:30.857022
    Duration: 6125.226 ms
     Changes:   
              ----------
              repo:
                  deb http://archive.ubuntu.com/ubuntu trusty multiverse
----------
          ID: enable multiverse
    Function: pkgrepo.managed
[...]

PR #250, opened 2016-03-14
Standardize pkgrepo usage

aneeshusa:
This changes our workaround to be more verbose with no-op states but
more robust in the case of pkrepo failures: the directory will not be
cleaned until after all of the pkgrepo states complete succesfully.

This also cleans up the MS TTF core fonts, and should help with #186.
Note that it appears the Vagrant `ubuntu/trusty64` image that we use
already has the `multiverse` repository enabled, which might be why
I can't reproduce #186 inside Vagrant; it may be worth spinning up a
temporary EC2 instance to test this out (probably in masterless/local
mode instead of having it connect to `servo-master`).

---
This change is [Review on Reviewable](https://reviewable.io/reviews/servo/saltfs/250)


edunham:
r+ 

Thanks for cleaning this up!

edunham:
Uh, someday I'll learn to syntax. 

@bors-servo r+

PR #253, opened 2016-03-15
Create user accounts instead of sharing root

edunham:
SSH best practices is to fully disallow remote root login. To do that, we have to start by getting everyone logging in as themselves. 

r? @aneeshusa

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/253)

### DIFF
--- a/common/init.sls
+++ b/common/init.sls
@@ -47,7 +47,7 @@ host-{{ hostname }}:
 {% for ssh_user in common.ssh_users %}
 {{ ssh_user }}:
   user.present:
-    - home: /home/{{ ssh_user }}
+    - createhome: True
     - optional_groups:
         - wheel
     - empty_password: True
--- a/common/init.sls
+++ b/common/init.sls
@@ -45,8 +45,15 @@ host-{{ hostname }}:
 {% endfor %}
 
 {% for ssh_user in common.ssh_users %}
+{{ ssh_user }}:
+  user.present:
+    - home: /home/{{ ssh_user }}
+    - optional_groups:
+        - wheel
+    - empty_password: True
+
 sshkey-{{ ssh_user }}:
   ssh_auth.present:
-    - user: root
+    - user: {{ ssh_user }}
     - source: salt://{{ tpldir }}/ssh/{{ ssh_user }}.pub
 {% endfor %}


asajeffrey:
Travis failures... https://travis-ci.org/servo/saltfs/jobs/116193785
```
----------

          ID: ajeffrey

    Function: user.present

      Result: False

     Comment: An exception occurred in this state: Traceback (most recent call last):

                File "/usr/local/Cellar/saltstack/2015.5.5_1/libexec/lib/python2.7/site-packages/salt/state.py", line 1561, in call

                  **cdata['kwargs'])

                File "/usr/local/Cellar/saltstack/2015.5.5_1/libexec/lib/python2.7/site-packages/salt/states/user.py", line 352, in present

                  __salt__['shadow.del_password'](name)

                File "/usr/local/Cellar/saltstack/2015.5.5_1/libexec/lib/python2.7/site-packages/salt/utils/lazy.py", line 87, in __getitem__

                  return self._dict[key]

              KeyError: 'shadow.del_password'

     Started: 18:49:29.585704

    Duration: 15.263 ms

     Changes:   

----------

          ID: sshkey-ajeffrey

    Function: ssh_auth.present

      Result: False

     Comment: Failed to add the ssh key. Is the home directory available, and/or does the key file exist?

     Started: 18:49:29.605247

    Duration: 14.908 ms

     Changes:   

----------
```

aneeshusa:
I've been waiting for this for so long :)

common/init.sls
```diff
@@ -45,8 +45,15 @@ host-{{ hostname }}:
 {% endfor %}
 
 {% for ssh_user in common.ssh_users %}
+{{ ssh_user }}:
+  user.present:
+    - home: /home/{{ ssh_user }}
```
aneeshusa: This won't work on the Macs, I believe we'll need to use /Users instead of /home there. We should do this via a Jinja variable that we pull in from the map.jinja file. I think this is also causing the Travis failure.

common/init.sls
```diff
@@ -45,8 +45,15 @@ host-{{ hostname }}:
...
+    - optional_groups:
+        - wheel
```
aneeshusa: Is this sufficient to provide `sudo` priveleges on OS X?

common/init.sls
```diff
@@ -45,8 +45,15 @@ host-{{ hostname }}:
...
+    - empty_password: True
```
aneeshusa: I'm ok with this for now (strictly better than status quo), but just wanted to leave a link to https://sourceforge.net/projects/pamsshagentauth/ as something I like to configure for password-less sudo that still has some security. I'm pretty sure there's nothing similar for OS X.

common/init.sls
```diff
@@ -45,8 +45,15 @@ host-{{ hostname }}:
...
+
 sshkey-{{ ssh_user }}:
   ssh_auth.present:
```
aneeshusa: We can combine this state with the previous one, so that they share the same ID.

aneeshusa:
After this lands, we should:

- update the wiki
- (probably as part of #254) add a state that removes root's authorized_keys file. It shouldn't matter with `PermitRootLogin no, but it's nice to have written down explicitly.

common/init.sls
edunham: AFAICT, OSX's BSD roots show through here and `wheel` does the usual thing. https://discussions.apple.com/thread/2210858?start=0&tstart=0

common/init.sls
edunham: As in going like this?
```
{{ssh_user}}
  user.present:                                                                 
    - createhome: True                                                          
    - optional_groups:                                                          
        - wheel                                                                 
    - empty_password: True
  ssh_auth.present:                                                             
    - user: {{ ssh_user }}                                                      
    - source: salt://{{ tpldir }}/ssh/{{ ssh_user }}.pub                
```
Sorry, still pretty new to Salt.

aneeshusa:
Hmm, it seems GitHub has removed the ability to comment on individual commits. FYI: I like to use `servo-macpro1` for testing OS X related things, since I can verify settings by hand afterwards.

common/init.sls
aneeshusa: Yup, that should work. Docs: https://docs.saltstack.com/en/2015.5/topics/tutorials/states_pt2.html#call-multiple-states

aneeshusa: Err, don't forget the colon after the {{ ssh_user }}: (in the id on the top line).

aneeshusa:
It looks like the Travis failures are because of the `empty_password`, which calls the `shadow.del_password` Salt function, which wasn't implemented for Mac until January of this year (and won't be available until Salt 2016.3). This module also implements setting passwords, so we can't set a (non-empty) password without it either.

We'll need to backport the module manually for now. Here's the docs if you want to give it a spin: https://docs.saltstack.com/en/2015.5/ref/modules/; let me know if you have questions.

aneeshusa:
Another alternative to backporting the `mac_shadow` module would be enabling passwordless sudo, so that the passwords on user accounts don't matter.

edunham:
Superseded by https://github.com/servo/saltfs/pull/628. Feedback from here is now applied there.

PR #259, opened 2016-03-21
NDK installation should run whenever it's not installed

edunham:
According to
https://docs.saltstack.com/en/latest/ref/states/all/salt.states.cmd.html#should-i-use-cmd-run-or-cmd-wait,
ndk installation will only try to run if the ndk was just downloaded.

In the old implementation, if you succeed at downloading the ndk but fail to
install it, subsequent highstates won't attempt to install the ndk because the
downloaded file will be unchanged.

With these changes, we decide whether to install the ndk based on whether the directory created by installation exists, rather than based on whether we downloaded the file this run.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/259)

### DIFF
--- a/android-dependencies.sls
+++ b/android-dependencies.sls
@@ -72,13 +72,12 @@ android-ndk:
       - file: android-ndk
 
 android-ndk-toolset-configuration:
-  cmd.wait:
+  cmd.run:
     - name: bash /home/servo/android-ndk-r10e/build/tools/make-standalone-toolchain.sh --platform=android-18 --toolchain=arm-linux-androideabi-4.8 --install-dir='/home/servo/ndk-toolchain' --ndk-dir='/home/servo/android-ndk-r10e'
     - user: servo
     - require:
       - cmd: android-sdk
-    - watch:
-      - cmd: android-ndk
+    - creates: /home/servo/ndk-toolchain
 
 /home/servo/.bash_profile:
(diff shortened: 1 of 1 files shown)


larsbergstrom:
@bors-servo retry
```
[0;31m----------[0m
    [0;31m      ID: android-dependencies[0m
    [0;31mFunction: pkg.installed[0m
    [0;31m  Result: False[0m
    [0;31m Comment: An error was encountered while installing package(s): W: Failed to fetch http://us.archive.ubuntu.com/ubuntu/dists/trusty-updates/main/binary-i386/Packages  Hash Sum mismatch
              
              W: Failed to fetch http://us.archive.ubuntu.com/ubuntu/dists/trusty-updates/universe/binary-i386/Packages  Hash Sum mismatch
              
              E: Some index files failed to download. They have been ignored, or old ones used instead.[0m
    [0;31m Started: 19:47:49.282790[0m
    [0;31mDuration: 6150.757 ms[0m
```

aneeshusa:
I'm almost ready to put in the PR I mentioned in #239 which touches the Android stuff heavily, and it's going to have merge conflicts with this/includes some of this work already. Can we hold off on this PR please?

Also, this isn't quite right - we still need to keep a requisite to `android-ndk` for ordering purposes being the biggest thing.

PR #260, opened 2016-03-21
updated version of the NDK

larsbergstrom:
r? @edunham 
I think this gets overridden by the env variables set by the buildbot master.cfg, but it was bugging me to have these incorrect.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/260)

Issue #209, opened 2016-02-02
Use Jinja templating for config file paths


aneeshusa:
This will help prevent issues like the one fixed in  #203.

PR #263, opened 2016-03-22
Android refactor

aneeshusa:
Main highlights:
  - Update to SHA512 hashes and HTTPS URLs
  - Use multiple directories + symlinks to be more robust during version updates for Android
  - Moves states around to make their purpose more clear

Helps with #209.
Supersedes #259, #260.

cc @larsbergstrom @edunham 

I recommend reviewing this commit by commit and reading the commit messages - feel free to ask questions.

Also, we should check that this doesn't break buildbot; I don't know how to do that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/263)


aneeshusa:
Also, the reason this is WIP is that we should version the URL that we download B2G from so I can give it a similar treatment as the Android states, preferably something like 

```
https://servo-rust.s3.amazonaws.com/B2G/v1/B2G.tgz
```

or

```
https://servo-rust.s3.amazonaws.com/B2G/v1/B2G-v1.tgz
```

aneeshusa:
(Also, if we're updating the B2G URL: I always find capital letters in URLs to look weird, so something like ```https://servo-rust.s3.amazonaws.com/b2g/v1/b2g-v1.tgz``` would be a nice-to-have. Disclaimer: I'm don't know about any branding for B2G.)

larsbergstrom:
I would not invest too much effort into the B2G rules - I suspect that once we have Aarch64 working (requires a servo fix before we can gate on it) there will be limited need to keep it running. Between Android & Arm32/Aarch64, we'll have most of the coverage we want for those platforms.

I don't know when we would remove it, but I can promise there won't be a v2 of the binaries :-)

aneeshusa:
Do we need to install the `servo-build-dependencies` on `servo-master`? I don't think it's running a buildbot instance, so I don't see why it would need the servo build dependencies.

larsbergstrom:
The only thing that I see in that list that we need is `git`. Good catch!

aneeshusa:
What would we need `git` for? I'm assuming for the buildbot master?

larsbergstrom:
@aneeshusa Maybe we're doing it wrong, but `servo-master` is where we have our "main" salt enlistment, and our process for doing the actual updates is after PRs land in this repo, we ssh to the master, `git pull origin master` in that directory and then do our `highstate`.

aneeshusa:
Salt actually has built-in support for using git as a fileserver backend: https://docs.saltstack.com/en/2015.5/topics/tutorials/gitfs.html.

Our current workflow is OK for now, so I'll add a state to install `git` on the master, and look at gitfs later.

aneeshusa:
Hmm, it looks like we're hitting the Travis log length limit due to how many files we extract: each extracted file shows up twice, once in the INFO logging and once at the end. How do you feel about dropping down the log level to warning? We'll miss out on all the lines starting with ```[INFO    ]```, but we should still be able to debug based on the highstate output at the end or locally with Vagrant.

larsbergstrom:
I think that dropping it down to warning is fine. I get the most out of seeing which task failed and what the error message was, personally.

edunham:
Reviewed 9 of 9 files at r1, 5 of 5 files at r2, 3 of 3 files at r3, 4 of 4 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, 2 of 2 files at r7, 1 of 1 files at r8.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/263)*

aneeshusa:
@edunham I needed to rebase to handle the fixup and squash commits anyways, so I rebased on top of the latest master.

edunham:
Reviewable did not handle that rebase very gracefully. Still good, though.

---

Reviewed 3 of 9 files at r1, 11 of 11 files at r9, 5 of 5 files at r10, 3 of 3 files at r11, 4 of 4 files at r12, 2 of 2 files at r13, 1 of 1 files at r14.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/servo/saltfs/263#-:-KDUfrF9-eOO3__L-OLJ:-1657613920)*

PR #268, opened 2016-03-23
Update ARM rootfs urls to nest version inside name

aneeshusa:
This allows updating each of the rootfs tarballs separately.

Note that the exact format of the URLs is open to some bikeshedding, but I recommend this for the following reasons:
  - Switching the version and the name (so that version is inside name) lets you update the tarballs separately
  - Keeping the version in the actual file name lets you download different versions without them clobbering each other on disk

cc @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/268)

### DIFF
--- a/servo-build-dependencies/arm.sls
+++ b/servo-build-dependencies/arm.sls
@@ -1,4 +1,5 @@
 {% from 'common/map.jinja' import common %}
+{% from tpldir ~ '/map.jinja' import arm %}
 
 arm-dependencies:
   pkg.installed:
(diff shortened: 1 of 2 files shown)


aneeshusa:
@larsbergstrom  Wait, did you update the URLs in the S3 bucket to match? I'm not seeing the updated URLs and it failed on Travis too; it only passed because of #274.

larsbergstrom:
hrm, I comletely misread ` https://servo-rust.s3.amazonaws.com/ARM/{{ target.download_name }}/{{ target.version }}/{{ target.download_name }}-{{ target.version }}.tgz`

Is that what we want? Doesn't `target.download_name` already have the .tgz on it? Urgh, I was so carefully checking that all of the removed bits were present in the new bits that I missed what the new bits actually said.

/me revokes own r+ privs

aneeshusa:
Sorry I wasn't more explicit about needing to update the S3 bucket!

You're right about the `.tgz` extension; I meant to do that and forgot to update it when I moved it over to the `map.jinja` file. Let me open another PR to fix it.

larsbergstrom:
@aneeshusa no need to apologize - I need to pay better attention. Thanks! We'll can avoid highstate'ing until the new PR lands and I *do* update the s3 bucket.

CC @edunham

Issue #264, opened 2016-03-22
Use Salt gitfs to keep master synced with repo


aneeshusa:
Docs: https://docs.saltstack.com/en/2015.5/topics/tutorials/gitfs.html

Salt has a built-in backend that uses a git repo (such as this one) to provide the files in the /srv/salt tree (the fileserver).

We should use it instead of manually `git pull`ing on the master.

metajack:
+1

Is this feature new? I don't remember seeing this back when we set this up.

aneeshusa:
This has been around for a long time, although I'm not sure exactly when it was added (at least as old as `2014.7`, possibly as old as `0.16.3` based on a quick skim of the docs).

I'm not yet sure about how this would interact with the workflow of iterating on changes directly on the master.

larsbergstrom:
If we don't have a local filesystem we can edit, what's the procedure for doing test `highstate` against the live system? There are a few pieces (e.g., the web hooks, buildbot, etc.) that we have no testing for and, in practice, I've needed to do and roll out 3 or 4 fixup tries before I actually get the original patch that was r+'d live.

cc @edunham

aneeshusa:
We can pass a command line flag to the `salt` command to point to a different configuration directory, which lets us set up test configurations that point at local files to use interactively. This way, regular `salt` commands continue to use the gitfs fileserver backend, and we can set up separate temporary directories for testing. I'd recommend putting such directories in `/tmp` so they don't stick around too long.

That being said, testing in production makes me uneasy and I'd prefer to build a better testing story for our configurations, both making testing inside Vagrant easier and adding more automated functional tests. It's just too easy to forget to pass a command line flag to Salt (I've done this more times on my own setup than I'd like to admit).

aneeshusa:
Hit the wrong button, whoops.

aneeshusa:
FYI, I've been looking into this for #350. The solution I mentioned above (custom config directories) [won't work](https://github.com/saltstack/salt/issues/25344), but I've found an alternative solution. See #350 for the details.

PR #271, opened 2016-03-23
Master needs a python git package for gitfs

edunham:
https://docs.saltstack.com/en/2015.5/topics/tutorials/gitfs.html says it's the preferred method. 

It looks like we're not managing any of the salt-master config within salt itself, so this dependency is the only change for #264

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/271)

### DIFF
--- a/git/init.sls
+++ b/git/init.sls
@@ -2,4 +2,4 @@
 git:
   pkg.installed:
     - pkgs:
-      - git
+      - python-git


aneeshusa:
Do we still need to install `git` explicitly, or will it get pulled in as a dependency or fulfilled by `python-pygit2`? I only added it to support the existing workflow of manually pulling, so let's remove it if it's not needed.

Also, I think it also makes sense to move this state to a new folder + file `salt/master.sls` instead, to make its purpose more clear, and get rid of the `git` folder.

aneeshusa:
Actually, the Salt docs note that `python-pygit2` isn't yet available in the Ubuntu repos, and Travis corroborates this.

edunham:
Might as well just use GitPython, then. 

> Ubuntu 14.04 LTS and Debian Wheezy (7.x) also have a compatible version packaged:
> $ sudo apt-get install python-git

Amended commit, and rebased onto current master branch to pull in the workaround for the Travis intermittents.

edunham:
Closing because I'm adding some other saltmaster-specific states and renaming the `git` directory for another PR, and I'll stick this change into there

PR #274, opened 2016-03-24
Allow failures on linux cross1 until travis fixes infra issues on GCE

larsbergstrom:
r? @edunham 

If this succeeds (I am always confused about the allow_failures syntax for this case), we can land all of the stuff that's failing due to the issue with the missing Ubuntu packages 'g++-arm-linux-gnueabihf' & 'g++-aarch64-linux-gnu' from the GCE instances.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/274)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,9 @@ matrix:
       os: linux
       sudo: required
       dist: trusty
-
+  allow_failures:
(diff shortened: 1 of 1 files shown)


edunham:
@bors-servo r+

We'll just have to be extra careful in manual review to examine the logs from linux-cross-1 till this change gets reverted.

Issue #275, opened 2016-03-24
Need to set up linux boxes to auto-update to stable packages


larsbergstrom:
I currently run `sudo dpkg-reconfigure --priority=low unattended-upgrades` on new Linux machines by hand (it has a GUI popup). Presumably there's a better way to set this up to happen automatically.

aneeshusa:
FYI, this is your third dupe of this issue :wink:
See #207 and #102.

I will say that I've never been a huge fan of automatic updates (see also: left-pad), preferring pinning + manual updating (possibly with automated rollout), but this is probably safe.

larsbergstrom:
@aneeshusa HA! Great catch. I keep using different words and never see it.

I tend to open these when I log into a machine and see "147 security updates needed," which always freaks me out. I'd personally like to have at least something basic that will get us security updates, even if it means we might left-pad our CI infra one day rather than leave machines unpatched or rely on manual updating.

aneeshusa:
Something else that might help is auditing the packages that are installed; I just checked `servo-master` and I'm seeing around 800 packages. This is something that traditional CM systems like Salt aren't so good at. For us, a simple `pkg.purged` state with a long list of `pkgs` will do, but better solutions (longer-term) would be starting from a more minimal Ubuntu image,  or using [NixOS](https://nixos.org/) or a similar declarative distribution, where only packages you explicitly list are installed (and packages you unlist get removed).

larsbergstrom:
I definitely agree! The `servo-master` machine is particularly bad, and I'm glad (though also terrified!) that @edunham and I plan to burn it down and rebuild with salt from scratch.

I believe that it's extra bad because in the dark, dark early days we all did our testing and part of our CI development directly on that machine. It's also survived major transitions (e.g., from the "bors" autolander to "homu," completely different methods of github events vs. polling tools, etc.).

Is the `pkg.purged` just supposed to have a list of all of the approved packages?

aneeshusa:
Well, `pkg.purged` will ensure those packages don't exist, so we would need to list any packages that we don't want installed (e.g. extra packages in the base image we're using). It's not perfect, but it should do for now.

As I mentioned in #215, look into using Salt multimaster to manage the `servo-master` transition; I'd also recommend setting up the new gitfs based configuration on a new master and cutting over once it's confirmed working, instead of deploying it on the existing one.

PR #276, opened 2016-03-24
Set unattended-upgrades on Ubuntu hosts

edunham:
Not sure if this is the best way, but http://blog.viraptor.info/post/75262988169/ubuntu-unattended-upgrades-the-salt-way suggests it.

Fixes https://github.com/servo/saltfs/issues/275

r? @aneeshusa

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/276)

### DIFF
--- a/common/init.sls
+++ b/common/init.sls
@@ -6,10 +6,11 @@ python2:
       - python
 
 {% if grains['os'] == 'Ubuntu' %}
-python2-dev:
+ubuntu-packages:
   pkg.installed:
     - pkgs:
       - python-dev
+      - unattended-upgrades
 
 unattended-upgrades:
   debconf.set:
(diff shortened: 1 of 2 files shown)


aneeshusa:
I don't think we actually need the `cmd.wait` here, the `debconf.set` should handle everything. We should check in Vagrant. Also FYI, `cmd.wait` has recently been deprecated, so we shouldn't use it in any new commits (I removed our last remaining uses during the Android refactor).

aneeshusa:
We'll also need a `pkg.installed` state to install the `unattended-upgrades` package, which should require the `debconf.set` state (we can nest them under the same ID).

aneeshusa:
Actually, I  think we'll need to replace the `cmd.wait` state with a `file.managed` state that does the same thing by putting the following into `/etc/apt/apt.conf.d/20auto-upgrades`:
```
APT::Periodic::Update-Package-Lists "1";
APT::Periodic::Unattended-Upgrade "1";
```

aneeshusa:
Do we want the machines to auto-reboot if necessary? I'm assuming not, but if so we'll need to do a little more to activate that feature.

edunham:
@aneeshusa I don't think that auto-rebooting would be a good idea right now. Latest commit implements the other 3 changes you suggested.

aneeshusa:
Didn't see you had updated this, sorry.

This is going to require some in Vagrant testing - I suspect the `unattended-upgrades` package has an associated service we'll need to start (due to #229).

common/init.sls
```diff
@@ -6,10 +6,24 @@ python2:
       - python
 
 {% if grains['os'] == 'Ubuntu' %}
-python2-dev:
+ubuntu-packages:
```
aneeshusa: I'd prefer to keep this state ID as just `python2-dev`, and add a new `pkg.installed` state under the `unattended-upgrades` ID to install the respective package. That lets us keep all of the unattended upgrades states together.

aneeshusa: Also, the new pkg.installed state should `require` the debconf state for ordering purposes.

PR #279, opened 2016-03-26
rust bindgen support

larsbergstrom:
r? @Manishearth @edunham 

I'm still not sure of the best way to handle the fact that we need to hand out a GH webhook secret to arbitrary people to set this up. I suspect we will need to set up homu.servo.org at some point.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/279)

### DIFF
--- a/homu/cfg.toml
+++ b/homu/cfg.toml
@@ -15,6 +15,7 @@ port = 54856
 secret = "{{ pillar["homu"]["web-secret"] }}"
 
 {% set travis_repos = [
+    ('crabtw', 'rust-bindgen'),
     ('servo', 'app_units'),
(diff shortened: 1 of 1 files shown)


aneeshusa:
Do we hand out the same GH webhook secret for every repo, or do we generate a new one for each repo?

nox:
Misclick. -_-

larsbergstrom:
@aneeshusa We only have one hook right now.

We could create and store new tokens for each repo (homu definitely supports per-repo tokens!), but I think I'd really like to get it installed and integrated as a first-class app via GitHub so that interaction through salt is not required to add a new repo.

PR #290, opened 2016-04-02
Update android-18 menu entry to 8 for SDK r24

aneeshusa:
We're ignoring Travis builds for the cross builder and this fix needs to be verified in Vagrant (this VM takes a while to provision unfortunately). We may want to actually try running a build in Vagrant to see if this works.

We updated to Android SDK r24 in November from r23, but did not update
the menu entry to install the android-18 target. The new number comes
from the output of `./android sdk list`.

We need to delete the `/home/servo/android/sdk/r24.4.1` folder before applying this fix to the builders. Ideally this would get checked for by Salt, but currently you can only specify one file for the ` - creates: ` key of `cmd.run`, and our state creates the `platform-tools` dir and the `android-18` dir. I'll open a Salt issue for this.

cc @larsbergstrom, https://github.com/servo/servo/issues/10339

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/290)

### DIFF
--- a/servo-build-dependencies/android.sls
+++ b/servo-build-dependencies/android.sls
@@ -47,14 +47,10 @@ android-sdk:
     - require:
       - user: servo
   cmd.run:
-    # The arguments to --filter are from running 'android list sdk'
-    # Currently these are:
-    #   platform-tool: Android SDK Platform-tools, revision 23.0.1
-    #   9: SDK Platform Android 4.3.1, API 18, revision 3
     - name: |
         expect -c '
         set timeout -1;
-        spawn {{ common.servo_home }}/android/sdk/{{ android.sdk.version }}/android-sdk-linux/tools/android - update sdk --no-ui --filter platform-tool,9;
+        spawn {{ common.servo_home }}/android/sdk/{{ android.sdk.version }}/android-sdk-linux/tools/android - update sdk --no-ui --filter platform-tool,android-18;
         expect {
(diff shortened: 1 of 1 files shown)


aneeshusa:
Hmm, this seems to have installed `android-20` in my Vagrant machine. :(

aneeshusa:
Update: newest version is much simpler and seems to work properly in Vagrant. Can't believe I never tried this before.

larsbergstrom:
Oh, nice! Is this ready to go?

aneeshusa:
Yup, this is ready to get reviewed - make sure to double check that it installs `android-18` properly in Vagrant.

aneeshusa:
Ping on this; `android-nightly` was failing to compile previously (I think due to an LLVM bug), but a recent rustup means compilation now succeeds and this is again a blocker for sucessful android-nightly runs.

aneeshusa:
Heads up - I believe this will require some manual intervention to get Salt to apply the changes. Deleting the `/home/servo/android/sdk/r24.4.1` path before running a highstate should do the trick.
(The issue is that we only check for the platform tools being installed, not the android-18 SDK, with the `creates` parameter.)
I'll file a follow-up PR to make this happen automatically.

larsbergstrom:
Thanks! Indeed, I rolled this out manually after landing it and noticing that highstate thought that nothing needed to be modified :-)

aneeshusa:
See #383 for the follow-up PR.

PR #291, opened 2016-04-03
Re-engage servo-linux-cross1 on Travis

aneeshusa:
Upstream has released new packages for Ubuntu Trusty which resolve
the installation conflicts encountered when installing the ARM g++ packages.

See https://bugs.launchpad.net/ubuntu/+source/gcc-4.8-armhf-cross/+bug/1557205/comments/16
for a comment confirming release of the upstream fix.

This reverts commit aa8435536091a125494786997ab3c408dc3b3386 which was part of #274.

Closes #280.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/291)

Issue #287, opened 2016-03-30
Don't hard-code /etc/hosts entries


aneeshusa:
The `common/init.sls` file adds a lot of entries to `/etc/hosts`. We should use DNS for name resolution instead.

@edunham reminded me of this by mentioning 'messy hacks around not trusting DNS' in IRC, so if there are any other hacks we should get rid of those too.

larsbergstrom:
Yeah, @edunham ran into this during the move.

I think those predate her (awesome) move of us to CloudFlare and DNS self-management. It used to take a couple of days to get a DNS entry changed via internal corporate IT DNS management, so that's clearly not needed anymore :-)

aneeshusa:
I can't read my IRC client, it was @edunham that mentioned it! My apologies.

Did you run into any other issues during the move?

aneeshusa:
Closing in favor of #288.

PR #292, opened 2016-04-03
Target servo-master\d+ with PCRE in the top file

aneeshusa:
This fixes the targeting for servo-master\d+ so that the extra states
are actually applied.

I noticed this while reviewing the travis builds for #291.
This was missed in both #284 and #285!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/292)

### DIFF
--- a/.travis/test_pillars/top.sls
+++ b/.travis/test_pillars/top.sls
@@ -3,7 +3,8 @@ base:
     - travis
     - buildbot.common
 
-  'servo-master':
+  'servo-master\d+':
+    - match: pcre
     - buildbot.master
     - homu
 
--- a/top.sls
+++ b/top.sls
@@ -30,6 +30,7 @@ base:
     - xvfb
 
   'servo-master\d+':
+    - match: pcre
     - git
     - buildbot.master
     - homu


larsbergstrom:
r? @edunham 

I think there were a couple of these edits you were also looking at after the moveover?

larsbergstrom:
Namely, does this also need:
```
root@servo-master1:/srv/salt# git diff
diff --git a/common/map.jinja b/common/map.jinja
index 52f99f4..c7d6239 100644
--- a/common/map.jinja
+++ b/common/map.jinja
@@ -3,7 +3,7 @@
       'defaults': {
         'servo_home': '/home/servo',
         'hosts': {
-          'servo-master': '96.126.125.232',
+          'servo-master1': '52.37.76.55',
           'servo-linux1': '66.228.48.56',
           'servo-linux2': '173.255.201.95',
           'servo-linux3': '45.79.167.177',
```

aneeshusa:
I asked about it before, but I think now we're actually trying to get rid of that entire list (see #287). If anything is using those hosts entries for lookups instead of DNS, it's a bug.

PR #298, opened 2016-04-05
Fix Vagrantfile after adding test suite.

aneeshusa:
The install_salt.sh script was moved, so use the new name,
and filter out the 'test' node when making the list of VMs.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/298)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -28,11 +28,13 @@ Vagrant.configure(2) do |config|
         { id: extract_id(node['env']), os: node['os'], box: 'ubuntu/trusty64' }
       end
     end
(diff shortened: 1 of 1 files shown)


aneeshusa:
Sadly I don't think it's feasible to test the Vagrant configuration on Travis :(

aneeshusa:
I'd appreciate if someone could review this - it would make testing a lot easier, especially when doing heavy rebasing.

edunham:
Tested locally and it looks good to me. 

@bors-servo r+

PR #299, opened 2016-04-05
Add check for Python 3.5+ in test script

aneeshusa:
The test script uses features introduced in Python 3.5 (namely
os.path.commonpath), but the shebang only specifies python3. Because
Python3.5 is fairly new, it's likely that python3 is not yet Python 3.5
(or later) for many people, so explicitly check for this.

This occured on Travis, requiring an explicit 'python: 3.5' config,
which is what spurred this commit.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/299)

### DIFF
--- a/test.py
+++ b/test.py
@@ -44,11 +44,17 @@ def run_tests(tests):
 
 
 def main():
+    if sys.version_info < (3, 5):  # We use features introduced in Python 3.5
+        sys.stderr.write('{}: Python 3.5 or later is needed for this script\n'
+                         .format(__file__))
+        return 1
+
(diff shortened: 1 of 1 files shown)


aneeshusa:
Rebased.

edunham:
@bors-servo r+

Relatedly, ought we be running `test.py` in Travis?

aneeshusa:
@edunham test.py is getting run on Travis, via the [`dispatch.sh` script](https://github.com/servo/saltfs/blob/master/.travis/dispatch.sh).

PR #301, opened 2016-04-06
Disable appveyor due to perf and instability

larsbergstrom:
r? @edunham @aneeshusa 

We're having WAY too many problems with appveyor right now and need to get it better before I can leave it on.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/301)


aneeshusa:
Would it be valuable to open a tracking issue for Appveyor bugs?

PR #308, opened 2016-04-07
Add test to check if Nginx is serving requests

aneeshusa:
urllib's error handling is not very helpful for this test, but I'm
using it in an effort to avoid extra dependencies. This only checks
if Nginx is alive and returning requests, not whether the HTTP status
codes indicate success. The reason for this is that the test pillars
we have in the repo are not valid, causing Buildbot and Homu to die
after starting, and thus Nginx returns 502 Bad Gateway errors when it
tries to reverse proxy requests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/308)

### DIFF
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -23,6 +23,6 @@ else
 
     # TODO: don't hard-code this
     if [ "${SALT_NODE_ID}" = "servo-master1" ]; then
-        ./test.py sls.buildbot.master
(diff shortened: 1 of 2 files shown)


edunham:
Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/308)*

PR #314, opened 2016-04-11
Update Salt Homebrew install URL for repo split

aneeshusa:
The Homebrew/homebrew repository has recently been split into two
separate repositories, Homebrew/brew for the package manager itself
and Homebrew/homebrew-core for the formulas (packages).
See this homebrew-discuss mailing list post announcing this change:
https://groups.google.com/forum/#!msg/homebrew-discuss/D_weS6pZesg/jLgZj6mUCQAJ

The old raw URL we were using seems to still be available, but it seems
likely this is just due to caching by GitHub and it may disappear at
any time.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/314)

### DIFF
--- a/.travis/install_salt.sh
+++ b/.travis/install_salt.sh
@@ -17,7 +17,7 @@ install_salt () {
     elif [ "${OS_NAME}" = "osx" ]; then
         printf "$0: installing salt for Mac OS X\n"
         brew update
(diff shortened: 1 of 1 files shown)


edunham:
fun, thanks for catching that before it broke us!

@bors-servo r+

PR #307, opened 2016-04-07
Add test to check Buildbot master config

aneeshusa:
A few things in this PR (in separate commits):
  - Refactor the Buildbot states to standard paths, use more Jinja templating
  - Switch the Buildbot master service to not restart automatically. **It will need to be restarted manually if there are configuration changes.** This should alleviate the need for `state.apply`.
  - Use the `buildbot checkconfig` command to confirm the syntax of the Buildbot master configuration, and run it on Travis. This can be run via `./test.py sls.buildbot.master`, but note that this won't work locally - it needs to be run in an environment where the Buildbot master states have run, i.e. Travis or Vagrant.

cc @larsbergstrom @edunham

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/307)

Issue #302, opened 2016-04-06
Set up Salt to configure a Windows buildbot machine


larsbergstrom:
It appears that there is a salt client:
https://repo.saltstack.com/#windows

We basically need to set it up per https://github.com/servo/servo/#prerequisites:
```
On Windows:

Download Python for Windows here. This is required for the SpiderMonkey build on Windows.

Install MSYS2 from here. After you have done so, open an MSYS shell window and update the core libraries and install new packages:

update-core
pacman -Sy git mingw-w64-x86_64-toolchain mingw-w64-x86_64-freetype \
    mingw-w64-x86_64-icu mingw-w64-x86_64-nspr mingw-w64-x86_64-ca-certificates \
    mingw-w64-x86_64-expat mingw-w64-x86_64-cmake tar diffutils patch \
    patchutils make python2-setuptools
easy_install-2.7 pip virtualenv

Open a new MSYS shell window as Administrator and remove the Python binaries (they are not compatible with our mach driver script yet, unfortunately):

cd /mingw64/bin
mv python2.exe python2-mingw64.exe
mv python2.7.exe python2.7-mingw64.exe

Now, open a MINGW64 (not MSYS!) shell window, and you should be able to build servo as usual!
```

And run commands like:
https://github.com/servo/servo/blob/master/appveyor.yml

larsbergstrom:
This is going to be substantially more complicated than expected. Following the steps here:
http://trac.buildbot.net/wiki/RunningBuildbotOnWindows

Required an *incredible* amount of manual effort.

PR #320, opened 2016-04-16
Initial Windows support

larsbergstrom:
r? @edunham @aneeshusa 

I don't expect it to succeed yet (windows was broken by a recent servo commit), so homu is not gated on it. But, like arm64, I'd like to get it building regularly so that I can see the output.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/320)

### DIFF
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -11,6 +11,7 @@ SERVO_REPO = "https://github.com/servo/servo"
 LINUX_RESERVED_SLAVES = ["servo-linux1", "servo-linux2"]
 MAC_SLAVES = ["servo-mac1", "servo-mac2", "servo-mac3"]
 CROSS_SLAVES = ["servo-linux-cross1", "servo-linux-cross2"]
+WINDOWS_SLAVES = ["servo-windows1"]
 
 c = BuildmasterConfig = {}
 c['protocols'] = {'pb': {'port': 9001}}
@@ -23,7 +24,7 @@ c['caches'] = {
 ####### BUILDSLAVES
 
 c['slaves'] = []
-for s in MAC_SLAVES + CROSS_SLAVES + LINUX_RESERVED_SLAVES:
+for s in MAC_SLAVES + CROSS_SLAVES + LINUX_RESERVED_SLAVES + WINDOWS_SLAVES:
     c['slaves'].append(buildslave.BuildSlave(s, SLAVE_PASSWORD, max_builds=1))
 
 ####### CHANGESOURCES
@@ -47,7 +48,7 @@ c['schedulers'] = []
 c['schedulers'].append(schedulers.AnyBranchScheduler(
     name="servo-auto",
     treeStableTimer=None,
-    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "arm32", "arm64"],
+    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "arm32", "arm64", "windows"],
     change_filter=util.ChangeFilter(filter_fn=servo_auto_try_filter),
 ))
 c['schedulers'].append(schedulers.SingleBranchScheduler(
@@ -58,7 +59,7 @@ c['schedulers'].append(schedulers.SingleBranchScheduler(
 ))
 c['schedulers'].append(schedulers.ForceScheduler(
     name="force",
-    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "arm32", "arm64", "android-nightly"]
+    builderNames=["linux-dev", "linux-rel", "mac-rel-wpt", "mac-dev-unit", "mac-rel-css", "android", "gonk", "arm32", "arm64", "android-nightly", "windows"]
 ))
 c['schedulers'].append(schedulers.Nightly(
     name="Nightly",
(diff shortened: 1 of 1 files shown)


buildbot/master/files/config/master.cfg
```diff
@@ -231,6 +232,16 @@ arm64_factory = create_servo_factory([
     steps.Compile(command=["./mach", "build", "--rel", "--target=aarch64-unknown-linux-gnu"], env=arm64_compile_env),
 ])
 
+windows_compile_env = dict({'PATH': 'C:\\msys64\\mingw64\\bin;C:\\msys64\\usr\\bin\\C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\;C:\\Program Files\\Amazon\\cfn-bootstrap\\',
```
aneeshusa: Let's use a raw string to make this nicer.

aneeshusa:
Just curious: did you make any progress with #302?

buildbot/master/files/config/master.cfg
```diff
@@ -232,7 +232,7 @@ arm64_factory = create_servo_factory([
     steps.Compile(command=["./mach", "build", "--rel", "--target=aarch64-unknown-linux-gnu"], env=arm64_compile_env),
 ])
 
-windows_compile_env = dict({'PATH': 'C:\\msys64\\mingw64\\bin;C:\\msys64\\usr\\bin\\C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\;C:\\Program Files\\Amazon\\cfn-bootstrap\\',
+windows_compile_env = dict({'PATH': r'C:\msys64\mingw64\bin;C:\msys64\usr\bin\C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Program Files\Amazon\cfn-bootstrap',
```
larsbergstrom: @aneeshusa like this?

larsbergstrom:
And re: #302, it's going to be WAY harder than expected. buildbot requires an incredibly manual setup process (involving a bunch of manual registry key edits, a few aborted service start/restarts to create initial keys with proper permissions, etc.), so I just muscled my way through the setup manually.

When we go to stand up a second machine and we're not in urgent-Windows-mode, I'd like to look at it again, because nobody should have to go through that manually.

buildbot/master/files/config/master.cfg
aneeshusa: Yes, much easier to work with. (For some reason GitHub highlights these weird, though.)

Did you remove the slash at the end on purpose?

aneeshusa:
I figured as much, that sounds painful. Which version of Windows is this machine running? I think that would be valuable to document (on the wiki).

larsbergstrom:
Will do! I went with Windows Server 2012. I like that it's got a longer support cycle, but the lack of a start button is incredibly disturbing :-)

buildbot/master/files/config/master.cfg
```diff
@@ -231,6 +232,16 @@ arm64_factory = create_servo_factory([
     steps.Compile(command=["./mach", "build", "--rel", "--target=aarch64-unknown-linux-gnu"], env=arm64_compile_env),
 ])
 
+windows_compile_env = dict({'PATH': r'C:\msys64\mingw64\bin;C:\msys64\usr\bin\C:\Windows\system32;C:\Windows;C:\Windows\System32\Wbem;C:\Windows\System32\WindowsPowerShell\v1.0\;C:\Program Files\Amazon\cfn-bootstrap',
+                            'MSYSTEM': 'MINGW64',
+                            'MSYS': 'winsymlinks=lnk'},
+                           **common_test_env)
```
aneeshusa: What about the `CARGO_HOME` and `SERVO_CACHE_DIR` environment variables we set for Linux and OS X?

(I think there's a port of ccache to windows, but we can worry about that later.)

larsbergstrom: I'm skipping them because we only have one build type, at least for now, and I'm trying to get something that works similarly to the current AppVeyor config (https://github.com/servo/servo/blob/master/appveyor.yml).

I really wish that we could use ccache, but it doesn't work on Windows due to known build problems with the components we pick up from Firefox. I already spent a bunch of time trying to get it working (and failing) on AppVeyor and locally.

aneeshusa: Do we need something similar to https://github.com/servo/servo/pull/10528?

larsbergstrom: Luckily, I have 100GB of disk on this machine :-) AppVeyor's cache is only 500MB.

That said, we have *no* cache reduction on any of the machines, so they will all slowly accrete more cargo packages & rust/cargo binaries over time until they fill all available disk. Opening #321

aneeshusa:
This LGTM after a squash.

Side note: As far as I can tell we haven't actually deployed any of the last 3-4 PRs to our buildbot machines. I'm sure you and @edunham both have enough on your plates already; the Homu queue only has one build right now, so can you show me exactly how to restart Buildbot cleanly to deploy this when it finishes?

larsbergstrom:
@aneeshusa I typically just wait for the queue to empty, log in, and then `highstate`. I've found that the "careful" buildbot restarts that we described have a downside that we end up with a buildbot master process that is running as an orphaned child process instead of as a child of the service framework, so I don't use it very often.

I *think* that if we change the salt rules to push out the files but only send a `SIGHUP` to a running buildbot process, that should have the best effect, and could be integrated with normal salt rules.

aneeshusa:
OK, can I try deploying these changes? I'm logged into a tmux instance on `servo-master1` if you want to watch.

larsbergstrom:
@aneeshusa Sure! Please go right ahead, and thanks for all of the late-Friday-night reviews & chatter :-)

aneeshusa:
Just deployed. BTW, I turned off the automatic buildbot-master restart in #307, so I also restarted it manually.

I'm running `tail +F twistd.log` in tmux and I'm seeing a lot of this:
```
invalid login from unknown user 'servo-windows1'
```
As well as the same message for `servo-macpro1`. I think we're missing some authentication?

larsbergstrom:
I know that macpro1 should not be working, but `servo-windows1` should be working because of this:
https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/master.cfg#L27-L28

I'm not sure what's going wrong here...

larsbergstrom:
Hrm, I re-ran the steps here:
https://github.com/servo/servo/wiki/Buildbot-administration#dealing-with-troubles

And now it's seeing `servo-windows1`. However, the buildbot process seems to be running as root instead of as the servo user.

aneeshusa:
Which steps in particular did you run? It's weird that a single restart didn't work properly.

larsbergstrom:
I attempted to run:
```
# su - servo
# buildbot restart --clean --nodaemon /home/servo/buildbot/master &
```
But it told me that the .pid file was not owned by the servo user. So I just ran it as root and the service turned over nicely :-)

aneeshusa:
Also, I think something may be wrong with our `buildbot-master.conf` file. `service buildbot-master status` says that it's not running, and checking in syslog says it was terminated with status 1, so maybe it's being invoked incorrectly?

PR #324, opened 2016-04-16
Do not use linux home path on macos

larsbergstrom:
r? @aneeshusa 

How can I get `common.Darwin.servo_home`? Right now, in these lines we're getting the linux ones, which breaks the build. I tried using that string above, but it didn't work. I've deployed this version just to get the MacOS builders off the floor.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/324)


aneeshusa:
@bors-servo r+

Deploying this for now to keep the repo in sync with prod. Buildbot is a bit weird in that we need to know details about the worker environment on the master - I'll have to think a little bit about how to do this cleanly.

Issue #316, opened 2016-04-12
Move buildbot logic to scripts in `/etc/ci`


edunham:
Possibly of interest to @shinglyu. 

Right now a lot of build logic is hard-coded in https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/master.cfg . Ideally, Buildbot would only invoke scripts from https://github.com/servo/servo/tree/master/etc/ci. 

It looks like the only templating in `master.cfg` is to insert the value of `common.servo_home`, which is set in https://github.com/servo/saltfs/blob/master/common/map.jinja because it differs between Mac and Linux.

to deploy:

- [x] PR the build scripts to `servo/servo`; land them. servo/servo#10849
- [ ] PR `servo/saltfs` changing buildbot config to invoke the build scripts
- [ ] @larsbergstrom and I will test the buildbot changes on the buildmaster to make sure they work
- [ ] land saltfs PR & highstate servo-master1

Ms2ger:
How about the "steps"?

shinglyu:
Cool, I'll take a look next monday

Ms2ger <notifications@github.com> 於 2016年4月16日 週六 下午3:37寫道：

aneeshusa:
Because we're executing arbitrary code from the git checkout, I don't feel comfortable moving the existing build scripts into the `servo` repo for two reasons: they depend on the `buildbot` python package which is not installed on the workers and has no need to be installed, and I'd rather not execute arbitrary Python code which will get injected back into our Buildbot process.

@Ms2ger I saw [your comment about keeping the "steps" in ](https://github.com/servo/servo/issues/10337#issuecomment-205307332) and I started looking into it then. It can be done, but it's not quite as simple as @edunham's checklist makes out. Buildbot coordinates everything on the master and runs 'remote' shell commands on the workers, so we need a step that dynamically creates new steps from the git checkout; just running a script on the worker will make everything look like one step.

I've prototyped a better solution for this, but it's a little hacky to achieve the dynamic steps behavior. (Buildbot Nine has explicit support for this though, so I feel comfortable using it as it appears to be sanctioned.)
@shinglyu I started working on this before this issue got opened but I don't want to step on your toes; I can push what I have already to a branch so you can take a look and finish it up.

shinglyu:
[...]

PR #310, opened 2016-04-08
Create error summary logs when running wptrunner-based tests

jgraham:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/310)

PR #325, opened 2016-04-18
Use yaml for buildbot steps

aneeshusa:
A bunch of changes here including various removals and rearrangements; I'd recommend checking this commit by commit (including commit messages).

This a) simplifies the Buildbot master configuration and b) uses YAML to configure build steps instead of Python code. This allows us to PR the YAML file over to the main servo repo for #316; I'd like to see this landed before working on the rest of #316.

cc @edunham @shinglyu

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/325)


jdm:
Having not looked at any of the rest of the implmentation, that's a really nice YAML configuration file!

aneeshusa:
A note I left myself yesterday:

We don't need the `GITHUB_STATUS_TOKEN` pillar value anymore, so when this lands, we need to do the following:

- [ ] Revoke the token (if it's still valid) (unless it's getting used for homu...)
- [ ] Remove the value from the production pillar

aneeshusa:
Also, to check that this works properly you can rename `master.cfg` to `master.py`, then in a Python2 repl `import master` and peruse the `master.c['builders']` object - I found this helpful while testing.

larsbergstrom:
Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, 1 of 1 files at r7, 4 of 4 files at r8, 2 of 2 files at r9, 3 of 3 files at r10, 2 of 2 files at r11.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/servo/saltfs/325#-KFf3LUy2-gQcM8wqsqa:-KFf3LUy2-gQcM8wqsqb:-686577304) (no related file):*
Can you please remove the `pkill` commit? It hasn't done so in a while, but it catches when `servo` processes spin out of control occasionally, so I'm loathe to remove it until we do something like reboot or image staging per-PR.

---

*[a discussion](https://reviewable.io:443/reviews/servo/saltfs/325#-KFf3alCw6AKL05wXMVK:-KFf3alCw6AKL05wXMVL:1107554539) (no related file):*
I think we need to keep the "duplicate" tests for manifest/lock file changes. They need to be run multiple times, because cargo can update different Cargo.lock files differently on different platforms and release vs. debug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/325)*

larsbergstrom:
This is an awesome PR! @edunham and I just spent some time reviewing it line by line and were in awe of how much cleaner everything will be after this. Nice work!

edunham:
These changes are beautiful, and were uniquely easy to review because of how well you structured your commits. Now our only changes are lines to add back into `steps.yml`, as Lars has mentioned: 

 * We do still need to `pkill -x servo` at the start or end of every build, or the platform-specific equivalent, because sometimes unusual test failures leave servo processes lying around. 
* We'd prefer to keep running the "duplicate" tests for now, to catch subtle differences between platforms

---

Reviewed 1 of 1 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, 2 of 2 files at r4, 2 of 2 files at r5, 3 of 3 files at r6, 1 of 1 files at r7, 4 of 4 files at r8, 2 of 2 files at r9, 3 of 3 files at r10, 2 of 2 files at r11.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/325#-:-KFf4DOLqYb0NUS1wlNQ:1262783618)*

aneeshusa:
Thanks, I'm a stickler for using git commits + messages effectively :)

Quick question about keeping the duplicates - the only two tests that should be duplicated are `lockfile_changed.sh` and `manifest_changed.sh`, not `test-tidy` or `check_no_unwrap.sh`, correct? Also, `lockfile_changed.sh` just runs `git diff`, so why would its results be platform specific?

larsbergstrom:
Yes, just lockfile/manifest_changed.

The Cargo.lock file contains a list of all of the crates and versions used across all platforms. But, in a given platform-specific build, only a subset of those might get touched. Further, different platforms (cef, stylo, servo, and gonk) even have their own different Cargo.lock files that are *only* touched on those builds! That check tends to catch errors on builds that are not performed by the original author (e.g., MacOS package updates for people who only have linux).

Manishearth:
This seems to get rid of the `logfiles` stuff we had? Perhaps we should add an extra `logfiles` key to the YAML?

aneeshusa:
@Manishearth the DynamicServoFactory.make_step method uses a little magic - sorry, heuristics :wink: - to automatically populate the logfiles parameter when necessary by looking for `--log-something` arguments to `./mach`. The logfiles parameter is easy to miss - see #310 for an example - but our usage of `mozlog` means this catches every important logfile so far.

In the future, we can make `make_step` smarter to handle explicit configuration of logfiles and such in the YAML file fairly easily, but I'd prefer to avoid that until we really need it.

aneeshusa:
@larsbergstrom should we be running the lockfile and manifest checks on all the other builders as well, e.g. `arm32` and `arm64` and `android`?

Manishearth:
Yes, because each platform might pull in platform-specific packages which won't have been updated in the committer's lockfile.

aneeshusa:
What's the equivalent of `pkill` for Windows?

aneeshusa:
Added back pkill (note, it's first removed then added back in a later commit), and updated the steps.yml file.

I also added a few TODO comments to do both of those things for Windows as well.

larsbergstrom:
@aneeshusa `taskkill`, https://technet.microsoft.com/en-us/library/bb491009.aspx

I think we want `taskkill /im servo.exe`, but will check it out tomorrow when I'm in front of my Windows VM.

larsbergstrom:
But yes, feel free to skip it on Windows for now!

aneeshusa:
Fixed the build error.

edunham:
Reviewable just made me go through every commit again, but showed everything unchanged from before you rebased as being empty diffs. Still looks good, though :)

---

Reviewed 9 of 9 files at r12, 1 of 1 files at r13, 1 of 1 files at r14, 2 of 2 files at r15, 2 of 2 files at r16, 1 of 1 files at r17, 3 of 3 files at r18, 1 of 1 files at r19, 4 of 4 files at r20, 2 of 2 files at r21, 2 of 2 files at r22, 8 of 8 files at r23, 1 of 1 files at r24, 3 of 3 files at r25, 1 of 1 files at r26, 4 of 4 files at r27, 2 of 2 files at r28, 2 of 2 files at r29, 1 of 1 files at r30.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/325#-:-KFgJIAtzG-_CYsKRIbo:-1361344851)*

shinglyu:
Do we need to quote the commands in the yaml file? We might have some steps that has special characters that may be parsed incorrectly?

Manishearth:
Yaml is pretty good that way, it uses newlines so you rarely need to quote

aneeshusa:
Rebased again; sorry to make you sit through another Reviewable session!

larsbergstrom:
@edunham I think this is probably good to go, right? At least, reviewable is just telling me it's a rebase with no edits, and I *think* everything is signed off...

edunham:
@larsbergstrom I agree, let's land it now and then test carefully. 

@bors-servo r+

PR #328, opened 2016-04-19
Add nox key

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/328)


aneeshusa:
@nox You also need to add your name to the list of `ssh_users` in the `common/map.jinja` file: https://github.com/servo/servo/wiki/Buildbot-administration#buildbot-builders

PR #331, opened 2016-04-21
Fix building on Windows

larsbergstrom:
r? @edunham 

This *appears* to fix the Windows build, when I've tested the environment vars and the command via RDP to the build machine. Here's hoping!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/331)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -114,13 +114,20 @@ doc = ServoFactory([
                        logEnviron=False),
 ])
 
+
+def make_cd_string(command):
+    return "cd /c/buildbot/slave/windows/build ; " + command
+
+
 windows = ServoFactory([
     # TODO: convert this to use DynamicServoFactory
     # We need to run each command in a bash login shell, which breaks the
     # heuristics used by DynamicServoFactory.make_step
-    steps.Compile(command=["bash", "-l", "-c", "./mach build -d -v"],
+    steps.Compile(command=["bash", "-l", "-c",
+                           make_cd_string("./mach build -d -v")],
                   env=envs.build_windows),
-    steps.Compile(command=["bash", "-l", "-c", "./mach test-unit"],
+    steps.Compile(command=["bash", "-l", "-c",
+                           make_cd_string("./mach test-unit")],
                   env=envs.build_windows),
     # TODO: run lockfile_changed.sh and manifest_changed.sh scripts
 ])


aneeshusa:
It may be more useful to put this in a .bashrc file instead, so it doesn't need to be repeated for every command.

larsbergstrom:
Put the `cd` into `.bashrc`? I'm not sure I'd want to do that, as at some point we should have windows-dev and windows-rel builders (and possibly split for different tests), each of which would have a different buildbot directory.

aneeshusa:
OK, that makes sense (we could do things with environment variables and .bashrc but that sounds like a bad idea). You might want to make a small function to build a full windows command array from just the './mach build' string of interest to avoid hitting the line length limit.

larsbergstrom:
Is the changed patch more of what you're looking for?

buildbot/master/files/config/factories.py
```diff
@@ -114,13 +114,16 @@ def make_step(self, command):
...
 
+def make_cd_string(command):
+    return "cd /c/buildbot/slave/windows/build ; " + command
+
 windows = ServoFactory([
     # TODO: convert this to use DynamicServoFactory
     # We need to run each command in a bash login shell, which breaks the
     # heuristics used by DynamicServoFactory.make_step
-    steps.Compile(command=["bash", "-l", "-c", "./mach build -d -v"],
+    steps.Compile(command=["bash", "-l", "-c", ""], make_cd_string("./mach build -d -v"),
```
jdm: This looks wrong compared to the next change.

larsbergstrom: Good catch! I hate that it is basically impossible to test any of this locally.

larsbergstrom:
That should hopefully be less worse.

larsbergstrom:
OK, it even passes Travis now :-)

buildbot/master/files/config/factories.py
aneeshusa: Actually, this could have been caught via `./test.py sls.buildbot.master` inside Vagrant - there's some extra setup that's required but I've been using it and it's pretty helpful. I'm trying to add local tests where I can :)

aneeshusa:
This looks OK, but I was thinking that the function could create the entire
```
["bash", "-l", "-c", "cd /c/buildbot/slave/windows/build; ./mach build --dev --verbose"]
```
array, instead of just the command at the end.

larsbergstrom:
@bors-servo r=aneeshusa

Sounds good; I will open an `E-Easy` to make that change.

PR #341, opened 2016-04-26
Run a subset of test-wpt with the --multiprocess option.

asajeffrey:
This is to check for regressions in multi-process. DO NOT MERGE THIS until after https://github.com/servo/servo/pull/10824 has landed, since it fixes https://github.com/servo/servo/issues/10853.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/341)


aneeshusa:
servo/servo#10824 has landed, so this should be good to go.

@bors-servo r+

Issue #338, opened 2016-04-25
Refactor python function to create commandline arguments


larsbergstrom:
In https://github.com/servo/saltfs/pull/331, we added a function to create the path string, `make_cd_string`:
https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/factories.py#L118-L119

However, this could be cleaner! As seen in https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/factories.py#L126-L127, there are more arguments in the array that are static.

In this fix, please rename the function and make it return an array with *all* of the arguments required, while still taking the same single argument (the command to execute).

ashrko619:
Are you suggesting something like this ?
```python
def make_cd_command(command):
    return ["bash", "-l", "-c", "cd /c/buildbot/slave/windows/build ; " + command ]
	
windows = ServoFactory([
    # TODO: convert this to use DynamicServoFactory
    # We need to run each command in a bash login shell, which breaks the
    # heuristics used by DynamicServoFactory.make_step
    steps.Compile(command= make_cd_command("./mach build -d -v"),
                  env=envs.build_windows),
    steps.Compile(command= make_cd_command("./mach test-unit"),
                  env=envs.build_windows),

    # TODO: run lockfile_changed.sh and manifest_changed.sh scripts
])
```

aneeshusa:
@ashrko619 Yes, that's the right direction. If you open a PR I have a few style nits, but it looks good otherwise.

ashrko619:
@aneeshusa Thanks, I will open a PR, btw what are the style nits here ?

aneeshusa:
Here, we need to rename the `make_cd_command` function to reflect its updated scope, and there are some extra spaces.

Generally, I prefer to do reviewing on PRs because inline comments are clearer, so go ahead and open a PR and I'll make any future comments there.

aneeshusa:
Fixed by #342.

PR #342, opened 2016-04-27
Refactored python function to create command line arguments

ashrko619:
This commit is for issue #338

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/342)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -115,19 +115,18 @@ doc = ServoFactory([
 ])
 
 
-def make_cd_string(command):
-    return "cd /c/buildbot/slave/windows/build ; " + command
+def make_win_command(command):
+    cd_command = "cd /c/buildbot/slave/windows/build; " + command
+    return ["bash", "-l", "-c", cd_command]
 
 
 windows = ServoFactory([
     # TODO: convert this to use DynamicServoFactory
     # We need to run each command in a bash login shell, which breaks the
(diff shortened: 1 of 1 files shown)


buildbot/master/files/config/factories.py
```diff
@@ -115,19 +115,17 @@ def make_step(self, command):
 ])
 
 
-def make_cd_string(command):
-    return "cd /c/buildbot/slave/windows/build ; " + command
+def make_cd_command(command):
```
aneeshusa: This should be renamed to express the intent of the function, instead of what it is doing. Let's use `make_win_command`, since this function takes a regular command and makes it palatable to Windows.

buildbot/master/files/config/factories.py
```diff
@@ -115,19 +115,17 @@ def make_step(self, command):
...
+    return ["bash", "-l", "-c", "cd /c/buildbot/slave/windows/build ; " + command]
```
aneeshusa: You'll need to split this onto two lines to fix the lint error (see the Travis logs for details). Also, the space between `build` and the semicolon can be removed.

I'd suggest using one line to construct the cd-prefixed string, and the second line to return the whole array.

aneeshusa:
This LGTM; please squash these commits together (you can use the commit message from the first commit).

aneeshusa:
Thank you for the PR!
@bors-servo r+

PR #347, opened 2016-04-29
Update arm libs

dati91:
This should fix the https://github.com/servo/servo/pull/10632 arm build errors.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/347)

### DIFF
--- a/servo-build-dependencies/map.jinja
+++ b/servo-build-dependencies/map.jinja
@@ -18,15 +18,15 @@
         'name': 'arm-linux-gnueabihf',
         'download_name': 'armhf-trusty-libs',
         'symlink_name': 'arm-unknown-linux-gnueabihf',
-        'version': 'v1',
-        'sha512': 'd9a31ed488e4f848efcd07f71aa167fc73252da2a2c3b53ba8216100e2b4302b5ccd273b27c434ad189650652a1877607d328ff6b8e1edb5aa68a8927c617b49'
+        'version': 'v2',
+        'sha512': '9273c980b45f2b8602afb98f904582b7de749cab5d388d834c085c95387f3423f330f30d85ce6e5ad55ebdd68fdd03602d3b6db23da07fb2f6dcafb429e79ac7'
       },
       {
         'name': 'aarch64-linux-gnu',
         'download_name': 'arm64-trusty-libs',
         'symlink_name': 'aarch64-unknown-linux-gnu',
-        'version': 'v1',
-        'sha512': '6c86097188b70940835b2fc936fe70f01890fae45ba4ef79dcccc6a552ad319dcba23e21d6c849fd9d396e0c2f4476a21c93f9b3d4abb4f34d69f22d18017b1b'
+        'version': 'v2',
+        'sha512': 'a08a90f8a91d644c9e8d7af7fe50fca8a92cdbe5dcbe5faebc16333c9429acd1f487add66a847b64f83ed87ee8e14d5b71efc0a3dba35657f527e649e6e3e35c'
       }
     ]
   }


dati91:
CC @larsbergstrom

aneeshusa:
This change wouldn't do anything; the list of packages you added `libdbus-glib-1-dev` to in the previous PR is installed on all of our builders, and this list is for additional packages that we need on the cross builders. This will require a little more investigation...

aneeshusa:
I think trying to install the armhf and arm64 versions in this file should work, so that they get picked up by pkg-config.

dati91:
@aneeshusa  You are right, my current approach won't solve the problem. :(

It seems to me that [this](https://github.com/servo/saltfs/blob/master/servo-build-dependencies/map.jinja#L19) sysroot needs to be updated with the `libdbus-1-dev` library (not necessary with `libdbus-glib-1-dev` because we only use this because the other one isn't whitelisted on travis).

And we should do the same for the aarch64 rootfs as well.

larsbergstrom:
Yup, that sounds right! If you could please regenerate those two tarballs, I can work with @aneeshusa to upload them to s3 and use the versioned directory structures to roll it out and make them available.

aneeshusa:
@larsbergstrom Could you point to some instructions on how those tarballs are generated, or better yet a repo with the scripts? I think it would be valuable to link to that information from this repo for reproducibility (and I'm curious about how it's done!).

mmatyas:
Hi, I've regenerated the lib package, you can find them [here](https://github.com/mmatyas/mmatyas.github.io/releases/tag/arm-libs), with the `_dbus` suffix. It's based on [this guide](https://mmatyas.github.io/blog/servo-short-cross-compilation-guide).

aneeshusa:
Thanks @mmatyas! I filed #352 about scripting that process.

@dati91, we'll need to wait for @larsbergstrom to upload the new tarballs to S3, but in the meantime can you update this PR to point our ARM builders to the new tarballs? Should be as easy as updating the version and sha512 fields in [this file](https://github.com/servo/saltfs/blob/master/servo-build-dependencies/map.jinja).

larsbergstrom:
I've made them available at the `v2` directory, so updating the version number when updating the sha512 should work almost immediately. I kept the existing names for the files (omitting the `_dbus` suffix) when uploading them.

aneeshusa:
@larsbergstrom I checked S3 and you put the tarballs in the wrong place - can you please upload the new tarballs to `https://servo-rust.s3.amazonaws.com/ARM/armhf-trusty-libs/v2/armhf-trusty-libs-v2.tgz` and `https://servo-rust.s3.amazonaws.com/ARM/arm64-trusty-libs/v2/arm64-trusty-libs-v2.tgz` instead? This is where the state files are expecting to find them.

servo-build-dependencies/arm.sls
```diff
@@ -36,7 +36,7 @@ arm-dependencies:
 libs-{{ target.name }}:
   archive.extracted:
     - name: {{ common.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }}
-    - source: https://servo-rust.s3.amazonaws.com/ARM/{{ target.download_name }}/{{ target.version }}/{{ target.download_name }}-{{ target.version }}.tgz
+    - source: https://servo-rust.s3.amazonaws.com/ARM/{{ target.version }}/{{ target.download_name }}.tgz
```
aneeshusa: Please keep this URL the way it is - it's more verbose but it's more robust:

- Can update arm32 and arm64 independently
- If you download a bunch of tarballs from S3 , they'll have different names and will happily coexist on your filesystem

dati91: But currently the `v2` lib can only be found on that specific url (and the others as well).

larsbergstrom: I think I put them there now:
```
[larsberg@lbergstrom ~]$ s3cmd cp s3://servo-rust/ARM/v2/armhf* s3://servo-rust/ARM/armhf-trusty-libs/v2/
remote copy: 's3://servo-rust/ARM/v2/armhf-trusty-libs.tgz' -> 's3://servo-rust/ARM/armhf-trusty-libs/v2/armhf-trusty-libs.tgz'
[larsberg@lbergstrom ~]$ s3cmd cp s3://servo-rust/ARM/v2/arm64* s3://servo-rust/ARM/arm64-trusty-libs/v2/
remote copy: 's3://servo-rust/ARM/v2/arm64-trusty-libs.tgz' -> 's3://servo-rust/ARM/arm64-trusty-libs/v2/arm64-trusty-libs.tgz'
[larsberg@lbergstrom ~]$ 
```

dati91: if you want the original link, you still missing the `*-{{target.version}}.tgz` from the end

larsbergstrom: Ugh, I'll try a third time later tonight.

On May 3, 2016, at 4:58 PM, Attila Dusnoki <notifications@github.com> wrote:

In servo-build-dependencies/arm.sls
<https://github.com/servo/saltfs/pull/347#discussion_r61963556>:

> @@ -36,7 +36,7 @@ arm-dependencies:
>  libs-{{ target.name }}:
>    archive.extracted:
>      - name: {{ common.servo_home }}/rootfs-trusty-{{ target.name }}/{{ target.version }}
> -    - source: https://servo-rust.s3.amazonaws.com/ARM/{{ target.download_name }}/{{ target.version }}/{{ target.download_name }}-{{ target.version }}.tgz
> +    - source: https://servo-rust.s3.amazonaws.com/ARM/{{ target.version }}/{{ target.download_name }}.tgz

if you want the original link, you still missing the
*-{{target.version}}.tgz from the end

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/servo/saltfs/pull/347/files/5303ab87b1e2bcb61ae42f3e27329ee677dd39d3#r61963556>

aneeshusa:
The tarballs should be uploaded to the right place now, but I'd like to check this out in Vagrant before approving to make sure the upgrade will work smoothly.

aneeshusa:
Everything looked good in Vagrant, thanks for the PR.

@bors-servo r+

PR #360, opened 2016-05-05
Run etc/ci/chaos_monkey_test.py, non-gating

edunham:
As requested in https://github.com/servo/servo/issues/10927

r? @larsbergstrom

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/360)


aneeshusa:
cc @asajeffrey 
I think we wanted to try running these on a nightly builder first before gating, since this test has the potential to be very intermittent.

asajeffrey:
Indeed, it would be nice to run this without gating, just so we can see if we're getting new intermittents. If that's not doable, we could gate as long as we can ungate in a hurry if necessary! This may make Monday a better day to start gating than a Friday when I'm travelling :)

edunham:
@aneeshusa Current changes create a new builder, linux-nogate, which will run chaos_monkey_test whenever the other Linux builders are triggered. Build results will not be considered by Homu, because we aren't telling Homu that the builder exists.

buildbot/master/files/config/steps.yml
```diff
@@ -34,6 +34,9 @@ linux-dev:
   - bash ./etc/ci/manifest_changed.sh
   - bash ./etc/ci/check_no_unwrap.sh
 
+linux-nogate:
+  - bash ./etc/ci/chaos_monkey_test.sh
```
aneeshusa: This is actually a python script.

aneeshusa: Also, I think we need to `./mach build --release` beforehand.

edunham:
failure on master is because 
```
    DynamicServoBuilder("linux-nogate", LINUX_SLAVES, envs.build_linux_headless),
``` 

is 81 characters. 

@aneeshusa which workaround for this lint failure would you consider least objectionable -- renaming `linux-nogate` to something shorter, renaming `envs.build_linux_headless` to something shorter, breaking the line, changing the lint...

aneeshusa:
Get rid of `envs.build_linux_headless`. https://github.com/servo/servo/pull/8641 got rid of the SERVO_HEADLESS environment variable (headless is a runtime option instead of a compile time option), so we can just use `envs.build_linux` instead.

aneeshusa:
r=me after switching the order of the commits (I think separate commits is better than squashing here).

aneeshusa:
Also, if you have enough space on the line, it would be nice to rename `linux-nogate` to `linux-rel-nogate`.

aneeshusa:
Ping on finishing this - could prevent issues like servo/servo#11926.

buildbot/master/files/config/master.cfg
```diff
@@ -160,6 +162,9 @@ class DynamicServoYAMLBuilder(util.BuilderConfig):
         )
 
 c['builders'] = [
+    DynamicServoBuilder("linux-dev", LINUX_SLAVES, envs.build_linux),
```
aneeshusa: Don't duplicate `linux-dev` and `linux-rel`.

buildbot/master/files/config/master.cfg
```diff
@@ -160,6 +162,9 @@ class DynamicServoYAMLBuilder(util.BuilderConfig):
...
+    DynamicServoBuilder("linux-nogate", LINUX_SLAVES, envs.build_linux),
```
aneeshusa: Ideally, this is `linux-rel-nogate` if it fits on one line.

aneeshusa:
LGTM aside from the nits, please rebase + squash.

edunham:
@aneeshusa Updated this one to account for the refactors since I accidentally forgot it existed.

aneeshusa:
Thanks @edunham for getting back to this!
@bors-servo r+

PR #361, opened 2016-05-05
Disable Homebrew Analytics

### LINKED ISSUE #340: Disable Homebrew analytics
Opening this as a reminder for later:

- [ ] Run command from https://github.com/Homebrew/brew/pull/146 on all OS X machines to disable analytics
- [ ] Update installation instructions for OS X machines on the wiki

aneeshusa:
This is not the most robust of solutions, but will do the job.
A better solution would be a custom Salt state, backed by a custom
Salt module, that uses the `brew analytics` command to stay decoupled
from implementation details.

Instead, directly change the git config settings of the local
Homebrew repo to disable analytics.

Fixes #340.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/361)

### DIFF
--- a/osx/init.sls
+++ b/osx/init.sls
@@ -11,3 +11,15 @@
     - group: wheel
     - mode: 644
     - source: salt://{{ tpldir }}/files/profile
+
+# Disable Homebrew Analytics
+# TODO: wrap this up into a proper state that uses the `brew analytics` command
+# instead of directly changing the git configuration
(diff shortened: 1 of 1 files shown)


aneeshusa:
In terms of testing, we'd need to `brew install python3` on Travis, and it seems this requiring compiling from source, which could take a long time, so I haven't added a test for this.

PR #362, opened 2016-05-06
Add caching of Cargo and Rust bits to Windows

### LINKED ISSUE #358: Windows builder is not caching cargo and rust deps
Need to set the `CARGO_HOME` and `SERVO_CACHE_DIR` environment variables to something sensible to avoid wasting a *ton* of time on our Windows builder.

Changes should be made to this file:
https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/environments.py

larsbergstrom:
r? @edunham @aneeshusa 

Fixes #358

I decided to use the msys home dir instead of `%APPDATA%` for consistency with the other unix builds, but could be convinced otherwise if we get to a point where Servo is not building inside of the msys/mingw environment.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/362)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -29,6 +29,7 @@ build_common = Environment({
 })
 
 build_windows = build_common + Environment({
+    'CARGO_HOME': r'C:\msys64\home\Administrator\.cargo',
     'MSYS': 'winsymlinks=lnk',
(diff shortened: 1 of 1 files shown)


aneeshusa:
I put all of the environment variables in alphabetical order, so please move `SERVO_CACHE_DIR` to be after `PATH` and then r=me.

PR #365, opened 2016-05-06
Disable caching

larsbergstrom:
r? @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/365)


aneeshusa:
Can you update the commit message to mention that it's reverting the caching, and that the reason is the paths aren't interpreted correctly, just for future reference since we clean out the Buildbot logs? (Maybe even paste in the error message if it isn't too long.) r=me after that.

IRC convo:  http://logs.glob.uno/?c=mozilla%23servo#c424015

PR #366, opened 2016-05-06
This increases the number of processes on the Mac builders.

larsbergstrom:
Based on testing on the minis, it takes test-wpt from 30min down to 15min3s.
It takes test-css from 22min down to 18min30s. test-css is now gated on the
`fontd` process; further increase in processes will not help.

r? @aneeshusa @edunham

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/366)


aneeshusa:
For reference, this was made possible by https://github.com/servo/servo/pull/10935; we were gated by thrashing the Dock before.

PR #368, opened 2016-05-10
Bug 316 - Load steps YAML from Servo repo

### LINKED ISSUE #316: Move buildbot logic to scripts in `/etc/ci`

shinglyu:
This should fix #316 

I only tested it with `buildbot checkconfig master` (in the `servo-master1` VM). I did try to execute the step in the python interactive console, but some part requires extensive mocking, so I gave up and submit a PR first.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/368)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -4,6 +4,7 @@ import re
 from buildbot.plugins import steps, util
 from buildbot.process import buildstep
 from buildbot.status.results import SUCCESS
+from twisted.internet import defer
 import yaml
 
 import environments as envs
@@ -102,6 +103,104 @@ class DynamicServoFactory(ServoFactory):
                 step_kwargs['logfiles'][logfile] = logfile
 
         return step_class(**step_kwargs)
+
+
+class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
+    """
+    Step which resolves the in-tree YAML and dynamically adds test steps.
+    """
+
+    haltOnFailure = True
+    flunkOnFailure = True
+
+    def __init__(self, builder_name, environment, yaml_path, **kwargs):
+        kwargs = self.setupShellMixin(kwargs)
+        buildstep.BuildStep.__init__(self, **kwargs)
+        self.builder_name = builder_name
+        self.environment = environment
+        self.yaml_path = yaml_path
+
+    @defer.inlineCallbacks
+    def run(self):
+        try:
+            print_yaml_cmd = "cat {}".format(self.yaml_path)
+            cmd = yield self.makeRemoteShellCommand(
+                command=[print_yaml_cmd],
+                collectStdout=True
+            )
+            yield self.runCommand(cmd)
+
+            result = cmd.results()
+            if result != util.SUCCESS:
+                raise Exception("Command failed with return code: {}" .format(
+                    str(cmd.rc)
+                ))
+            else:
+                builder_steps = yaml.safe_load(cmd.stdout)
+                commands = builder_steps[self.builder_name]
+                dynamic_steps = [self.make_step(command)
+                                 for command in commands]
+        except Exception as e:  # Bad step configuration, fail build
+            # Capture the exception and re-raise with a friendly message
+            raise Exception("Bad step configuration for {}: {}".format(
+                self.builder_name,
+                str(e)
+            ))
+
+        # TODO: windows compatibility (use a custom script for this?)
+        pkill_step = steps.ShellCommand(command=["pkill", "-x", "servo"],
+                                        decodeRC={0: SUCCESS, 1: SUCCESS})
+        static_steps = [pkill_step]
+
+        self.build.steps += static_steps + dynamic_steps
+
+        defer.returnValue(result)
+
+    def make_step(self, command):
+        step_kwargs = {}
+        step_kwargs['env'] = self.environment
+
+        command = command.split(' ')
+        step_kwargs['command'] = command
+
+        step_class = steps.ShellCommand
+        args = iter(command)
+        for arg in args:
+            # Change Step class to capture warnings as needed
+            # (steps.Compile and steps.Test catch warnings)
+            if arg == './mach':
+                mach_arg = next(args)
+                if re.match('build(-.*)?', mach_arg):
+                    step_class = steps.Compile
+                elif re.match('test-.*', mach_arg):
+                    step_class = steps.Test
+
+            # Capture any logfiles
+            elif re.match('--log-.*', arg):
+                logfile = next(args)
+                if 'logfiles' not in step_kwargs:
+                    step_kwargs['logfiles'] = {}
+                step_kwargs['logfiles'][logfile] = logfile
+
+        return step_class(**step_kwargs)
+
+
+class DynamicServoYAMLFactory(ServoFactory):
+    """
+    Smart factory which takes a list of shell commands from a YAML file
+    located in the main servo/servo repository and creates the appropriate
+    Buildbot Steps.
+    Uses heuristics to infer Step type, if there are any logfiles, etc.
+    """
+
+    def __init__(self, builder_name, environment):
+
+        # util.BuildFactory is an old-style class so we cannot use super()
+        # but must hardcode the superclass here
+        ServoFactory.__init__(self, [
+            StepsYAMLParsingStep(builder_name, environment,
+                                 "etc/ci/buildbot_steps.yml")
+        ])
 
 
 doc = ServoFactory([
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -65,6 +65,7 @@ c['schedulers'].append(schedulers.AnyBranchScheduler(
     builderNames=[
         "linux-dev",
         "linux-rel",
+        "linux-dev-yaml",
         "mac-rel-wpt",
         "mac-dev-unit",
         "mac-rel-css",
@@ -86,6 +87,7 @@ c['schedulers'].append(schedulers.ForceScheduler(
     builderNames=[
         "linux-dev",
         "linux-rel",
+        "linux-dev-yaml",
         "mac-rel-wpt",
         "mac-dev-unit",
         "mac-rel-css",
@@ -134,6 +136,23 @@ class DynamicServoBuilder(util.BuilderConfig):
         )
 
 
+class DynamicServoYAMLBuilder(util.BuilderConfig):
+    """
+    Builder which uses DynamicServoYAMLFactory to run steps from a
+    yaml file in the Servo source tree.
+    """
+    def __init__(self, name, slavenames, environment):
+        # util.BuilderConfig is an old-style class so we cannot use super()
+        # but must hardcode the superclass here
+        util.BuilderConfig.__init__(
+            self,
+            name=name,
+            slavenames=slavenames,
+            factory=factories.DynamicServoYAMLFactory(name, environment),
+            nextBuild=branch_priority,
+            category="auto",
+        )
+
 c['builders'] = [
     DynamicServoBuilder("linux-dev", LINUX_SLAVES, envs.build_linux),
     DynamicServoBuilder("linux-rel", LINUX_SLAVES, envs.build_linux),
@@ -158,6 +177,9 @@ c['builders'] = [
         factory=factories.doc,
         category="auto",
     ),
+    # Testing the In Tree YAML build
+    DynamicServoYAMLBuilder("linux-dev-yaml", LINUX_SLAVES,
+                            envs.build_linux)
 ]
 
 


edunham:
Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[buildbot/master/files/config/master.cfg, line 141 \[r1\]](https://reviewable.io:443/reviews/servo/saltfs/368#-KHWWjW1S7dpcS4KYH-4:-KHWWjW1S7dpcS4KYH-5:-1575511023) ([raw file](https://github.com/servo/saltfs/blob/368ffc9e51659293f647efe26baaa34e97f1a0e6/buildbot/master/files/config/master.cfg#L141)):*
> ```INI
>     """
>     Builder which uses DynamicServoIntreeYAMLFactory to run steps from a 
>     yaml file in the Servo source tree.
> ```

This comment is unclear to me -- I expect that the yaml builder would only parse the YAML file from the Servo repo then hand that config back to the master, yet the comment makes it sound like the yaml builder will be what then runs the loaded steps.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/368)*

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
 
 
+class StepsYAMLParsingStep(buildstep.BuildStep):
+    """
+    Step which resolves the in-tree YAML and dynamically add test steps
+    """
+    yaml_path = ""
+
+    def __init__(self, builder_name, environment, yaml_path):
+        print(yaml_path)
```
aneeshusa: No need to print this.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+        self.builder_name = builder_name
+        self.environment = environment
+        self.yaml_path = yaml_path
+
+    def run(self):
+        print(self.yaml_path)
```
aneeshusa: No need to print this.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+        except Exception as e:  # Bad step configuration, fail build
+            print(str(e))
+            dynamic_steps = [BadConfigurationStep(e)]
+
+        # TODO: windows compatibility (use a custom script for this?)
+        pkill_step = [steps.ShellCommand(command=["pkill", "-x", "servo"],
+                                         decodeRC={0: SUCCESS, 1: SUCCESS})]
+
+        # util.BuildFactory is an old-style class so we cannot use super()
+        # but must hardcode the superclass here
```
aneeshusa: Comment doesn't apply here.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+        self.build.steps = pkill_step + dynamic_steps
```
aneeshusa: Let's use `+=` instead of `=` for now, so that we have a record of the StepsYAMLParsingStep in the buildbot logs.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+        print(self.yaml_path)
+        try:
+            full_yaml_path = os.path.join(self.build.workdir, self.yaml_path)
+            with open(full_yaml_path) as steps_file:
+                builder_steps = yaml.safe_load(steps_file)
+            commands = builder_steps[self.builder_name]
+            dynamic_steps = [self.make_step(command) for command in commands]
+        except Exception as e:  # Bad step configuration, fail build
+            print(str(e))
+            dynamic_steps = [BadConfigurationStep(e)]
```
aneeshusa: Instead of creating a `BadConfigurationStep` to raise an exception, we should just raise the exception here since we're already in a step. (It was a separate step before because the parsing was done at Buildbot start-up time, and raising an exception them would kill the Buildbot master.)

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
         return step_class(**step_kwargs)
 
 
+class StepsYAMLParsingStep(buildstep.BuildStep):
+    """
+    Step which resolves the in-tree YAML and dynamically add test steps
+    """
+    yaml_path = ""
```
aneeshusa: A default of `""` seems dangerous. Let's just remove this line entirely.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+
```
aneeshusa: We should set `haltOnFailure = True` and `flunkOnFailure = True` here, just as we do for `BadConfigurationStep`.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+class DynamicServoIntreeYAMLFactory(ServoFactory):
+    """
+    Smart factory which takes a list of shell commands from a yaml file
+    and creates the appropriate Buildbot Steps. Uses heuristics to infer
+    Step type, if there are any logfiles, etc.
+    """
+
+    def __init__(self, builder_name, environment):
+        self.environment = environment
```
aneeshusa: No need to stash a copy of this locally, we can remove this line.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+
+        # util.BuildFactory is an old-style class so we cannot use super()
+        # but must hardcode the superclass here
+        # TODO: pass environments
```
aneeshusa: Looks like you took care of this TODO already.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+                step_kwargs['logfiles'][logfile] = logfile
+
+        return step_class(**step_kwargs)
+
+
+class DynamicServoIntreeYAMLFactory(ServoFactory):
+    """
+    Smart factory which takes a list of shell commands from a yaml file
+    and creates the appropriate Buildbot Steps. Uses heuristics to infer
+    Step type, if there are any logfiles, etc.
```
aneeshusa: Please update this comment to make it clear that the YAML file is coming from the main servo/servo repo (and capitalize YAML also).

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
         return step_class(**step_kwargs)
 
 
+class StepsYAMLParsingStep(buildstep.BuildStep):
+    """
+    Step which resolves the in-tree YAML and dynamically add test steps
```
aneeshusa: style nit: Period at the end of this line.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +103,86 @@ def make_step(self, command):
...
+            with open(full_yaml_path) as steps_file:
+                builder_steps = yaml.safe_load(steps_file)
+            commands = builder_steps[self.builder_name]
+            dynamic_steps = [self.make_step(command) for command in commands]
+        except Exception as e:  # Bad step configuration, fail build
+            print(str(e))
+            dynamic_steps = [BadConfigurationStep(e)]
+
+        # TODO: windows compatibility (use a custom script for this?)
+        pkill_step = [steps.ShellCommand(command=["pkill", "-x", "servo"],
```
aneeshusa: Let's not put this step in a list here, since `pkill_step` is singular.

Instead, we can wrap it in a list by adding a line here:
```
static_steps = [pkill_step]
```
and later on using `static_steps + dynamic_steps`.

(We could also directly use `[pkill_step] + dynamic_steps`, but I think that's a little less clear.)

aneeshusa:
Since we're just testing the YAML loading, let's get rid of `linux-rel-yaml` and rename `linux-yaml` to `linux-dev-yaml`. Using the debug build will be faster and use less extra resources.

You'll also need to make a PR to servo/servo to add a linux-dev-yaml section in the `etc/ci/buildbot_steps.yml` file, which will need to land before this does.

`DynamicServoIntreeYAMLFactory` is a pretty long name - once we confirm this is working we should rename it to `DynamicServoFactory` to replace the current version in a follow-up, but we may want to shorten it now.

I believe we need to create and invoke a Buildbot `RemoteCommand`(http://docs.buildbot.net/current/developer/cls-remotecommands.html) (or possibly `RemoteShellCommand`?). From my understanding, steps are run on the master, and use `BuildStep.runCommand`(https://docs.buildbot.net/current/developer/cls-buildsteps.html#buildbot.process.buildstep.BuildStep.runCommand) to actually run commands on the worker. (N.B.: `RemoteCommand` is just a Buildbot class representing a command to run, not a Buildbot `Step`. Same for `RemoteShellCommand`. This tripped me up a bunch.) Because the `buildbot_steps.yml` file is on the worker, we need to invoke `runCommand` to extract the contents of the file on the worker and send them back. I haven't looked into this 100%, but https://github.com/isotoma/buildbot_travis/blob/master/buildbot_travis/steps/base.py should be a decent example. I'll try to look into this more to provide more concrete advice.

Please also address the test suite failure from Travis.

shinglyu:
@aneeshusa So in my `StepsYAMLParsingStep::run()`, I should run a `RemoteCommand` to get the content of the YAML file. (Something like `cat /etc/ci/buildbot_steps.yml`), then parse it and generate the dynamic steps?

shinglyu:
@anneeshusa Also I have a question regarding @edunham 's comment. If I change the `StepsYAMLParsingStep::run()` to get the YAML back with `RemoteCommand`, and then I call the `self.build.steps += ...`, does that effectively make the `DynamicServoYAMLBuilder` (the old `DynamicServoIntreeYAMLBuilder`) run the parsed step? Or the `self.build.steps` is run by other Builders? 

Or perhaps I should just return the parsed YAML content back to the master, and let the master start another builder with the parsed steps? But I don't see a clear place in the `master.cfg` to do this?

aneeshusa:
As far as I understand, you're correct that you'll need to `runCommand` a `RemoteCommand` (or maybe `RemoteShellCommand`), then parse the output and generate the dynamic steps. I highly recommend cloning the [buildbot source tree](https://github.com/buildbot/buildbot), checking out the `eight` branch, and reading the Buildbot code to figure out exactly how this works - I'm not 100% sure.

DynamicServoYAMLBuilder should pick up on the new steps and continue to run them. Buildbot 9 (next major version) explicitly adds support for this behavior: [docs here](http://docs.buildbot.net/latest/manual/cfg-buildfactories.html#dynamic-build-factories) and [code here](https://github.com/buildbot/buildbot/blob/cdfb4a2934be456c5ad2820feebc10dd08480677/master/buildbot/process/build.py#L378).

I think the end result should look pretty close to the `GenerateStagesCommand` in the linked docs.

shinglyu:
@aneeshusa Thanks! Let me read that and get back to you.

shinglyu:
@aneeshusa I ran the script throught REPL and didn't see any error, but I might need a full setup to varify it works.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
         return step_class(**step_kwargs)
 
 
+class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
+    """
+    Step which resolves the in-tree YAML and dynamically add test steps.
```
aneeshusa: grammar nit: add -> adds

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+    def run(self):
+        try:
+            full_yaml_path = os.path.join(self.build.workdir, self.yaml_path)
+            print_yaml_cmd = "cat {}".format(full_yaml_path)
+            cmd = yield self.makeRemoteShellCommand(command=[print_yaml_cmd],
+                                                    collectStdout=True)
+            yield self.runCommand(cmd)
+
+            if cmd.didFail():
+                raise Exception(str(cmd.result()))
```
aneeshusa: I don't think there's a `result` method on `RemoteShellCommand`.
It may be better to capture stderr and output that, or maybe return a message with the return code (`cmd.rc`).

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+    @defer.inlineCallbacks
+    def run(self):
+        try:
+            full_yaml_path = os.path.join(self.build.workdir, self.yaml_path)
+            print_yaml_cmd = "cat {}".format(full_yaml_path)
+            cmd = yield self.makeRemoteShellCommand(command=[print_yaml_cmd],
+                                                    collectStdout=True)
+            yield self.runCommand(cmd)
+
+            if cmd.didFail():
```
aneeshusa: We should check that the result is SUCCESS, not just that it didn't fail, because there are other possible results (e.g. WARNING, RETRY, etc.). We can use this snippet from the Buildbot 9 docs:

```
result = cmd.results()
if result != util.SUCCESS:
    # raise here
# continue extracting stdout and building steps, etc. here
```

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+                raise Exception(str(cmd.result()))
+            else:
+                builder_steps = yaml.load(cmd.stdout)
+                commands = builder_steps[self.builder_name]
+                dynamic_steps = [self.make_step(command) for command in commands]
+        except Exception as e:  # Bad step configuration, fail build
+            # Capture the expception and re-raise with a friendly message
```
aneeshusa: spelling: expception -> exception

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+                            str(e))
+
+        # TODO: windows compatibility (use a custom script for this?)
+        pkill_step = steps.ShellCommand(command=["pkill", "-x", "servo"],
+                                        decodeRC={0: SUCCESS, 1: SUCCESS})
+        static_steps = [pkill_step]
+
+        self.build.steps += static_steps + dynamic_steps
+
+        defer.returnValue(cmd.reslts())
```
aneeshusa: We can just re-use the `result` variable I mentioned adding earlier.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+    def __init__(self, builder_name, environment, yaml_path):
+        self.builder_name = builder_name
+        self.environment = environment
+        self.yaml_path = yaml_path
+
+    @defer.inlineCallbacks
+    def run(self):
+        try:
+            full_yaml_path = os.path.join(self.build.workdir, self.yaml_path)
+            print_yaml_cmd = "cat {}".format(full_yaml_path)
```
aneeshusa: I believe the command will be run in the workdir by default, so just passing `self.yaml_path` as a relative path should work. Also, I don't think the current code wouldn't work for Windows, as the `os.path.join` would take place on the master (and use `/`), but the generated command would actually execute on the worker.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+            cmd = yield self.makeRemoteShellCommand(command=[print_yaml_cmd],
+                                                    collectStdout=True)
+            yield self.runCommand(cmd)
+
+            if cmd.didFail():
+                raise Exception(str(cmd.result()))
+            else:
+                builder_steps = yaml.load(cmd.stdout)
```
aneeshusa: We should use `yaml.safe_load`.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+    haltOnFailure = True
+    flunkOnFailure = True
+
+    def __init__(self, builder_name, environment, yaml_path):
+        self.builder_name = builder_name
+        self.environment = environment
+        self.yaml_path = yaml_path
+
+    @defer.inlineCallbacks
+    def run(self):
```
aneeshusa: Just for now, can you also wrap the entire `run` function in a try/except block? I want to make sure that the Buildbot master process doesn't come down even if this code doesn't work properly. Once it's working we can get rid of it.

buildbot/master/files/config/factories.py
```diff
@@ -103,6 +104,95 @@ def make_step(self, command):
...
+class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
+    """
+    Step which resolves the in-tree YAML and dynamically add test steps.
+    """
+
+    haltOnFailure = True
+    flunkOnFailure = True
+
+    def __init__(self, builder_name, environment, yaml_path):
+        self.builder_name = builder_name
```
aneeshusa: I believe you also need to setup the shell mixin and call `BuildStep.__init__` as done in the example in the Buildbot Nine docs.

aneeshusa:
Please also rebase when you address the review comments. Looks pretty close, so I'm hoping this works on the first try!

buildbot/master/files/config/factories.py
shinglyu: That was addressed but GitHub didn't know

buildbot/master/files/config/master.cfg
```diff
@@ -158,6 +177,9 @@ c['builders'] = [
         factory=factories.doc,
         category="auto",
     ),
+    # Testing the In Tree YAML build
+    DynamicServoYAMLBuilder("linux-dev-yaml", LINUX_SLAVES,
+                            envs.build_linux_headless)
```
aneeshusa: `envs.build_linux_headless` has been removed; use `envs.build_linux` instead. (I think it might fit on the previous line, not sure.)

buildbot/master/files/config/factories.py
```diff
@@ -104,6 +105,106 @@ def make_step(self, command):
...
+                cmd = yield self.makeRemoteShellCommand(
+                    command=[print_yaml_cmd],
+                    collectStdout=True
+                )
+                yield self.runCommand(cmd)
+
+                result = cmd.results()
+                if result != util.SUCCESS:
+                    raise Exception(
+                        "Command failed with return code: {}" + str(cmd.rc)
```
aneeshusa: I think you meant to use `.format()` here.

buildbot/master/files/config/factories.py
```diff
@@ -104,6 +105,106 @@ def make_step(self, command):
...
+                else:
+                    builder_steps = yaml.safe_load(cmd.stdout)
+                    commands = builder_steps[self.builder_name]
+                    dynamic_steps = [self.make_step(command)
+                                     for command in commands]
+            except Exception as e:  # Bad step configuration, fail build
+                # Capture the exception and re-raise with a friendly message
+                raise Exception("Bad configuration, " +
+                                "unable to convert to steps" +
+                                str(e))
```
aneeshusa: Let's change this block so that the error message is uninterrupted (e.g. greppable) and a little friendlier:
```
raise Exception("Bad steps configuration for {}: {}".format(
    self.builder_name,
    str(e)
))
```

buildbot/master/files/config/factories.py
```diff
@@ -104,6 +105,106 @@ def make_step(self, command):
...
+            # TODO: windows compatibility (use a custom script for this?)
+            pkill_step = steps.ShellCommand(command=["pkill", "-x", "servo"],
+                                            decodeRC={0: SUCCESS, 1: SUCCESS})
+            static_steps = [pkill_step]
+
+            self.build.steps += static_steps + dynamic_steps
+
+            defer.returnValue(result)
+        except Exception as e:
+            print(str(e))
```
aneeshusa: Please re-raise the exception so that the build fails and we're notified. (Otherwise it may silently succeed if the new steps haven't been added yet!)

aneeshusa: Actually, the entire outermost try/except block isn't helpful and we don't need it, because Buildbot handles exceptions safely (e.g. http://build.servo.org/builders/mac-rel-css/builds/1606). Please remove it, sorry for the go-around.

aneeshusa:
This looks good to me, let's see what Buildbot thinks of it. @bors-servo r+

Issue #288, opened 2016-03-30
Remove hardcoded IPs


edunham:
Follow-up from discussion in https://github.com/servo/saltfs/issues/281 . 

We have a bunch of IPs hardcoded into `common/map.jinja`. This was useful when we had to manage DNS through Mozilla support tickets, but we control everything directly through CloudFlare now. 

Todo: 
- [ ] Remove the IPs from map.jinja
- [ ] Remove the states which add the IPs to the hosts file
- [ ] Test everything in Vagrant
- [ ] Deploy changes to one buildslave, see if it still connects and runs builds successfully

edunham:
Duplicate of https://github.com/servo/saltfs/issues/287

aneeshusa:
This is clearer and has better instructions than the other issue I opened, so closing it in favor of reopening this one.

setupminimal:
Hi! I'm looking to contribute to Servo, and this was mentioned on Servo Starters, so I'm going to try it. Thanks!

aneeshusa:
@setupminimal It's yours, let us know if you have any questions!

tictakk:
Hey I'm a little confused, is this still an issue? Is this still assigned?

aneeshusa:
This actually got taken care of in https://github.com/servo/saltfs/pull/577/commits/95ab185a7be4cc5f245ea2c925ecf7185c0cf84a already, thanks for checking!

PR #369, opened 2016-05-10
Changed IPs to Hostnames where possible, and removed host.present tes…

setupminimal:
…ts from common/init.sls

This theoretically solves Issue #288. It certainly prevents  /etc/hosts from being messed with, and the recommended testing in Vagrant returns all green. I'm still not entirely sure that I didn't miss something, just because I'm unfamiliar with saltfs's code, so I would appreciate some feedback.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/369)

### DIFF
--- a/common/init.sls
+++ b/common/init.sls
@@ -47,14 +47,13 @@ sshkey-{{ ssh_user }}:
     - source: salt://{{ tpldir }}/ssh/{{ ssh_user }}.pub
 {% endfor %}
 
-hosts-file:
+/etc/hosts:
   file.managed:
-    - name: /etc/hosts
     - user: root
     - mode: 644
     {% if grains['os'] == 'MacOS' %}
     - group: wheel
     {% elif grains['os'] == 'Ubuntu' %}
-    - group: root"
+    - group: root
     {% endif %}
-    - location: /common/files/hosts
+    - source: salt://{{ tlpdir }}/files/hosts
--- a/common/init.sls
+++ b/common/init.sls
@@ -40,16 +40,21 @@ servo:
     - shell: /bin/bash
     - home: {{ common.servo_home }}
 
-{% for hostname, ip in common.hosts.items() %}
-host-{{ hostname }}:
-  host.present:
-    - name: {{ hostname }}
-    - ip: {{ ip }}
-{% endfor %}
-
 {% for ssh_user in common.ssh_users %}
 sshkey-{{ ssh_user }}:
   ssh_auth.present:
     - user: root
     - source: salt://{{ tpldir }}/ssh/{{ ssh_user }}.pub
 {% endfor %}
+
+hosts-file:
+  file.managed:
+    - name: /etc/hosts
+    - user: root
+    - mode: 644
+    {% if grains['os'] == 'MacOS' %}
+    - group: wheel
+    {% elif grains['os'] == 'Ubuntu' %}
+    - group: root"
+    {% endif %}
+    - location: /common/files/hosts
--- a/common/files/hosts
+++ b/common/files/hosts
@@ -1,0 +1,3 @@
+127.0.0.1	localhost
+::1		localhost
+255.255.255.255	broadcasthost
--- a/common/map.jinja
+++ b/common/map.jinja
@@ -2,15 +2,6 @@
   set common = salt['grains.filter_by']({
       'defaults': {
         'servo_home': '/home/servo',
-        'hosts': {
-          'servo-master': '96.126.125.232',
-          'servo-linux1': '66.228.48.56',
-          'servo-linux2': '173.255.201.95',
-          'servo-linux3': '45.79.167.177',
-          'servo-linux-android1': '72.14.176.110',
-          'servo-mac1': '208.52.161.130',
-          'servo-mac3': '63.135.170.19',
-        },
         'ssh_users': [
           'ajeffrey',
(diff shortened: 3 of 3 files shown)


aneeshusa:
Thanks for the PR! We actually have moved off Linode, and are using DNS instead of the hosts file for name resolution, so the `hosts` dictionary in the `map.jinja` file should be removed entirely to make sure everything is going through DNS.

setupminimal:
Fixed. Thank you.

aneeshusa:
LGTM, please squash these commits together.

aneeshusa:
Actually, we should use a Salt `file.managed` state to set up the hosts file; this makes sure that the old entries get flushed out.

I checked on the builders and this should be suitable for the contents of that file:
```
127.0.0.1           localhost
::1                 localhost
255.255.255.255     broadcasthost
```
Here are some docs on `file.managed`: https://docs.saltstack.com/en/2015.5/ref/states/all/salt.states.file.html#salt.states.file.managed
We also use it in many other files if you want some examples.

setupminimal:
There. All tested, /etc/hosts is managed, all the extraneous stuff is gone, and it's all squashed into one commit. Is there anything else I should do?

common/init.sls
```diff
@@ -40,16 +40,14 @@ servo:
...
 {% for ssh_user in common.ssh_users %}
 sshkey-{{ ssh_user }}:
   ssh_auth.present:
     - user: root
     - source: salt://{{ tpldir }}/ssh/{{ ssh_user }}.pub
 {% endfor %}
+
+/etc/hosts:
+  file.managed:
+    - contents: "127.0.0.1           localhost\n::1                 localhost\n255.255.255.255     broadcasthost"
```
aneeshusa: We should put these into a file and refer to them using `source` instead of `contents`. Please make a `common/files` directory, and put these into a file called `hosts` in that directory.

setupminimal:
I have to go to bed, and a busy day tomorrow, so I will get back to you Thursday or Friday.

common/init.sls
```diff
@@ -40,16 +40,14 @@ servo:
...
+    -
```
aneeshusa: Please also add `user: root` and `mode: 644`. 

The group is variable and depends on the OS, so we need to use Jinja templating to set it up. It should be `root` on Ubuntu and `wheel` on OS X.

jdm:
@setupminimal Any progress? Any questions?

setupminimal:
I have been banging my head on this for a while, interspersed with studying for finals, and I can't find the problem. I tried to put the /etc/hosts contents in a file, as you said, but then I haven't been able to refer to that file properly. Would you look at what I'm doing and tell me where I went wrong please?

aneeshusa:
@setupminimal 

The way to refer to the file is adding this key/value pair to the `file.managed` state: `- source: salt://{{ tpldir }}/files/hosts`.

For reference, here is the docs link again: https://docs.saltstack.com/en/2015.5/ref/states/all/salt.states.file.html#salt.states.file.managed.

I'm not sure where you got the `location` parameter - that isn't one of the possible arguments to `file.managed`. The right one to use is `source`.

`source` takes a URL value; we want to reference the file from the Salt file tree (instead of a file on the Internet or S3, for example), so we use the `salt://` scheme. The `{{ tpldir }}` part uses Jinja templating to insert the path to the directory the file is in - so, because we're using it from the `common/init.sls` file, it will resolve to just `common`. Then the `/files/hosts` completes the path.

Hopefully that helps; let me know if that's unclear.

common/init.sls
```diff
@@ -40,16 +40,21 @@ servo:
...
-{% endfor %}
-
 {% for ssh_user in common.ssh_users %}
 sshkey-{{ ssh_user }}:
   ssh_auth.present:
     - user: root
     - source: salt://{{ tpldir }}/ssh/{{ ssh_user }}.pub
 {% endfor %}
+
+hosts-file:
```
aneeshusa: This can be `/etc/hosts` instead, and then the `name` parameter is unnecessary down below.

common/init.sls
```diff
@@ -40,16 +40,21 @@ servo:
...
+  file.managed:
+    - name: /etc/hosts
+    - user: root
+    - mode: 644
+    {% if grains['os'] == 'MacOS' %}
+    - group: wheel
+    {% elif grains['os'] == 'Ubuntu' %}
+    - group: root"
```
aneeshusa: There's a trailing `"` here.

aneeshusa:
@setupminimal Sorry to leave this hanging, I didn't see you had updated this! This LGTM and should be good to go after a squash.

common/init.sls
```diff
@@ -40,16 +40,20 @@ servo:
...
+/etc/hosts:
+  file.managed:
+    - user: root
+    - mode: 644
+    {% if grains['os'] == 'MacOS' %}
+    - group: wheel
+    {% elif grains['os'] == 'Ubuntu' %}
+    - group: root
+    {% endif %}
+    - source: salt://{{ tlpdir }}/files/hosts
```
aneeshusa: typo: Actually, this should be `{{ tpldir }}`.

aneeshusa:
Also please rebase against master when you squash.

aneeshusa:
Closing due to lack of response; these commits have been folded into #577.

PR #370, opened 2016-05-11
Enable caching deps on Windows, this time with UNIX paths

larsbergstrom:
r? @aneeshusa @edunham 

*this* time, I tested it locally on the build machine itself inside of a MINGW64 shell that I believe reproduces the exact environment, and it seemed to work.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/370)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -29,6 +29,7 @@ build_common = Environment({
 })
 
 build_windows = build_common + Environment({
+    'CARGO_HOME': '/home/Administrator/.cargo',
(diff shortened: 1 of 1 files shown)


aneeshusa:
We're running builds as Administrator?? Is that equivalent to root? We should be using a separate user for running builds.

larsbergstrom:
I certainly agree! I'd hoped this would be a temporary thing until we could get AppVeyor working, but clearly we should do a bunch more here (salt rules, normal user account - assuming msys/mingw work without admin privs, etc.).

PR #371, opened 2016-05-12
Switch to fresh method for the buildbot Git step

metajack:
This switches from method=clobber to method=fresh, which means instead
of a completely full new clone, we will run git clean -fxxd to reset the
tree and use git fetch to synchronize the repository. This also adds a
single retry on failure for the git fetch step. Failure of the git clean
step will result in a full reclone.

This should alleviate CI failures related to git clone errors, and also
make the step slightly faster and quite a bit lighter on the network load.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/371)


metajack:
@bors-servo r=larsbergstrom

Fixed a too long line.

PR #374, opened 2016-05-15
Also test Salt config changes as an upgrade

aneeshusa:
The current testing always starts from scratch.
However, in practice we do not rebuild all our machines from scratch
on every change, but instead usually incrementally upgrade them.

Run the tests both from scratch, which is used whenever we stand up
a new builder, and as an upgrade from the previous configuration, which
is the more common case.

Additionally, the Salt installation and Salt grains.items and
state.show_highstate commands produce output that is useful for
debugging, but is normally not necessary to view.
Add special markers to the build log that fold these sections on Travis.
This is especially useful with the longer build logs.

Note that this doubles the number of builds run on Travis.

r? @larsbergstrom @edunham

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/374)

### DIFF
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -3,6 +3,34 @@
 set -o errexit
 set -o nounset
 
+travis_fold_start () {
+    printf "travis_fold:start:$1\n"
+    printf "$2\n"
+}
+
+travis_fold_end () {
+    printf "travis_fold:end:$1\n"
+}
+
+run_salt () {
+    travis_fold_start "salt.install.$1" 'Installing and configuring Salt'
+    .travis/install_salt.sh -F -c .travis -- "${TRAVIS_OS_NAME}"
+    .travis/setup_salt_roots.sh
+    travis_fold_end "salt.install.$1"
+
+    travis_fold_start "grains.items.$1" 'Printing Salt grains for debugging'
+    sudo salt-call --id="${SALT_NODE_ID}" grains.items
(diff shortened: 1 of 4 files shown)


edunham:
@bors-servo r+ 

Thank you!

.travis/dispatch.sh
```diff
@@ -3,24 +3,52 @@
...
+    # Using .travis.yml to specify Python 3.5 to be preinstalled, just to check
+    printf "Using $(python3 --version) at $(which python3)\n"
+
+    # Run test suite separately for parallelism
+    ./test.py
+else
+    if [ "${SALT_FROM_SCRATCH}" = "true" ]; then
+        run_salt 'scratch'
+    else
+        git checkout master
```
edunham: Travis doesn't seem to want you to do this; was it working in tests on your branch?

The approved way to use multiple branches seems to just be the matrix (https://docs.travis-ci.com/user/customizing-the-build/#Building-Specific-Branches), though I'm not sure if it'll meet this patch's needs.

aneeshusa: I'm pretty sure this was working on Travis last time I checked.
I think this is caused by Travis doing a shallow clone and running git checkout in different ways on PRs vs branches (auto); I think I can find a workaround.

aneeshusa: I added a git fetch line that should take care of it.

PR #375, opened 2016-05-16
Use Nginx to make Buildbot logs available via CORS

### LINKED ISSUE #354: Make buildbot log files available via CORS
I'm trying to add XHR support to http://hoppipolla.co.uk/410/reftest-analyser-structured.xhtml so we could make investigating reftest failures really easy. Unfortunately attempts to grab URLs like `http://build.servo.org/builders/mac-rel-css/builds/711/steps/test/logs/test-css.log/text` via XHR end up with CORS errors like:
```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at http://build.servo.org/builders/mac-rel-css/builds/711/steps/test/logs/test-css.log/text. (Reason: CORS header 'Access-Control-Allow-Origin' missing).
```

aneeshusa:
Buildbot itself (neither the current version eight nor the upcoming
verison nine) does not seem to have built-in CORS support, so use Nginx
to supply CORS headers while reverse proxying.

This is only needed to download log files, so start with a very
conservative set of CORS headers. In particular:
- Only support basic CORS requests (those that don't require preflight
  requests via OPTIONS). This means supporting only GET requests.
- Do not allow sending cookies or exposing any headers.
  In return, whitelist all domains for CORS, and send the Allow-Origin
  header unconditionally, as the actual value of the Origin header does
  not matter.

Because we are using proxy_pass to reverse proxy Buildbot, we need to
tell Nginx to always add the CORS headers, instead of only adding them
for successful responses. (I.e. add CORS headers even for 404s and other
erroneous status codes). This features is not available in the Nginx
packaged for Ubuntu Trusty, so replace it with a newer Nginx from
upstream Nginx's respositories.

Because Debian/Ubuntu modify the default configuration for Nginx, use
Salt to ensure that the installed nginx.conf properly looks into the
sites-enabled directory (which is a Debian-ism) and that the default
handler in /etc/nginx/conf.d/ is purged.

In order to only support CORS on GET requests, use the 'if' directive.
However, because Nginx's if is evil,
(https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)
add a bunch of tests to ensure that the configuration is working.

Also, update the Nginx states to follow Salt best practices.

Fixes #354. Blocked on #374 because I encountered various failures
while upgrading from the Ubuntu Nginx to upstream Nginx, and I'd
like to make sure I caught them all.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/375)


aneeshusa:
cc @jdm: Is this sufficient to pull the logs you needed or do I need to support OPTIONS/preflight requests too?

jdm:
I don't know of any reason that preflight requests should be required.

aneeshusa:
No longer blocked since #374 merged a while ago. @bors-servo try

PR #381, opened 2016-05-22
replacing "31" with RED constant

d-jh:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/381)

### DIFF
--- a/tests/lint/trailing_whitespace.py
+++ b/tests/lint/trailing_whitespace.py
@@ -1,7 +1,7 @@
 import itertools
 import re
 
-from tests.util import colon, color, display_path, paths, Failure, Success
+from tests.util import *
 
 
 def display_trailing_whitespace(whitespace):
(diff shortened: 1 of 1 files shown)


emilio:
This looks good to me, but it seems that travis complains:

```
./tests/lint/trailing_whitespace.py:11:18: F821 undefined name 'RED'
```

You should import the constant it seems.

d-jh:
How is that?

aneeshusa:
@d-jh Thanks for the PR! @emilio is correct that you need to import the constant. Instead of using `*` to import everything, please add `RED` to the list of imports instead. Using the glob-style import makes it impossible to know what you're importing from where, which causes problems over time as files get bigger and definitions move around.

jdm:
@d-jh Are you planning to finish this up?

KiChjang:
Closing due to lack of reply for 23 days.

PR #384, opened 2016-05-25
Use reduced set of builders for try builds

aneeshusa:
Running all the builds for try builds is unnecessary and uses up extra
resources, slowing down the rate of queue drainage. In particular,
the recently-added linux-rel-gczeal build takes a long time, but is not
yet gated on, and adds no value being run for try builds. Reduce the set
of builders for try builds down to a minimal set that still has full
platform coverage.

cc @metajack @mbrubeck
Conversation on IRC: http://logs.glob.uno/?c=mozilla%23servo&s=25+May+2016&e=25+May+2016#c439048

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/384)

### DIFF
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -46,65 +46,77 @@ c['change_source'] = [changes.PBChangeSource(passwd=CHANGE_PASSWORD)]
 ##################
 
 
-def servo_auto_try_filter(c):
-    return (c.project == 'servo/servo' and
-            c.who.startswith('bors-servo') and
-            c.branch in ["auto", "try"])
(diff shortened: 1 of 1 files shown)


aneeshusa:
FYI, the whitespace tip from https://github.com/blog/967-github-secrets is helpful for reviewing the diff for this on GitHub.

larsbergstrom:
We should just rent more machines from macstadium and Amazon. We have plenty of budget right now  for 2016 :-)

aneeshusa:
The main goal of this was to disable `linux-rel-gczeal` to decrease build times, which was taken care of by #385.

PR #385, opened 2016-05-25
Disable the gczeal builder from auto and force

larsbergstrom:
r? @aneeshusa 

Note that I'm leaving the steps around in the hopes that @nox finds a way to make this useful later :-) We might need to split it out into two builds, though.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/385)

### DIFF
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -65,7 +65,6 @@ c['schedulers'].append(schedulers.AnyBranchScheduler(
     builderNames=[
         "linux-dev",
         "linux-rel",
-        "linux-rel-gczeal",
         "mac-rel-wpt",
         "mac-dev-unit",
         "mac-rel-css",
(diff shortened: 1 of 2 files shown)


aneeshusa:
This also removes the gczeal builder from try - not sure if you wanted that.

If so, I'd rather remove the steps entirely, as it's easy to get them back from git and they're isolated from other things thanks to steps.yml.

aneeshusa:
You'll also need to remove the actual linux-rel-gczeal builder from the list of builders.

aneeshusa:
Another option is just leaving it around for force, so @nox can run builds from the web UI, but it's not run on auto or try. (At least, I think that's what force is.)

larsbergstrom:
OK, I just removed the whole thing and we can ressurect from the git history if @nox decides he's interested in incremental GC in the future :-)

PR #386, opened 2016-05-25
Test packaging on android builder

aneeshusa:
Our android-nightly builder has been busted for a long time, which meant
our packaging code was not being exercised. Also, the nightly builder
only runs once a day instead of on each build, which recently allowed
a regression to slip in due to an uncaught syntax error. To prevent
future regressions, test packaging on the android builder, which runs
on try and auto builds.

See https://github.com/servo/servo/pull/11406 for the bug.
This should not be merged until that fix is merged and android-nightly has
had a chance to run, to ensure that packaging is no longer broken.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/386)

### DIFF
--- a/buildbot/master/files/config/steps.yml
+++ b/buildbot/master/files/config/steps.yml
@@ -47,6 +47,7 @@ linux-rel:
 
 android:
   - ./mach build --android --dev
+  - ./mach package --android --dev
   - bash ./etc/ci/lockfile_changed.sh
   - bash ./etc/ci/manifest_changed.sh
   - python ./etc/ci/check_dynamic_symbols.py


edunham:
Removing blocked-on-external as the Android fixes have landed.

aneeshusa:
Re-adding the label because there are still other issues that need to be resolved. Looking at the android nightly output, the next issue is fixing the build script to invoke `./mach package --android` instead of just `./mach package` now that servo/servo#11210 has landed. Once we get a clean android-nightly run (or at least clean through packaging, the S3 upload can be fixed later), this will get unblocked.

aneeshusa:
Updated this PR to use the new `--android` flag to `./mach package`.

aneeshusa:
Next blocker is #401.

aneeshusa:
It looks like this change was deployed to `servo-master1` without rolling out #388 to the cross builders, causing breakage like http://build.servo.org/builders/android/builds/1582. Please run a full highstate on the builders - I saw that there were some uncommitted changes in `/srv/salt` and didn't want to overwrite anything. This is why I wanted to wait for android-nightly to build cleanly.

PR #387, opened 2016-05-26
Use buildbot GitHub status integration

aneeshusa:
Buildbot 0.8.12 has a GitHub status integration; it would be better to
use that directly instead of making Homu act as a middleman. Note that
this commit does not change the Homu configuration, but only adds a
GitHub status integration for Buildbot.

Note that this requires a new secret value in the Pillar - a token
that Buildbot can use to access the GitHub status API.

See http://docs.buildbot.net/current/manual/cfg-statustargets.html#githubstatus
for more details on the integration.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/387)


edunham:
This looks like it might report more builds than Homu is doing, which would be an improvement. 

@bors-servo r+

Sorry for the delay in review, but this looks good to go after fixing the merge conflict.

aneeshusa:
Rebased. @bors-servo r=edunham

Fun fact: I completely forgot I had already removed Buildbot categories as the first commit in this PR and opened #412 recently as an E-easy, hence only one commit left.

PR #388, opened 2016-05-27
Install Android build tools for cross compilation

aneeshusa:
The Android build tools are required for building an APK.
They are versioned separately from the SDK and do not have a version
requirement - best practice is to use the latest (stable) release.

After servo/servo#11406 landed, I forced an android-nightly build, which now gets further but still fails. The new failure is http://build.servo.org/builders/android-nightly/builds/60/steps/shell_1/logs/stdio, so install the build tools to fix this.

Edit: relevant error for posterity:
```
BUILD FAILED
/home/servo/android/sdk/current/tools/ant/build.xml:483: SDK does not have any Build Tools installed.
```

Another step for servo/servo#10339.

I think this would benefit from #374 - I had a lot of errors trying to get the android tool to install the build tools where blowing away the directory fixed it. I think the --all parameter should take care of it, but I'd like to be more sure.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/388)

### DIFF
--- a/servo-build-dependencies/android.sls
+++ b/servo-build-dependencies/android.sls
@@ -50,7 +50,7 @@ android-sdk:
     - name: |
         expect -c '
         set timeout -1;
-        spawn {{ common.servo_home }}/android/sdk/{{ android.sdk.version }}/android-sdk-linux/tools/android - update sdk --no-ui --filter platform-tool,android-{{ android.platform }};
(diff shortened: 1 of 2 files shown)


aneeshusa:
May also help with servo/servo#8464, since we can depend on a more stable version of the build tools. (We can just require the same build tools version for packagers that we run on the CI, at least for now.)

PR #393, opened 2016-06-08
Install freetype on OSX

talklittle:
To fix this broken Mac build: https://github.com/servo/servo/pull/11649#issuecomment-224275471

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/393)

### DIFF
--- a/servo-build-dependencies/init.sls
+++ b/servo-build-dependencies/init.sls
@@ -10,6 +10,7 @@ servo-dependencies:
       - automake
       - pkg-config
       - openssl
+      - freetype
       {% else %}
       - libglib2.0-dev
       - libgl1-mesa-dri


servo-build-dependencies/init.sls
```diff
@@ -10,6 +10,7 @@ servo-dependencies:
       - automake
       - pkg-config
       - openssl
+      - freetype2
```
aneeshusa: We use Homebrew for installing packages on OS X, which calls this package `freetype`.

talklittle:
@aneeshusa Thanks for the correction, fixed

aneeshusa:
Thanks for the PR! @bors-servo r+

PR #395, opened 2016-06-10
Remove Angle gating on appveyor

emilio:
It seems to depend on the `__cpuid` intrinsic only exposed by MSVC
for windows, so my quick [attempt](https://github.com/servo/angle/pull/11)
to make it build with mingw has failed.

Will probably try again later (so please don't remove the hook), but
removing the gate seems worth it so pending PRs can merge while I
experiment with that :)

r? @larsbergstrom / @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/395)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -160,8 +160,5 @@ master = "servo"
 
 # Additional AppVeyor testing for relevant repos
 
-[repo.angle.status.appveyor]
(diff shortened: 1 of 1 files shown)


aneeshusa:
To be clear, you still want to gate on Travis, yes?

PR #404, opened 2016-06-23
Update SSH key to RSA now that DSA no longer works

metajack:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/404)


aneeshusa:
DSA keys are certainly weak, so I'm glad to see them go. However, while we are in the process of standardizing and securing our OpenSSH configuration, I personally would like to see everyone move to Ed25519 keys instead of RSA keys. Everyone will need to change how they connect anyways for #253, so I think this would be a nice opportunity to also bump our key strength and take advantage of ECC. Thoughts?

(Note that Ed25519 keys won't work just yet, they will need some Salt work.)

PR #409, opened 2016-06-29
Nightlies

edunham:
2 separate config files because the Android upload to the servo-rust bucket in the Rust account, and the everything-else upload to the servo-developer-preview bucket in the Servo account, require different credentials. @aneeshusa, could you remind me of the right way to salt the near-identical files which end up with different pillar data and different names?

Windows nightly requires: 

* Appropriate package command
* Either refactoring the DynamicServoYAMLFactory to handle how all windows commands need to be run in their own bash shells (correct solution) or writing a new custom Windows factory (fast but bad solution)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/409)


aneeshusa:
Just woke up and haven't looked at this yet, but you must be psychic  @edunham - I was about to open an issue/PR for this last night! This should also fix servo/servo#10339.

buildbot/master/files/config/master.cfg
```diff
@@ -85,23 +85,29 @@ c['schedulers'].append(schedulers.SingleBranchScheduler(
...
     ],
 ))
 c['schedulers'].append(schedulers.Nightly(
     name="Nightly",
     branch="master",
-    builderNames=["android-nightly"],
+    builderNames=[
+        "android-nightly",
+        "linux-nightly",
+        "mac-nightly"
```
aneeshusa: nit: leave a comma at the end of this line to reduce future diff noise

aneeshusa:
I would prefer to use environment variables + `logEnviron: false` to pass the AWS IAM Key Id and Secret Access Key. Using a config file implies it's readable by any process running as `servo`, and we don't have any user isolation yet (e.g. IIRC Homu also runs as `servo`). Using environment variables limits the scope to which these credentials are exposed.
In the future I'd even like to see us using https://docs.aws.amazon.com/STS/latest/APIReference/API_GetSessionToken.html to generate temporary session tokens so even if a token gets leaked it's NBD.

This can be done by modifying the `make_step` functions and adding a new heurisitic for `s3cmd`. By the way - any reason we are using `s3cmd` instead of the AWS CLI?

buildbot/master/files/config/steps.yml
```diff
@@ -22,6 +22,12 @@ mac-rel-css:
   - bash ./etc/ci/lockfile_changed.sh
   - bash ./etc/ci/manifest_changed.sh
 
+mac-nightly:
+  - ./mach build --release
+  - ./mach package --release
+  - s3cmd -c ~/.s3cfg-servo put target/release/*.tar.gz s3://servo-developer-preview/nightly/`date +%Y`/
```
aneeshusa: Two things:
- Prefer "$(command)" form to backticks - [style guide](https://github.com/servo/servo/blob/master/docs/STYLE_GUIDE.md).
- I believe that since make_step splits the command up into a list before passing it to Buildbot, Buildbot will exec the command directly, instead of through a shell, and the command substitution won't work. IMO it would be better to make a script for this in the servo repo in the `etc/ci` directory and call it from here.

buildbot/master/files/config/steps.yml
```diff
@@ -22,6 +22,12 @@ mac-rel-css:
...
+  - rm -rf target/
```
aneeshusa: Why did you add the `rm -rf` lines? Buildbot will clean out the working directory at the start of each build IIRC (or I think maybe the `git` step does this), so it should not be necessary to have as a standalone step.

aneeshusa:
It may also be nice to set up a [`MailNotifier`](http://docs.buildbot.net/current/manual/cfg-statustargets.html#mailnotifier) so Buildbot sends emails when a nightly build fails, but that can be saved for a follow-up.

buildbot/master/files/config/steps.yml
ConnorGBrewster: Also the mac package script now generates a `.dmg` instead of a `.tar.gz` and it is put in the `target` directory (not `debug` or `release` directories)

metajack:
Is there a reason to continue to upload android builds to a Rust controlled bucket? Why don't we just dump everything in the servo bucket?

larsbergstrom:
servo-master1 builds get:
```
[ FAIL ] Buildbot master config check failed:

         Configuration Errors:

           Unknown builder 'linux-nightly' in scheduler 'Nightly'

           Unknown builder 'mac-nightly' in scheduler 'Nightly'

           Unknown builder 'linux-nightly' in scheduler 'force'

           Unknown builder 'mac-nightly' in scheduler 'force'
```

larsbergstrom:
I think they need to be added here:
https://github.com/edunham/saltfs/blob/281a9e39d65248a0e681b81086aa45330d1892cc/buildbot/master/files/config/master.cfg#L171

aneeshusa:
I would rather not rush this PR too much, since we have some time before release and we can always make a build manually for release. @edunham if you want I can make a PR to your branch to implement passing credentials as environment variables.

larsbergstrom:
I'm not super excited about manual builds from user machines and would rather hold off on releasing binaries until we do it from machines that were at least originally installed from a clean state with only the dependencies we've documented & listed in salt. Maybe manual builds from those machines would make me feel less stressed, but publishing bits to the world from laptops the team has taken through security checkpoints in all sorts of interesting countries seems dubious :-)

nightly/files/s3cfg-rust
```diff
@@ -0,0 +1,75 @@
...
+get_continue = False
+gpg_command = /usr/bin/gpg
+gpg_decrypt = %(gpg_command)s -d --verbose --no-use-agent --batch --yes
+--passphrase-fd %(passphrase_fd)s -o %(output_file)s %(input_file)s
+gpg_encrypt = %(gpg_command)s -c --verbose --no-use-agent --batch --yes
+--passphrase-fd %(passphrase_fd)s -o %(output_file)s %(input_file)s
+gpg_passphrase =
+guess_mime_type = True
+host_base = s3.amazonaws.com
+host_bucket = %(bucket)s.s3.amazonaws.com
```
larsbergstrom: Is the `s` after `%(bucket)` supposed to be there?

nightly/files/s3cfg-servo
```diff
@@ -0,0 +1,75 @@
...
+get_continue = False
+gpg_command = /usr/bin/gpg
+gpg_decrypt = %(gpg_command)s -d --verbose --no-use-agent --batch --yes
+--passphrase-fd %(passphrase_fd)s -o %(output_file)s %(input_file)s
+gpg_encrypt = %(gpg_command)s -c --verbose --no-use-agent --batch --yes
+--passphrase-fd %(passphrase_fd)s -o %(output_file)s %(input_file)s
+gpg_passphrase =
+guess_mime_type = True
+host_base = s3.amazonaws.com
+host_bucket = %(bucket)s.s3.amazonaws.com
```
larsbergstrom: Same question about the `s` here.

nightly/files/s3cfg-servo
```diff
@@ -0,0 +1,75 @@
...
+skip_existing = False
+socket_timeout = 300
+stats = False
+stop_on_error = False
+storage_class =
+urlencoding_mode = normal
+use_https = True
+use_mime_magic = True
+verbosity = WARNING
+website_endpoint = http://%(bucket)s.s3-website-%(location)s.amazonaws.com/
```
larsbergstrom: And here :-)

nightly/files/s3cfg-rust
```diff
@@ -0,0 +1,75 @@
...
+skip_existing = False
+socket_timeout = 300
+stats = False
+stop_on_error = False
+storage_class =
+urlencoding_mode = normal
+use_https = True
+use_mime_magic = True
+verbosity = WARNING
+website_endpoint = http://%(bucket)s.s3-website-%(location)s.amazonaws.com/
```
larsbergstrom: Same question :-)

aneeshusa:
This will need servo/servo#11943 at the least.

aneeshusa:
Filed #413 and #414 as follow-ups. Closing in favor of #410 (and #411).

PR #415, opened 2016-06-30
Fix json's anger at newlines

edunham:
Error without this fix:

```
credentials = json.loads('{...}')
      File "/usr/lib/python2.7/json/__init__.py", line 338, in loads
        return _default_decoder.decode(s)
      File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
        obj, end = self.raw_decode(s, idx=_w(s, 0).end())
      File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
        obj, end = self.scan_once(s, idx)
    exceptions.ValueError: Invalid control character at: line 1 column 444 (char 443)
```

note: line 1 column 444 of the giant blob of secrets that I elided is the first occurrence of `\n` in the string.

Reason I think this works: Testing it in prod (sorry!) and also
http://stackoverflow.com/questions/9295439/python-json-loads-fails-with-valueerror-invalid-control-character-at-line-1-c

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/415)

### DIFF
--- a/buildbot/master/files/config/passwords.py
+++ b/buildbot/master/files/config/passwords.py
@@ -3,7 +3,8 @@ import json
 # Jinja will replace the inside with double-quote-using JSON,
 # so use single quotes to delimit the string.
 # Use double quotes inside to keep the expression as a single string.
-credentials = json.loads('{{ pillar["buildbot"]["credentials"]|json }}')
+credentials = json.loads('{{ pillar["buildbot"]["credentials"]|json }}',
+                         strict=False)
 # json.loads creates unicode strings but Buildbot requires bytestrings.
 # Python 2's Unicode situation makes me sad.
 credentials = {k: v.encode('utf-8') for k, v in credentials.items()}


metajack:
Can we not fix the JSON?

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/415#-:-KLXyjhRh_S8ekQcp19V:-894015821)*

edunham:
@metajack the newlines come in from pillar data which would get even less legible/editable if we removed them. So we could fix the json, but I'd prefer to keep the pillar pretty and just tell the parser not to break.

metajack:
@bors-servo r+

Ok. Definitely seem the right thing to do then.

aneeshusa:
How urgent is this fix? I think there is a better way to do this (i.e. doesn't insert the newlines) but it will take me a while to investigate.

aneeshusa:
The secrets causing the problem are `minion-public-key` and `minion-private-key`, both of which were removed in #222 and are no longer used. Removing them from the production pillar should fix the problem.

edunham:
@aneeshusa oh awesome! I'll take them out and close this. Thank you!

Issue #407, opened 2016-06-27
Add test that generated `cfg.toml` for Homu is a valid TOML file


aneeshusa:
Buildbot has a `checkconfig` command to check that the master configuration file is valid, which we test in `tests/sls/buildbot/master/config.py`. AFAIK there is no similar command for Homu yet, but it should be possible to check that the `cfg.toml` file parses as TOML at the least using a Python TOML library. This should help prevent issues like the one addressed in #405.

A more comprehensive alternative that requires more work is creating such a `checkconfig` command for Homu, and writing a test that uses that instead.

Files:
- Create the `tests/sls/homu/` folder and make a new Python file inside with the test. Also include an empty `__init__.py` file in that folder.
- Add a pinned version of the `toml` library to the `requirements.txt` file, so it is available on Travis. This is the library used by Homu; ideally, this is pinned to the same version that Homu uses.
- Update the `.travis/dispatch.sh` file to run the new test on the `servo-master1` node.

Crimack:
Hey, I've made an attempt at doing this on a fork by using a Python library. Two problems though:

1. I can't actually get the Vagrant file to run so I can test? Maybe I'm doing something silly. I keep getting `'extract_id': undefined method 'split' for ["SALT_NODE_ID=servo-linux-cross1", "SALT_FROM_SCRATCH=true"]:Array (NoMethodError)` when I try to run `vagrant up`
2. Is the homu/files/cfg.toml in the master branch a valid TOML file? I think I've written the test correctly, but it keeps failing.

This is my first pull request to a public repo, so thank you very much for the detailed description of what you wanted done! Made making a first contribution pretty easy :)

aneeshusa:
Welcome to Servo!

1. The Vagrantfile is actually broken (again!), this is my fault :sweat_smile:. I'm working on a PR to add a test for it so I don't break it again, but in the meantime the changes in [this commit](https://github.com/servo/saltfs/pull/350/commits/99e02dfd6b34449d5a9cf04ca8297d25cfd35a63) should get you up and running. (You don't need the `is_salt_master` function.)
2. The `homu/files/cfg.toml` file in the repo is not valid TOML; we are running it through the [Jinja template engine](http://jinja.pocoo.org/docs/dev/templates/) to reduce some boilerplate. However, the end result which is deployed by Salt needs to be a valid TOML file, so you should check that file in the test.

[...]

PR #420, opened 2016-07-01
Validate Homu TOML file

Crimack:
Draft of resolution to #407. `config_path` in sls/homu/toml.py is definitely wrong and should be corrected

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/420)

### DIFF
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -52,6 +52,6 @@ else
     # Only run tests against the new configuration
     # TODO: don't hard-code this
     if [ "${SALT_NODE_ID}" = "servo-master1" ]; then
-        ./test.py sls.buildbot.master sls.nginx
+        ./test.py sls.buildbot.master sls.homu sls.nginx
     fi
 fi
--- a/homu/init.sls
+++ b/homu/init.sls
@@ -15,7 +15,9 @@ homu:
       - pkg: python3
       - pip: virtualenv
   pip.installed:
-    - name: git+https://github.com/servo/homu@{{ homu.rev }}
+    - pkgs:
+      - git+https://github.com/servo/homu@{{ homu.rev }}
+      - toml == 0.9.1  # Please ensure this is in sync with requirements.txt
     - bin_env: /home/servo/homu/_venv
     - require:
       - virtualenv: homu
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 # Python modules needed for testing
 # (Travis-CI will auto-install them from this file)
 flake8 == 2.5.4
+toml == 0.9.1  # Please ensure this is in sync with homu/init.sls
--- a/tests/sls/homu/toml.py
+++ b/tests/sls/homu/toml.py
@@ -1,0 +1,11 @@
+import toml
+
+from tests.util import Failure, Success
+
+
+def run():
+    try:
+        toml.load('/home/servo/homu/cfg.toml')
+    except Exception as e:
+        return Failure('Homu config file is not valid TOML:', str(e))
+    return Success('Homu config file is valid TOML')


Crimack:
@aneeshusa I think I see what I've done wrong. I should be checking the generated TOML file at `/etc/init/homu.conf` on `servo-master1` rather than the Jinja template, right?

aneeshusa:
Correct!

tests/sls/homu/toml.py
```diff
@@ -0,0 +1,16 @@
...
+import toml
+
+from tests.util import Failure, Success
+
+
+def run():
+    config_path = os.path.join('/home', 'servo', 'homu', 'cfg.toml')
+    with open(config_path) as conf:
+        try:
+            toml.loads(conf.read())
```
aneeshusa: We can use `toml.load` instead which will handle opening and closing the file for us.

tests/sls/homu/toml.py
```diff
@@ -0,0 +1,16 @@
...
+            return Success('Homu config file is valid TOML')
```
aneeshusa: Pull this out of the `try` block and have it as the last statement in `run()`.

tests/sls/homu/toml.py
```diff
@@ -0,0 +1,16 @@
...
+        except Exception as e:
+            return Failure('Homu config file is not valid TOML: ',
+                           '{}'.format(e))
```
aneeshusa: This can be `str(e)`.

tests/sls/homu/toml.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def run():
+    config_path = os.path.join('/home', 'servo', 'homu', 'cfg.toml')
+    with open(config_path) as conf:
+        try:
+            toml.loads(conf.read())
+            return Success('Homu config file is valid TOML')
+        except Exception as e:
+            return Failure('Homu config file is not valid TOML: ',
```
aneeshusa: No need for the space after the colon. (The output will start from the next line).

tests/sls/homu/toml.py
aneeshusa: Also, this should then fit on the previous line.

tests/sls/homu/toml.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+import toml
+
+from tests.util import Failure, Success
+
+
+def run():
+    config_path = os.path.join('/home', 'servo', 'homu', 'cfg.toml')
```
aneeshusa: It looks strange to see `os.path.join` when we have an explicit `/` in the arguments. Because I don't forsee us ever running Homu on a non-Unix platform, and because this is also hardcoded in the `homu/init.sls` file, let's just hardcode the whole path and inline it where it gets used.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Python modules needed for testing
 # (Travis-CI will auto-install them from this file)
 flake8 == 2.5.4
+toml == 0.9.1
```
aneeshusa: Please also pin this in `homu/init.sls`, and put a comment in both places that mentions the need to keep them in sync. See the [style guide](https://github.com/servo/saltfs/blob/master/STYLE_GUIDE.md#use-pkgs-instead-of-name) for tips.

Crimack:
I think that's all your comments cleared up. Thank you for the feedback by the way, it was very easy to follow!

The additions to `homu/init.sls` might be wrong though? I wasn't sure if I should make a new 'block' (not sure what the proper SaltStack term for this is), or add it to one of the existing ones.

homu/init.sls
```diff
@@ -44,3 +44,9 @@ homu:
     - require:
       - pip: homu
       - file: /home/servo/homu/cfg.toml
+
+toml:
+  pip.installed:
+    - pkgs:
+      - toml == 0.9.1  # Ensure this is up to date with requirements.txt
```
aneeshusa: Instead of a separate Salt state (block), add this to the `pip.installed` state above that installed Homu. Also, let's say 'in sync' instead of 'up to date'.

requirements.txt
```diff
@@ -1,3 +1,4 @@
 # Python modules needed for testing
 # (Travis-CI will auto-install them from this file)
 flake8 == 2.5.4
+toml == 0.9.1  # Please ensure this is also in homu/init.sls
```
aneeshusa: 'is also in' -> 'version is in sync with'

homu/init.sls
Crimack: Done :)

.travis/dispatch.sh
```diff
@@ -52,6 +52,6 @@ else
     # Only run tests against the new configuration
     # TODO: don't hard-code this
     if [ "${SALT_NODE_ID}" = "servo-master1" ]; then
-        ./test.py sls.buildbot.master sls.nginx
+        ./test.py sls.buildbot.master sls.nginx sls.homu
```
aneeshusa: Let's keep these in alphabetical order, so insert `sls.homu` in between `sls.buildbot.master` and `sls.nginx`.

aneeshusa:
LGTM aside from that last comment, so go ahead and squash when you address that.

Crimack:
All good now I think!

aneeshusa:
Checked this locally in Vagrant and everything looks good. @bors-servo r+

PR #424, opened 2016-07-03
Pre-split cat step.yml command

aneeshusa:
We are passing the command as an array, so we need to pre-split the
command into separate words.

Naughtily tested in prod.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/424)


edunham:
@bors-servo r+ 

Well, if it didn't break prod ;)

PR #426, opened 2016-07-05
Convert Windows builders to use DynamicServoFactory

UK992:
I'm not sure if this is the right approaches, but i try.

For process kill i used ```powershell kill```, which works from bash and cmd, but needs admin privileges like any other process kill command.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/426)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -31,6 +31,8 @@ build_common = Environment({
 
 build_windows = build_common + Environment({
     'CARGO_HOME': '/home/Administrator/.cargo',
+    # Set home directory, to avoid adding `cd` command on every command
+    'HOME': r'C:\buildbot\slave\windows\build',
     'MSYS': 'winsymlinks=lnk',
     'MSYSTEM': 'MINGW64',
     'PATH': ';'.join([
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -123,5 +123,3 @@ upload_nightly = Environment({
     'AWS_ACCESS_KEY_ID': S3_UPLOAD_ACCESS_KEY_ID,
     'AWS_SECRET_ACCESS_KEY': S3_UPLOAD_SECRET_ACCESS_KEY,
 })
-
-upload_nightly_windows = build_windows + upload_nightly
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -58,6 +58,7 @@ class DynamicServoFactory(ServoFactory):
 
     def __init__(self, builder_name, environment):
         self.environment = environment
+        self.is_windows = re.match('windows.*', builder_name) is not None
         try:
             config_dir = os.path.dirname(os.path.realpath(__file__))
             yaml_path = os.path.join(config_dir, 'steps.yml')
@@ -69,9 +70,7 @@ class DynamicServoFactory(ServoFactory):
             print(str(e))
             dynamic_steps = [BadConfigurationStep(e)]
 
-        # TODO: windows compatibility (use a custom script for this?)
-        pkill_step = [steps.ShellCommand(command=["pkill", "-x", "servo"],
-                                         decodeRC={0: SUCCESS, 1: SUCCESS})]
+        pkill_step = [self.make_pkill_step("servo")]
 
         # util.BuildFactory is an old-style class so we cannot use super()
         # but must hardcode the superclass here
@@ -82,7 +81,10 @@ class DynamicServoFactory(ServoFactory):
         step_env = copy.deepcopy(self.environment)
 
         command = command.split(' ')
-        step_kwargs['command'] = command
+
+        # Add bash -l before every command on Windows builders
+        bash_args = ["bash", "-l"] if self.is_windows else []
+        step_kwargs['command'] = bash_args + command
 
         step_class = steps.ShellCommand
         args = iter(command)
@@ -106,10 +108,21 @@ class DynamicServoFactory(ServoFactory):
             # Provide environment variables for s3cmd
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
-                step_env = copy.deepcopy(envs.upload_nightly)
+                step_env += envs.upload_nightly
 
         step_kwargs['env'] = step_env
         return step_class(**step_kwargs)
+
+    def make_pkill_step(self, target):
+        if self.is_windows:
+            pkill_command = ["powershell", "kill", "-n", target]
+        else:
+            pkill_command = ["pkill", "-x", target]
+
+        return steps.ShellCommand(
+            command=pkill_command,
+            decodeRC={0: SUCCESS, 1: SUCCESS}
+        )
 
 
 class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
@@ -129,6 +142,7 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
 
     @defer.inlineCallbacks
     def run(self):
+        self.is_windows = re.match('windows.*', self.builder_name) is not None
         try:
             print_yaml_cmd = "cat {}".format(self.yaml_path)
             cmd = yield self.makeRemoteShellCommand(
@@ -154,12 +168,9 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
                 str(e)
             ))
 
-        # TODO: windows compatibility (use a custom script for this?)
-        pkill_step = steps.ShellCommand(command=["pkill", "-x", "servo"],
-                                        decodeRC={0: SUCCESS, 1: SUCCESS})
-        static_steps = [pkill_step]
+        pkill_step = [self.make_pkill_step("servo")]
 
-        self.build.steps += static_steps + dynamic_steps
+        self.build.steps += pkill_step + dynamic_steps
 
         defer.returnValue(result)
 
@@ -168,7 +179,10 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
         step_env = copy.deepcopy(self.environment)
 
         command = command.split(' ')
-        step_kwargs['command'] = command
+
+        # Add bash -l before every command on Windows builders
+        bash_command = ["bash", "-l"] if self.is_windows else []
+        step_kwargs['command'] = bash_command + command
 
         step_class = steps.ShellCommand
         args = iter(command)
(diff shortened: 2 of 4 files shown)


aneeshusa:
Thanks for tackling this!

buildbot/master/files/config/environments.py
```diff
@@ -31,6 +31,7 @@ def __add__(self, other):
 
 build_windows = build_common + Environment({
     'CARGO_HOME': '/home/Administrator/.cargo',
+    'HOME': r'C:\buildbot\slave\windows\build',
```
aneeshusa: This seems like an odd choice of `$HOME`. Why do we need to set it, and why this value?

buildbot/master/files/config/steps.yml
```diff
@@ -76,3 +76,12 @@ arm64:
   - ./mach build --rel --target=aarch64-unknown-linux-gnu
   - bash ./etc/ci/lockfile_changed.sh
   - bash ./etc/ci/manifest_changed.sh
+
+windows:
+  - ./mach build -d -v
```
aneeshusa: Let's do `./mach build --dev` for consistency. Also, please rename to `windows-dev` as well.

buildbot/master/files/config/factories.py
```diff
@@ -82,7 +86,10 @@ def make_step(self, command):
         step_env = copy.deepcopy(self.environment)
 
         command = command.split(' ')
-        step_kwargs['command'] = command
+
+        # Add bash -l before every command on Windows builders
+        bash_args = ["bash", "-l"] if is_windows() else []
```
aneeshusa: We had `["bash", "-l", "-c"]` earlier - missed the `"-c"`.

buildbot/master/files/config/factories.py
```diff
@@ -216,6 +235,10 @@ def __init__(self, builder_name, environment):
         ])
 
 
+def is_windows():
+    return platform.system() == "Windows"
```
aneeshusa: This doesn't work because this code runs on the Buildbot master (Linux). We need to somehow interrogate the builder properties and use that throughout the build. I don't know exactly how to do that.

buildbot/master/files/config/factories.py
```diff
@@ -69,8 +70,11 @@ def __init__(self, builder_name, environment):
             print(str(e))
             dynamic_steps = [BadConfigurationStep(e)]
 
-        # TODO: windows compatibility (use a custom script for this?)
-        pkill_step = [steps.ShellCommand(command=["pkill", "-x", "servo"],
+        if is_windows():
+            pkill_command = ["powershell", "kill", "-n", "servo"]
+        else:
+            pkill_command = ["pkill", "-x", "servo"]
```
aneeshusa: Would be nice to factor this out into a little helper function:
```
def make_pkill_command(target):
   ...

make_pkill_command("servo")
```

aneeshusa: Another option is a script in `etc/ci`, which neatly steps around the "how to detect Windows" problem.

buildbot/master/files/config/factories.py
```diff
@@ -192,7 +208,10 @@ def make_step(self, command):
             # Provide environment variables for s3cmd
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
-                step_env = copy.deepcopy(envs.upload_nightly)
+                if is_windows():
+                    step_env = copy.deepcopy(envs.upload_nightly_windows)
+                else:
+                    step_env = copy.deepcopy(envs.upload_nightly)
```
aneeshusa: We should just do `step_env += envs.upload_nightly` and get rid of `envs.upload_nightly_windows` - would have mentioned this on the earlier PR if it wasn't right before release :)

buildbot/master/files/config/factories.py
aneeshusa: FYI, this is the hard part.

buildbot/master/files/config/factories.py
UK992: i leave ```-c```, to avoid quoting every command, i've tested with appveyor and it work that way too.

buildbot/master/files/config/environments.py
UK992: this set home directory before bash login, so it not needed to add ```cd``` command before every command.

buildbot/master/files/config/steps.yml
UK992: done

buildbot/master/files/config/environments.py
aneeshusa: That's a reasonable hack, but please leave a comment with that explanation for future reference.

buildbot/master/files/config/factories.py
```diff
@@ -1,6 +1,7 @@
 import copy
 import os.path
 import re
+import platform
```
aneeshusa: No longer needed.

buildbot/master/files/config/factories.py
UK992: with new approaches not detect windows builder based on builder name

buildbot/master/files/config/factories.py
```diff
@@ -58,31 +59,37 @@ class DynamicServoFactory(ServoFactory):
 
     def __init__(self, builder_name, environment):
         self.environment = environment
+        is_windows = True if re.match('windows(.*)', builder_name) else False
```
aneeshusa: I think this would make sense as a field on `self.`

buildbot/master/files/config/factories.py
```diff
@@ -215,6 +228,12 @@ def __init__(self, builder_name, environment):
                                  "etc/ci/buildbot_steps.yml")
         ])
 
+def make_pkill_command(target, is_windows):
```
aneeshusa: Similarly, I think this makes sense as a class method - please mind the duplication for now :)

aneeshusa:
I will finish reviewing this either later today or tomorrow.

buildbot/master/files/config/factories.py
UK992: i don't quite understand, should i add this function to classes ```DynamicServoYAMLFactory``` and ```DynamicServoFactory```?

buildbot/master/files/config/factories.py
UK992: removed.

buildbot/master/files/config/factories.py
```diff
@@ -58,31 +58,38 @@ class DynamicServoFactory(ServoFactory):
 
     def __init__(self, builder_name, environment):
         self.environment = environment
+        self.is_windows = True if re.match('windows(.*)',
```
aneeshusa: We can do `self.is_windows = re.match(...) is not None` directly instead. Also, no need for the parentheses around the `.*`.

buildbot/master/files/config/factories.py
```diff
@@ -58,31 +58,38 @@ class DynamicServoFactory(ServoFactory):
...
+            command=self.make_pkill_command("servo", self.is_windows),
+            decodeRC={0: SUCCESS, 1: SUCCESS}
+        )]
 
         # util.BuildFactory is an old-style class so we cannot use super()
         # but must hardcode the superclass here
         ServoFactory.__init__(self, pkill_step + dynamic_steps)
 
-    def make_step(self, command):
+    def make_step(self, command, is_windows):
```
aneeshusa: We can just read from `self.is_windows` instead of using an extra argument.

buildbot/master/files/config/factories.py
```diff
@@ -106,11 +113,17 @@ def make_step(self, command):
             # Provide environment variables for s3cmd
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
-                step_env = copy.deepcopy(envs.upload_nightly)
+                step_env += envs.upload_nightly
 
         step_kwargs['env'] = step_env
         return step_class(**step_kwargs)
 
+    def make_pkill_command(self, target, is_windows):
```
aneeshusa: Use `self.is_windows` here too.

buildbot/master/files/config/factories.py
UK992: done

buildbot/master/files/config/factories.py
UK992: done

buildbot/master/files/config/factories.py
UK992: done

buildbot/master/files/config/factories.py
```diff
@@ -58,20 +58,23 @@ class DynamicServoFactory(ServoFactory):
...
         self.environment = environment
+        self.is_windows = re.match('windows.*', builder_name) is not None
         try:
             config_dir = os.path.dirname(os.path.realpath(__file__))
             yaml_path = os.path.join(config_dir, 'steps.yml')
             with open(yaml_path) as steps_file:
                 builder_steps = yaml.safe_load(steps_file)
             commands = builder_steps[builder_name]
-            dynamic_steps = [self.make_step(command) for command in commands]
+            dynamic_steps = [self.make_step(command, self.is_windows)
```
aneeshusa: No need to pass `self.is_windows` to `make_step()`.

buildbot/master/files/config/factories.py
```diff
@@ -58,20 +58,23 @@ class DynamicServoFactory(ServoFactory):
...
         except Exception as e:  # Bad step configuration, fail build
             print(str(e))
             dynamic_steps = [BadConfigurationStep(e)]
 
-        # TODO: windows compatibility (use a custom script for this?)
-        pkill_step = [steps.ShellCommand(command=["pkill", "-x", "servo"],
-                                         decodeRC={0: SUCCESS, 1: SUCCESS})]
+        pkill_step = [steps.ShellCommand(
+            command=self.make_pkill_command("servo"),
+            decodeRC={0: SUCCESS, 1: SUCCESS}
```
aneeshusa: Does this `decodeRC` apply for the powershell command as well? I think it would be better if the `make_pkill_command` function returned a `steps.ShellCommand` directly instead of just the command array.

UK992: yes it does, if powershell kill doesnt find process, exit with code 1

buildbot/master/files/config/factories.py
UK992: done!

buildbot/master/files/config/factories.py
UK992: also done

buildbot/master/files/config/factories.py
```diff
@@ -58,20 +58,20 @@ class DynamicServoFactory(ServoFactory):
...
+        self.is_windows = re.match('windows.*', builder_name) is not None
         try:
             config_dir = os.path.dirname(os.path.realpath(__file__))
             yaml_path = os.path.join(config_dir, 'steps.yml')
             with open(yaml_path) as steps_file:
                 builder_steps = yaml.safe_load(steps_file)
             commands = builder_steps[builder_name]
-            dynamic_steps = [self.make_step(command) for command in commands]
+            dynamic_steps = [self.make_step(command)
+                             for command in commands]
```
aneeshusa: Put this back on one line as it was before.

buildbot/master/files/config/factories.py
```diff
@@ -106,11 +109,22 @@ def make_step(self, command):
             # Provide environment variables for s3cmd
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
-                step_env = copy.deepcopy(envs.upload_nightly)
+                step_env += envs.upload_nightly
 
         step_kwargs['env'] = step_env
         return step_class(**step_kwargs)
 
+    def make_pkill_command(self, target):
```
aneeshusa: Let's rename to `make_pkill_step` since it returns a `Step`.

buildbot/master/files/config/factories.py
UK992: done

buildbot/master/files/config/factories.py
UK992: done

aneeshusa:
LGTM to me after squash. Nice thinking with using a regex on the builder name to determine `is_windows` - I thought we were going to have to use complicated Buildbot properties or something.

UK992:
Squashed!

aneeshusa:
@bors-servo r+
Thanks for handling not one but two TODOs!

PR #427, opened 2016-07-05
Add ./mach build-stable to linux-dev builder.

gpoesia:
This is the second and last PR related to this issue in servo: https://github.com/servo/servo/issues/11806

This PR added the ./mach build-stable sub-command: https://github.com/servo/servo/pull/11945

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/427)

### DIFF
--- a/buildbot/master/files/config/steps.yml
+++ b/buildbot/master/files/config/steps.yml
@@ -35,6 +35,7 @@ linux-dev:
   - ./mach test-unit
   - ./mach build-cef
   - ./mach build-geckolib
+  - ./mach build-stable --dev
   - bash ./etc/ci/lockfile_changed.sh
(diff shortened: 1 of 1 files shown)


SimonSapin:
I discussed this with Jack and we concluded that the servo/servo repository is unlikely to get code that we want to build on stable Rust other than `ports/geckolib` and its dependencies. And with https://github.com/servo/servo/pull/12263 landed, we can do so now.

So I’m going to remove the `./mach build-stable` sub-command (making this PR unnecessary) and instead make `./mach build-geckolib` use the stable compiler using the bootstrapping and setup that you introduced in https://github.com/servo/servo/pull/11945. Thanks again for that!

PR #432, opened 2016-07-06
Test rustc download code on arm32 builder

aneeshusa:
We recently had a bug that caused our rustc download code to be broken,
but this was not caught because the builders simply used rustc from
cache instead of downloading fresh. Futhermore, the breakage was only
for cross target downloads, not regular downloads. Hence, unset
SERVO_CACHE_DIR on the arm32 builder in order to force downloading the
host and cross stdlibs for each build (cross builders download both).
(arm32 is the fastest cross builder at ~30 minutes.)

See IRC discussion at http://logs.glob.uno/?c=mozilla%23servo&s=6+Jul+2016&e=6+Jul+2016#c474869.
r? @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/432)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -1,3 +1,5 @@
+import copy
+
 from passwords import GITHUB_DOC_TOKEN
 from passwords import S3_UPLOAD_ACCESS_KEY_ID, S3_UPLOAD_SECRET_ACCESS_KEY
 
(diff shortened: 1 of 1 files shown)


buildbot/master/files/config/environments.py
```diff
@@ -85,8 +99,9 @@ def __add__(self, other):
     'PKG_CONFIG_ALLOW_CROSS': '1',
 })
 
-
-build_arm32 = build_arm + Environment({
+# Unset SERVO_CACHE_DIR to test our download code for host and cross targets.
+# Use arm32 because it is the fastest cross builder.
+build_arm32 = build_arm.without(['SERVO_CACHE_DIR']) + Environment({
```
larsbergstrom: I unfortunately don't think this will work. All that `SERVO_CACHE_DIR` does is allow us to share the cached bits between different builds that run on the same slave. I think that in order to do this we will have to either:
1) Clear out the .servo directory entirely on each run
2) Add a step that does a `./mach boostrap-rust --force`

aneeshusa: Right, but IIRC the default value for this is `.servo` in the current directory, which on Buildbot should be the build directory, which Buildbot wipes during the git step at the start of every build. So by not setting the env var we should get option 1.

larsbergstrom: Aha, you're totally right - good call!

PR #435, opened 2016-07-08
Add geckolib dev build for windows.

emilio:
cc @bholley
r? @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/435)

### DIFF
--- a/buildbot/master/files/config/steps.yml
+++ b/buildbot/master/files/config/steps.yml
@@ -80,6 +80,7 @@ arm64:
 windows-dev:
   - ./mach build --dev
   - ./mach test-unit
+  - ./mach build-geckolib
 
 windows-nightly:
   - ./mach build --release


upsuper:
I guess you probably should put `./mach build-geckolib` after `./mach test-unit`, which is the order elsewhere in that file.

emilio:
@upsuper yep, that makes sense, done.

We *might* want also a release build, though IIRC windows is the slowest machine (so I don't know if it's doable without slowing all the queue).

Probably @larsbergstrom can tell us, and also maybe review it :)

larsbergstrom:
Nah, that's fine. Windows is only 18 minutes (http://build.servo.org/builders/windows/builds/1700), so we've got plenty of time budget to add other things, given the 40 minute long poles of linux & mac release builds w/ WPT testing.

PR #436, opened 2016-07-11
Build geckolib on Windows

upsuper:
In servo/servo#12338, geckolib build is fixed for Windows. This pr is for adding it to Windows builder to prevent regressions on that.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/436)


upsuper:
Hmmm, I did see #435, so it is a duplicate to that.

Issue #337, opened 2016-04-22
buildbot steps need description and descriptionDone


edunham:
Waterfall currently just says "test" on all test steps. It would be better to have a short description of each step, or even just the full text of the command for that step. 

Buildbot docs at http://docs.buildbot.net/latest/manual/cfg-buildsteps.html#common-parameters describe `description` and `descriptionDone`, and [here](https://github.com/rust-lang/rust-buildbot/blob/master/master/master.cfg#L1064) is an example of them in use. 

Steps are created [here](https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/factories.py#L86) with configs drawn from [here](https://github.com/servo/saltfs/blob/master/buildbot/master/files/config/steps.yml). 

@aneeshusa, what's your opinion on whether this should be implemented by adding the information to the yaml file or by inferring it from the commands?

aneeshusa:
I want to keep the YAML file as simple as possible, so I'd say these should be generated from the commands. Given a shell command such as `./mach build --dev`, we can use `Running './mach build --dev'` for the `description`, and simply `./mach build --dev` for `descriptionDone`.

I think this could be added to `make_step` fairly easily.

aneeshusa:
For `description`, we could also have this change based on the `Step` class: `Compiling`, `Testing`, `Running`, and not include the command as the docs say to keep the `description` field as a single imperfect-tense verb.

UK992:
I want to claim this

KiChjang:
@UK992 It's yours!

PR #438, opened 2016-07-13
Add description to buildbot steps

UK992:
Fixes https://github.com/servo/saltfs/issues/337

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/438)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -79,7 +79,7 @@ class DynamicServoFactory(ServoFactory):
 
         command = command.split(' ')
 
-        # Add bash -l before every command on Windows builders
+        # Add `bash -l` before every command on Windows builders
         bash_args = ["bash", "-l"] if self.is_windows else []
         step_kwargs['command'] = bash_args + command
         if self.is_windows:
@@ -90,6 +90,7 @@ class DynamicServoFactory(ServoFactory):
                 ),
             })
 
+        step_desc = []
         step_class = steps.ShellCommand
         args = iter(command)
         for arg in args:
@@ -97,6 +98,7 @@ class DynamicServoFactory(ServoFactory):
             # (steps.Compile and steps.Test catch warnings)
             if arg == './mach':
                 mach_arg = next(args)
+                step_desc = [mach_arg]
                 if re.match('build(-.*)?', mach_arg):
                     step_class = steps.Compile
                 elif re.match('test-.*', mach_arg):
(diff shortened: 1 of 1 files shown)


aneeshusa:
Looking at the docs again and the current waterfall, I actually think a slightly different approach is better here. The raw ShellCommands (not Compile or Test) already put the whole command as the description, so let's not do anything in that case. Furthermore, what we need to do is set `descriptionSuffix` for Compile and Test, which will get appended to the `description` and `descriptionDone`.

We should be smart about setting `descriptionSuffix` to avoid noise; here's some good heuristics:
- No need to include "./mach"
- Include the mach subcommand, like "test-wpt"
- Don't include logging info, like `--log-raw test-wpt.log`
- Keep anything else, e.g. `--release`.

An example:
```
./mach test-wpt --release --processes 24 --log-raw wpt.log --log-errorsummary wpt-errorsummary.log
```
would turn into a `descriptionSuffix` of `test-wpt --release --processes 24`.

Since the `descriptionSuffix` already includes the operation (e.g. `test-wpt` implies testing), I think we could just set `description` to `running`, and `descriptionDone` to `ran`.

Thoughts @edunham ?

aneeshusa:
Also, please make the same change to both `make_step` functions.

UK992:
Done!

buildbot/master/files/config/factories.py
```diff
@@ -110,6 +110,22 @@ def make_step(self, command):
                 step_kwargs['logEnviron'] = False
                 step_env += envs.upload_nightly
 
+        step_desc_suffix = []
```
aneeshusa: I'd prefer to integrate this code into the loop above instead of trying to re-parse things like `--log-.*`. Here's a rough sketch of one way to do it, feel free to modify as needed for ergonomics:
- Create a `step_desc` list before the `for arg in args:` loop.
- Inside the `./mach` block, append both the `./mach` token and the `mach_arg` token to `step_desc`. 
- Don't change the `log` block, so those tokens won't get added to the description.
- Add an `else` or similar at the end to append any other tokens to the description. (I think you might have to change the last `elif` to accommodate this properly.)

buildbot/master/files/config/factories.py
```diff
@@ -82,7 +82,7 @@ def make_step(self, command):
 
         command = command.split(' ')
 
-        # Add bash -l before every command on Windows builders
+        # Add `bash -l` before every command on Windows builders
```
aneeshusa: Please make the same change in the other `make_step` function.

UK992:
Done. 
Now command like: 
```./mach test-wpt --release --binary-arg=--multiprocess --processes 24 --log-raw test-wpt-mp.log --log-errorsummary wpt-mp-errorsummary.log eventsource```

will be turned into: ```test-wpt --release --binary-arg=--multiprocess --processes 24 eventsource```

It is ok to keep ``eventsource`` or should be also ignored?

edunham:
Looks good to me; r=me with the merge conflict fixed

UK992:
rebased.

aneeshusa:
r- from me on this, this doesn't look like it works (confirmed with a quick Python 2 REPL), but leaves the description as just one token.

I will follow up with details shortly.

UK992:
i fixed missing ``+`` in ``step_desc``, if that what you think.

jdm:
This is waiting on feedback from @aneeshusa.

aneeshusa:
Sorry about the delay, looks good to me. @bors-servo r+

Issue #303, opened 2016-04-06
Use UTC Timezone


aneeshusa:
This should be applied to all machines, so this should go into the `common/init.sls` file.

We can use the `timezone.system` state to configure a UTC timezone ([docs](https://docs.saltstack.com/en/2015.5/ref/states/all/salt.states.timezone.html)).

I think this would be a great E-Easy issue.

(A follow-up would be ensuring that NTP is on and configured on all machines, but there's no Salt state for this yet so that wouldn't be E-Easy.)

RowanEB:
I'll start working on this.

aneeshusa:
It's all yours! Let me know if you have any questions.

RowanEB:
I think I've correctly added this, should I go about working on adding NTP support, or will that be done from a separate issue?

aneeshusa:
Let's save NTP for another issue/PR; that one will be more involved because we will need to write custom Salt code for that. (It's also blocked on #350, which will enable us to load custom modules and states).

aneeshusa:
@charlesvdv is interested in taking this up now: https://github.com/servo/saltfs/pull/439#issuecomment-269638754

jdm:
Looks like this was fixed by #569.

PR #439, opened 2016-07-15
Made UTC timezone explicit in common/init.sls.

RowanEB:
Relevant to #303.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/439)

### DIFF
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -53,6 +53,6 @@ else
     # Only run tests against the new configuration
     # TODO: don't hard-code this
     if [[ "${SALT_NODE_ID}" == "servo-master1" ]]; then
-        ./test.py sls.buildbot.master sls.homu sls.nginx
+        ./test.py sls.buildbot.master sls.common.timezone sls.homu sls.nginx
     fi
 fi
--- a/common/init.sls
+++ b/common/init.sls
@@ -40,6 +40,9 @@ servo:
     - shell: /bin/bash
     - home: {{ common.servo_home }}
 
+Etc/UTC:
+    timezone.system
+
 {% for hostname, ip in common.hosts.items() %}
 host-{{ hostname }}:
   host.present:
--- a/common/init.sls
+++ b/common/init.sls
@@ -40,7 +40,7 @@ servo:
     - shell: /bin/bash
     - home: {{ common.servo_home }}
 
-Etc/UTC:
+UTC:
     timezone.system
 
 {% for hostname, ip in common.hosts.items() %}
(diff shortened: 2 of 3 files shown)


common/init.sls
```diff
@@ -40,6 +40,9 @@ servo:
     - shell: /bin/bash
     - home: {{ common.servo_home }}
 
+Etc/UTC:
```
aneeshusa: This can just be `UTC`.

aneeshusa:
Please also add a test for this. I think running the `date` command and ensuring `UTC` is in the output will work.

Add a new directory `tests/sls/common` and put the test in a new file `tests/sls/common/timezone.py`. Also, please make an empty `tests/sls/common/__init__.py` file, and update the `.travis/dispatch.sh` code to run the test on all nodes.

RowanEB:
I've added the test and simplified to just UTC.

.travis/dispatch.sh
```diff
@@ -53,6 +53,6 @@ else
     # Only run tests against the new configuration
     # TODO: don't hard-code this
     if [[ "${SALT_NODE_ID}" == "servo-master1" ]]; then
-        ./test.py sls.buildbot.master sls.homu sls.nginx
+        ./test.py sls.buildbot.master sls.common.timezone sls.homu sls.nginx
```
aneeshusa: Two things:
- this can be just `sls.common`, all of the test files in that dir will get run automatically.
- we want to run the common tests on all nodes, not just `servo-master1`, so that should happen outside the conditional.

tests/sls/common/timezone.py
```diff
@@ -0,0 +1,18 @@
+import subprocess
+
+from tests.util import Failure, Success
+
+
+def run():
+    command = "date | grep -v UTC"
```
aneeshusa: Instead of using pipes and an extra `grep` process, let's just look for `UTC` in the output of the `date` command inside Python.

tests/sls/common/timezone.py
```diff
@@ -0,0 +1,18 @@
...
+    ret = subprocess.run(command,
+                         stdout=subprocess.PIPE,
+                         stderr=subprocess.PIPE,
+                         universal_newlines=True,
+                         shell=True)
```
aneeshusa: Please remove `shell=True`, it can cause security problems (and also spawns an extra process).

KiChjang:
@RowanEB Are you still planning on finishing this up?

charlesvdv:
@aneeshusa @KiChjang I'm interested in continuing the work of @RowanEB if that's ok for you ? (as he doesn't respond in a long time)

aneeshusa:
@charlesvdv please go ahead!

PR #441, opened 2016-07-19
Fix Windows HOME directory to have the proper builder name

larsbergstrom:
The change from the builder name of `windows` to `windows-dev` means that many commands were being executed in the incorrect directory. This should fix that!

r? @edunham @aneeshusa 

cc @Ms2ger

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/441)


larsbergstrom:
r? @Ms2ger 

This one finally seemed to pass travis servo-master1 testing :-)

PR #442, opened 2016-07-19
Windows changes to make the HOME env var work

larsbergstrom:
There are two big changes here:

1) Make setting the `HOME` env var work properly, filling in the builder name
2) Just allow exceptions to fail to load buildbot and log normally. Right now, the `print` is being eaten by something, so it's super hard to debug failures with dynamic steps.

r? @edunham

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/442)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -59,16 +59,13 @@ class DynamicServoFactory(ServoFactory):
     def __init__(self, builder_name, environment):
         self.environment = environment
         self.is_windows = re.match('windows.*', builder_name) is not None
-        try:
-            config_dir = os.path.dirname(os.path.realpath(__file__))
-            yaml_path = os.path.join(config_dir, 'steps.yml')
-            with open(yaml_path) as steps_file:
-                builder_steps = yaml.safe_load(steps_file)
-            commands = builder_steps[builder_name]
-            dynamic_steps = [self.make_step(command) for command in commands]
-        except Exception as e:  # Bad step configuration, fail build
-            print(str(e))
-            dynamic_steps = [BadConfigurationStep(e)]
+        self.builder_name = builder_name
+        config_dir = os.path.dirname(os.path.realpath(__file__))
+        yaml_path = os.path.join(config_dir, 'steps.yml')
+        with open(yaml_path) as steps_file:
+            builder_steps = yaml.safe_load(steps_file)
+        commands = builder_steps[builder_name]
+        dynamic_steps = [self.make_step(command) for command in commands]
 
         pkill_step = [self.make_pkill_step("servo")]
 
@@ -85,6 +82,13 @@ class DynamicServoFactory(ServoFactory):
         # Add bash -l before every command on Windows builders
         bash_args = ["bash", "-l"] if self.is_windows else []
         step_kwargs['command'] = bash_args + command
+        if self.is_windows:
+            step_env += envs.Environment({
+                # Set home directory, to avoid adding `cd` command every time
+                'HOME': r'C:\buildbot\slave\{}\build'.format(
+                    self.builder_name
+                ),
+            })
 
         step_class = steps.ShellCommand
         args = iter(command)
@@ -109,6 +113,17 @@ class DynamicServoFactory(ServoFactory):
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
                 step_env += envs.upload_nightly
+                if self.is_windows:
+                    # s3cmd on Windows only works within msys
+                    step_env['MSYSTEM'] = 'MSYS'
+                    step_env['PATH'] = ';'.join([
+                        r'C:\msys64\usr\bin',
+                        r'C:\Windows\system32',
+                        r'C:\Windows',
+                        r'C:\Windows\System32\Wbem',
+                        r'C:\Windows\System32\WindowsPowerShell\v1.0',
+                        r'C:\Program Files\Amazon\cfn-bootstrap',
+                    ])
 
         step_kwargs['env'] = step_env
         return step_class(**step_kwargs)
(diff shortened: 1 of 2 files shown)


aneeshusa:
I mentioned this in https://github.com/servo/saltfs/issues/316#issuecomment-232344297 but unswallowing the exception here won't help. The exception Buildbot encounters is actually internal and occurs after our dynamic step is finished, when Buildbot tries to load the next step. We aren't respecting some of Buildbot 8's internal invariants, causing the exception. You can find out more about this by looking at the log files/digging into the source code of buildbot and the buildbot-travis project I linked.

buildbot/master/files/config/factories.py
```diff
@@ -85,6 +82,12 @@ def make_step(self, command):
         # Add bash -l before every command on Windows builders
         bash_args = ["bash", "-l"] if self.is_windows else []
         step_kwargs['command'] = bash_args + command
+        if self.is_windows:
+            step_env += envs.Environment({
+                # Set home directory, to avoid adding `cd` command every time
+                'HOME': r'C:\buildbot\slave\{}\build'.format(
+                self.builder_name),
```
aneeshusa: Add 4 spaces of indentation.

buildbot/master/files/config/factories.py
```diff
@@ -182,10 +185,12 @@ def make_step(self, command):
...
         step_kwargs['command'] = bash_command + command
-        step_env += envs.Environment({
-            # Set home directory, to avoid adding `cd` command on every command
-            'HOME': r'C:\buildbot\slave\{}\build'.format(self.builder_name),
-            })
+        if self.is_windows:
+            step_env += envs.Environment({
+                # Set home directory, to avoid adding `cd` command every time
+                'HOME': r'C:\buildbot\slave\{}\build'.format(
+                self.builder_name),
```
aneeshusa: Add 4 spaces of indentation.

buildbot/master/files/config/factories.py
```diff
@@ -85,6 +82,12 @@ def make_step(self, command):
         # Add bash -l before every command on Windows builders
         bash_args = ["bash", "-l"] if self.is_windows else []
         step_kwargs['command'] = bash_args + command
+        if self.is_windows:
+            step_env += envs.Environment({
+                # Set home directory, to avoid adding `cd` command every time
+                'HOME': r'C:\buildbot\slave\{}\build'.format(
+                self.builder_name),
+                })
```
aneeshusa: Subtract 4 spaces of indentation.

buildbot/master/files/config/factories.py
```diff
@@ -182,10 +185,12 @@ def make_step(self, command):
...
-        step_env += envs.Environment({
-            # Set home directory, to avoid adding `cd` command on every command
-            'HOME': r'C:\buildbot\slave\{}\build'.format(self.builder_name),
-            })
+        if self.is_windows:
+            step_env += envs.Environment({
+                # Set home directory, to avoid adding `cd` command every time
+                'HOME': r'C:\buildbot\slave\{}\build'.format(
+                self.builder_name),
+                })
```
aneeshusa: Subtract 4 spaces of indentation.

buildbot/master/files/config/factories.py
aneeshusa: Also, the `),` can then go on the next line for balance.

buildbot/master/files/config/factories.py
aneeshusa: Also, the ), can then go on the next line for balance.

larsbergstrom:
Latest commit fixes https://github.com/servo/servo/issues/12231

buildbot/master/files/config/factories.py
```diff
@@ -215,6 +215,11 @@ def make_step(self, command):
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
                 step_env += envs.upload_nightly
+                if self.is_windows:
+                    # s3cmd on Windows only works within msys
+                    step_env += envs.Environment({
+                        'MSYSTEM': 'MSYS',
+                        })
```
aneeshusa: -4 spaces indentation please

aneeshusa: Also please add this to the other `make_step` function.

larsbergstrom:
OK, I *think* this addresses all of the formatting requests.

Note that I left the removal of the `print` statement and `try` block around the dynamic steps because otherwise right now there's no way (AFAIK) to debug the exception, as it gets swallowed, printed to /dev/null, and the only way you can tell is when you queue a job to that builder the steps look "off." @edunham mentioned she might switch it to use buildbot's logging infrastructure instead.

aneeshusa:
Please keep the print and try block intact. The log you need to look at is `/home/servo/buildbot/master/twistd.log` on `servo-master1`; search for `setText` to jump to the exception trace. Removing the try will not help because it exception is not thrown within our dynamic step, but after that step is finished, in the internals of buildbot when it tries to start the next step.

+1 for integrating with buildbot's logging infrastructure in the future though.

aneeshusa:
BTW, I started looking into that exception if you want to chat on IRC about it.

buildbot/master/files/config/factories.py
```diff
@@ -109,6 +113,17 @@ def make_step(self, command):
             elif arg == './etc/ci/upload_nightly.sh':
                 step_kwargs['logEnviron'] = False
                 step_env += envs.upload_nightly
+                if self.is_windows:
+                    # s3cmd on Windows only works within msys
+                    step_env['MSYSTEM'] = 'MSYS'
```
aneeshusa: Use another `envs.Environment` here.

larsbergstrom: hrm, when I was using `envs.Environment`, the debugging seemed to indicate that it wasn't overwriting it - we were still getting the old value in testing. I can give it another try, though, and see what happens.

aneeshusa: This is a stylistic change we can always make later, so I'm OK with this as-is if we need the PR for windows nightlies now.

aneeshusa:
After some IRC chat it turns out I was looking at the wrong thing for the exception handling and removing the try is the best thing to do here; this is r=me with or without using an `envs.Environment` for the `MSYSTEM` and `PATH` variables.

buildbot/master/files/config/factories.py
larsbergstrom: Yeah, it's already in deployment as-formatted, so I'm kinda tempted to get this landed (since I already feel bad about having an ad-hoc branch in production).

PR #464, opened 2016-08-18
Set token for homebrew repository

### LINKED ISSUE #463: update_brew.sh: line 19: TOKEN: unbound variable
buildbot/master/files/config/environments.py set `TOKEN` for `upload_nightly`, but `TOKEN` is used in `update_brew.sh`.

Should we create an environments for update_brew? Or use `GITHUB_HOMEBREW_TOKEN` in `update_brew.sh` (would that work?)

paulrouget:
Fix #463 

I don't know how to test locally. I just assume that's enough.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/464)


paulrouget:
(failures look unrelated to my changes)

larsbergstrom:
Yes, this is right! I'd assumed we set all the tokens for all the builds on the nightly. Great catch & fix.

PR #460, opened 2016-08-16
Add brew scripts to nightly

paulrouget:
Follow up of https://github.com/servo/servo/pull/12688

Also, see https://github.com/servo/servo/issues/12888

/cc @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/460)

PR #465, opened 2016-08-18
Add colon for python if

larsbergstrom:
r? @edunham 

cc @paulrouget

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/465)

PR #467, opened 2016-08-18
Add fake GitHub homebrew token

aneeshusa:
Fixes the buildbot master file.recurse failures we've been seeing recently.
Also updates the highstate output to increase the SNR.
See commit messages for details.

Follow-up to #460. This failure was missed because we were experiencing pip failures at the time, causing Travis bustage, and the PR was merged by hand without noticing the additional failures on Travis.

r? @larsbergstrom @edunham 

cc @paulrouget

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/467)

### DIFF
--- a/.travis/minion
+++ b/.travis/minion
@@ -1,3 +1,3 @@
 master: localhost  # Used to test the salt-master
-state_output: terse
+state_output: mixed
 state_tabular: True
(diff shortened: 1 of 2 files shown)


aneeshusa:
#465 should be merged first to fix the syntax error.

aneeshusa:
Also note that the scratch builder is the important Travis result; the other one (that builds the old config then the new config) will fail because the old config is busted.

aneeshusa:
@larsbergstrom please larsbors this, the failure is expected since the old config is busted

PR #471, opened 2016-08-22
Add Vagrant version upper bound of "< 1.8.3"

aneeshusa:
A recent PR to Vagrant, which is in released 1.8.3+, breaks our usage of
`minion_id` to pass additional args to Salt.
See https://github.com/mitchellh/vagrant/pull/7207 for the PR.

Also add comments in the Vagrantfile about the version requirements.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/471)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -17,7 +17,11 @@ def is_salt_master(id)
   !id.match(/servo-master\d+/).nil?
 end
 
-Vagrant.require_version '>= 1.8.0'
+# Need Vagrant >= 1.8.0, in which the Vagrant Salt provisioner was overhauled
+# See https://github.com/servo/saltfs/pull/180
+# Vagrant 1.8.3+ breaks our usage of minion_id to pass additional args to Salt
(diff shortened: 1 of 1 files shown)


aneeshusa:
Wiki changes have been made in https://github.com/servo/servo/wiki/Buildbot-administration/_compare/93246a24380b22cfcee6009849b8e43a24652503...276f2c110c6aad1767ef0b164289a1b945de840c.

metajack:
@bors-servo r+

Should we change our code so that 1.8.4 and later work?

aneeshusa:
@metajack That would be nice, but Vagrant 1.8.3+ no longer has a way of passing arbitrary args to Salt (at least, that I know of). Being able to use newer versions of Vagrant will require a PR to Vagrant to add `file_root` and `pillar_root` options as I mentioned [here](https://github.com/servo/saltfs/pull/248#issuecomment-240919119), and then having that PR be included in a future Vagrant version.

PR #472, opened 2016-08-23
Fix buildbot step description for shell commands

UK992:
This should fix description for shell commands where after https://github.com/servo/saltfs/pull/438 shows only "ran", instead of command.
r? @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/472)


aneeshusa:
@bors-servo r+

Thanks for catching that, I missed that during review!

PR #484, opened 2016-09-22
Add servo/osmesa-src

larsbergstrom:
r? @aneeshusa @edunham

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/484)

PR #483, opened 2016-09-22
Fix Homebrew 1.0.0 fallout

aneeshusa:
Homebrew 1.0.0 has just been released with some breaking changes. To handle all of them, this PR has a lot of changes, here are the main points:
- Add a new `homebrew` execution module and `hombrew_analytics` state module to fix analytics disabling
- Fix `autoconf`/`autoconf213` installing and linking
- Use env vars to specify OpenSSL location because linking fails (fixes #479) 
- Upgrade to XCode 8/OS X 10.11 on Travis (fixes #382) to fix installing git
- Update Salt, leaping two major versions ahead to 2016.3.3 (the latest), due to changes in Homebrew Python/Salt packaging not available in our pinned old Salt
- Make our install and dispatch scripts more robust:
  - Force apt-get to use existing configuration files and ignore updates in Salt packages during installation
  - Unlink Salt if we have installed it via Homebrew previously before installing a new one
  - Allow failures when building old revs when checking upgrades (upstreams change)
  - Manually invalidate the Salt cache between runs

See commit messages for full details.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/483)

PR #486, opened 2016-09-23
Fix homebrew analytics directory for brew 1.0

larsbergstrom:
r? @aneeshusa @edunham 

Homebrew in version 1.0 has moved their git repository to `/usr/local/Homebrew` to avoid the various problems with Apple fooling around with `/usr/local`. See https://github.com/Homebrew/install/pull/60

Note that I *believe* we could also set `HOMEBREW_DISABLE_ANALYTICS=1`, but there's a risk that any `brew` command not done with that set would still perform analytics reporting.

This bug is currently preventing us from landing other saltfs changes. See https://github.com/servo/saltfs/pull/484

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/486)

### DIFF
--- a/osx/init.sls
+++ b/osx/init.sls
@@ -22,4 +22,4 @@ disable-homebrew-analytics:
   git.config:
     - name: 'homebrew.analyticsdisabled'
     - value: 'true'
-    - repo: /usr/local
+    - repo: /usr/local/Homebrew


larsbergstrom:
Well, that fixed the crash, but the two OSX builds still hit:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
```

aneeshusa:
IMO #483 is a better fix, but currently hitting the same problems. Since they work in different ways, I think this may be an intermittent issue with Travis ATM.

aneeshusa:
That being said, I am OK with landing this as a quicker-fix in the meantime if it starts working, but we need to be careful about upgrading Homebrew (to >= 1.0.0) on all of our Mac builders so this state works properly.

larsbergstrom:
Aha! I didn't see #483 for some reason and apologize. Would you rather land that one? I'm happy to close this and manually-land that one instead.

aneeshusa:
Well, #483 is untested code I worked on at 3 AM, so I don't trust it one bit until it passes Travis. If you want to land something right now, I'd feel comfortable hand-merging this PR, although we'll need to keep landing manually until the issue gets fixed. We just need to upgrade Homebrew on all the Macs for this one.

(Also, don't forget that with gitfs manual deploys have a different process now.)

larsbergstrom:
Sounds good - thanks! Then I'll merge this (to unblock the other PR for osmesa-src), and hopefully we can look more at #483 once things are less of a hot mess on macOS for travis.

aneeshusa:
I just realized we could have just merged the osmesa-src PR directly, since this one is also broken...

PR #491, opened 2016-09-27
Add an MSVC builder and rename the old one as GNU, but do not gate.

larsbergstrom:
This does a few things:
1) Adds a `servo-windows2` buildbot
2) Adds a `-msvc` version of our build rules
3) Renames the default windows build to `-gnu`

r? @aneeshusa or @edunham

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/491)


buildbot/master/files/config/master.cfg
```diff
@@ -179,8 +181,9 @@ c['builders'] = [
     DynamicServoBuilder("mac-rel-css", MAC_SLAVES, envs.build_mac),
     DynamicServoBuilder("mac-rel-intermittent", MAC_SLAVES, envs.build_mac),
     DynamicServoBuilder("mac-rel-wpt", MAC_SLAVES, envs.build_mac),
-    DynamicServoBuilder("windows-dev", WINDOWS_SLAVES, envs.build_windows),
-    DynamicServoBuilder("windows-nightly", WINDOWS_SLAVES, envs.build_windows),
+    DynamicServoBuilder("windows-dev-gnu", WINDOWS_SLAVES, envs.build_windows_gnu),
+    DynamicServoBuilder("windows-dev-msvc", WINDOWS_SLAVES, envs.build_windows_msvc),
```
aneeshusa: Our factories have extra logic for `"windows-*"` builders, and I don't think it would all apply to MSVC. You'll likely need to either change that logic or use a raw util.BuilderConfig with a different factory (similar to the doc builder).

edunham:
Reviewable has not handled the rebasing and force pushing here gracefully at all. I've inspected the GitHub diffs, since they show latest commit vs master, and it looks like you've successfully addressed Aneesh's concerns about the windows.* logic in the factories. Assuming the path wizardry is correct, this looks good to me.

---

Reviewed 2 of 4 files at r2, 1 of 3 files at r3, 1 of 3 files at r4, 2 of 2 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[buildbot/master/files/config/environments.py, line 74 at r2](https://reviewable.io:443/reviews/servo/saltfs/491#-KTjT_OHVISQIhmu3lmI:-KTjT_OINoXiJkWh0rUA:b3dpfng) ([raw file](https://github.com/servo/saltfs/blob/02aa806eafbe354cb865dfa993b66cfc266e3648/buildbot/master/files/config/environments.py#L74)):*
> ```Python
>         r'C:\Program Files\Amazon\cfn-bootstrap',
>         r'C:\Program Files\Git\cmd',
>         r'C:\Program Files (x86)\WiX Toolset v3.10\bin',
> ```

I'll take your word for it that these paths are correct; I haven't dug around in the Windows builders enough to recognize if they were wrong

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/servo/saltfs/491#-:-KTjUDLRub7glOAUA2Xd:b-5ki3r9)*

aneeshusa:
In some places you've updated one implementation but not the other, please make sure they both match so we don't have strange regressions after #504.

aneeshusa:
Didn't see you had updated this, will review after my exam.

larsbergstrom:
No worries! I just did it and need to redo it (splitting it into servo/servo, too), as I just bitrotted it by landing your other fix, which I'd unfortunately not reviewed while I was out sick for most of the last week.

Possibly I can pre-land the "extra" builders into servo/servo first and if that doesn't break anything then I can land the salt/buildbot-side changes second without breaking things. And then a follow-up in servo/servo to remove the now-deprecated builders?

aneeshusa:
I'd like to get #504 deployed and working before we change the builder/steps list again, in case we need to do a rollback/revert/quickfix. But yes, that 3-step process sounds good.

aneeshusa:
Filed https://github.com/servo/servo/issues/13838 to make it easier to change `buildbot_steps.yml` in the future without causing breakage.

larsbergstrom:
r? @edunham @aneeshusa 

I think this'll do it for adding the new normal and nightly builders. Note that it can't be deployed until I land the corresponding servo-side changes to the steps.yml file (which I'm currently testing and should have up shortly).

buildbot/master/files/config/factories.py
```diff
@@ -211,10 +213,11 @@ def make_step(self, command):
                 step_kwargs['logfiles'][logfile] = logfile
 
             # Provide environment variables for s3cmd
-            elif arg == './etc/ci/upload_nightly.sh':
+            elif arg == './etc/ci/upload_nightly.sh' or
+                 next(args) == r'.\etc\ci\upload_nightly.sh':
```
aneeshusa: Why use `next(args)` here instead of `arg`? Seems like a bug.

buildbot/master/files/config/factories.py
```diff
@@ -211,10 +213,11 @@ def make_step(self, command):
...
                 step_kwargs['logEnviron'] = False
                 step_env += envs.upload_nightly
-                if self.is_windows:
+                if self.is_win_gnu:
                     # s3cmd on Windows only works within msys
```
aneeshusa: No idea, but should this comment be updated to reference Windows GNU only?

buildbot/master/files/config/factories.py
larsbergstrom: I was originally using `python .\etc\ci\upload_nightly.sh` in the buildbot_steps.yml file :-) I just found out that it works OK without the python prefix, so I'll remove the next and updated the servo-side PR to the yml file.

buildbot/master/files/config/factories.py
```diff
@@ -90,6 +90,8 @@ def setDefaultWorkdir(self, workdir):
     @defer.inlineCallbacks
     def run(self):
         self.is_windows = re.match('windows.*', self.builder_name) is not None
+        self.is_win_gnu = re.match('windows.*gnu', builder_name) is not None
+        self.is_win_msvc = re.match('windows.*msvc', builder_name) is not None
```
larsbergstrom: Removing.

buildbot/master/files/config/factories.py
larsbergstrom: Yes, addressed.

buildbot/master/files/config/factories.py
aneeshusa: Wait, that's definitely not a Python script but bash - will bash scripts run properly on MSVC?

larsbergstrom:
Also, w.r.t. the question, I tested the bash script on my Windows box and CMD.EXE and it all works. We'll have to see how things turn out on the builders themselves, but I expect to see a bunch of iteration on the packaging steps for the MSVC build after this lands anyway (e.g., it currently shares with the `-gnu` version, so it packages a bunch of binaries that are not required). I just wanted to have a first version available to iterate on :-)

buildbot/master/files/config/factories.py
```diff
@@ -90,6 +90,7 @@ def setDefaultWorkdir(self, workdir):
     @defer.inlineCallbacks
     def run(self):
         self.is_windows = re.match('windows.*', self.builder_name) is not None
+        self.is_win_gnu = re.match('windows.*gnu', builder_name) is not None
```
aneeshusa: `self.builder_name`

aneeshusa: Also, with the modification in https://github.com/servo/servo/pull/14651#issuecomment-268398195, these regexes can just become `windows` and `windows-gnu`, as `re.match` only checks the beginning of the string for a match and ignores any further characters.

aneeshusa: Also, you can remove the `is not None` for both of these, which will let them fit under the line limit.

buildbot/master/files/config/factories.py
```diff
@@ -211,11 +212,12 @@ def make_step(self, command):
                 step_kwargs['logfiles'][logfile] = logfile
 
             # Provide environment variables for s3cmd
-            elif arg == './etc/ci/upload_nightly.sh':
+            elif arg == './etc/ci/upload_nightly.sh' or
+                 arg == r'.\etc\ci\upload_nightly.sh':
```
aneeshusa: I recommend surrounding this in brackets and using block ident:
```python
elif (
    arg == './etc/ci/upload_nightly.sh'
    or arg == '.\etc\ci\upload_nightly.sh'
):
    step_kwargs['logEnviron'] = ...
```
Another option that IMO looks nicer (although loses short-circuiting) is:
```python
elif any([
    arg == './etc/ci/upload_nightly.sh',
    arg == r'.\etc\ci\upload_nightly.sh',
]):
    step_kwargs['logEnviron'] = ...
```

buildbot/master/files/config/master.cfg
```diff
@@ -101,8 +102,10 @@ c['schedulers'].append(schedulers.ForceScheduler(
         "mac-rel-intermittent",
         "mac-rel-wpt1",
         "mac-rel-wpt2",
-        "windows-dev",
-        "windows-nightly",
+        "windows-dev-gnu",
```
aneeshusa: Please change to `windows-gnu-dev` per https://github.com/servo/servo/pull/14651#issuecomment-268398195, as well as the other instances.

PR #493, opened 2016-09-27
Vagrant: Use `salt_call_args` instead of minion_id

jarondl:
The hacky `minion_id` method to pass `file-root` and `pillar-root` to
`salt-call` does not work since Vagrant 1.8.3.

If Vagrant accepts mitchellh/vagrant#7836 , this PR will use the new
argument. It also sets the hostname using the appropriate option,
superseeding #248.

DO NOT MERGE until Vagrant accepts the PR, and releases a new version.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/493)

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,9 +19,8 @@ end
 
 # Need Vagrant >= 1.8.0, in which the Vagrant Salt provisioner was overhauled
 # See https://github.com/servo/saltfs/pull/180
-# Vagrant 1.8.3+ breaks our usage of minion_id to pass additional args to Salt
-# See https://github.com/mitchellh/vagrant/pull/7207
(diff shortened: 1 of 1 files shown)


aneeshusa:
The upstream PR mitchellh/vagrant#7836 was superseded by mitchellh/vagrant#8927, which has been merged, so this should be available when the next version of Vagrant is released.

@jarondl, are you interested in reviving this PR when that happens? It will need to be updated and rebased on top of master, and the docs will need to be updated (the `docs/salt.md` file in this repo) as welll.

aneeshusa:
Superseded by #729. Thanks again @jarondl for your initial work here!

PR #492, opened 2016-09-27
Remove backported modules now included in our Salt

aneeshusa:
These modules were backported from Salt 2015.8 or Salt 2016.3, but they
should no longer be necessary now that we are on Salt 2016.3.3.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/492)

PR #497, opened 2016-09-29
Split the macOS WPT into two chunks an add another builder

larsbergstrom:
r? @aneeshusa 

cc @jgraham 

This splits the WPT tests into two chunks, using the default strategy. The first chunk is significantly smaller than the second (900s vs 1800s, locally), so I've loaded the extra tasks onto that machine. The other chunking strategy (hash) is not yet fully supported in WPT, but we might want to go to it after this to provide more even partitioning.

I've also added a name for another mac mini builder and will probably bring such a machine online shortly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/497)

### DIFF
--- a/buildbot/master/files/config/steps.yml
+++ b/buildbot/master/files/config/steps.yml
@@ -1,11 +1,15 @@
-mac-rel-wpt:
+mac-rel-wpt1:
   - ./mach build --release
   - ./mach test-wpt-failure
-  - ./mach test-wpt --release --processes 8 --log-raw test-wpt.log --log-errorsummary wpt-errorsummary.log
+  - ./mach test-wpt --release --processes 8 --total-chunks 2 --this-chunk 1 --log-raw test-wpt.log --log-errorsummary wpt-errorsummary.log
   - ./mach test-wpt --release --binary-arg=--multiprocess --processes 8 --log-raw test-wpt-mp.log --log-errorsummary wpt-mp-errorsummary.log eventsource
   - ./mach build-cef --release
   - bash ./etc/ci/lockfile_changed.sh
   - bash ./etc/ci/manifest_changed.sh
+
+mac-rel-wpt2:
+  - ./mach build --release
+  - ./mach test-wpt --release --processes 8 --total-chunks 2 --this-chunk 2 --log-raw test-wpt.log --log-errorsummary wpt-errorsummary.log
 
 mac-dev-unit:
   - ./mach build --dev
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -9,7 +9,7 @@ from passwords import HOMU_BUILDBOT_SECRET, GITHUB_STATUS_TOKEN
 
 
 LINUX_SLAVES = ["servo-linux1", "servo-linux2", "servo-linux3"]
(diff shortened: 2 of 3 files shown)


buildbot/master/files/config/steps.yml
```diff
@@ -1,12 +1,16 @@
-mac-rel-wpt:
+mac-rel-wpt1:
   - ./mach build --release
   - ./mach test-wpt-failure
-  - ./mach test-wpt --release --processes 8 --log-raw test-wpt.log --log-errorsummary wpt-errorsummary.log
+  - ./mach test-wpt --release --processes 8 --total-chunks 2--this-chunk 1 --log-raw test-wpt.log --log-errorsummary wpt-errorsummary.log
```
jdm: Need a space after `2`.

larsbergstrom:
Great catch - thanks!

jgraham:
https://bugzilla.mozilla.org/show_bug.cgi?id=1305877 is the gecko bug to improve the hash chunker fwiw.

notriddle:
Is there any way we could compile Servo once, then copy the binary over to the other Macs to run the tests? It wouldn't speed up the fully-linear case, but it would help when there's a try build and an auto build running in parallel.

larsbergstrom:
@notriddle I *think* that we could do this with `Dependent` schedulers (http://docs.buildbot.net/current/manual/cfg-schedulers.html#dependent-scheduler ), but I'm frankly terrified of how long it would take us to test and roll this out. Perhaps in Hawaii when we have a bunch of the team members together in one place for a bunch of hands poking at it at the same time :-)

aneeshusa:
r=me for manual merge, as only the macOS upgrade build is failing, which is currently expected.

@larsbergstrom, if you're setting up a new Mac builder I *highly* recommend
1) Using an older revision of saltfs (i.e. before #483)
2) Using `salt-call` on the new builder instead of `salt` on the master to run Salt completely locally, allowing us to use the known good older configuration.
Ask me on IRC if you have questions about how to do that.

larsbergstrom:
Good points - thanks for the tips! I will bring the machine up first using your instructions BEFORE attempting to merge this and turn over buildbot.

larsbergstrom:
Trying to follow: https://docs.saltstack.com/en/latest/topics/tutorials/standalone_minion.html

I have a local clone of the saltfs repo, checked out at the older version of salt (and older satl installed. Getting an error now, but still working on it...

```
server:saltfs administrator$ sudo salt-call state.apply --local test=True
[INFO    ] Loading fresh modules for state activity
local:
    ----------
    no_|-states_|-states_|-None:
        ----------
        __run_num__:
            0
        changes:
            ----------
        comment:
            No Top file or external nodes data matches found
        name:
            No States
        result:
            False
```

larsbergstrom:
Aha! I needed `--file-root=/Users/administrator/saltfs`.

larsbergstrom:
Hrm, it's running, but I'm gonna need some stuff from the pillar (e.g., the password to connect to the buildbot)...

larsbergstrom:
Yup, got it after making a local pillar!

larsbergstrom:
Merging this manually until #492 is happy in Travis, so that I can update the buildbot master.

aneeshusa:
Yep, those are the two extra params needed, glad to have some extra capacity :)

PR #501, opened 2016-10-03
Respect Buildbot invariants when adding new steps

aneeshusa:
This will prevent crashes when Buildbot tries to execute dynamically
added steps.

I'm not sure if I got everything needed, so this may require some followups.

Part of #316.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/501)


aneeshusa:
OK, fixed the syntax bugs and tidy.

Issue #372, opened 2016-05-13
Use steps.Compile for `mach package` on Buildbot


aneeshusa:
We use `./mach package -r` to [package Android APKs](https://github.com/servo/saltfs/blob/783e4f66c2fce1bc23176619b30bf7bf4133bb6c/buildbot/master/files/config/steps.yml#L56) on the android-nightly builder; the next step is then to upload the built package via S3.

Currently, the packaging step is broken (should be fixed by #290); however, even though it fails, Buildbot still tries to run the next step of uploading to S3. This is because there is no heuristic for `./mach package` and it is run as a regular `ShellCommand`.

Instead, we should add a heuristic in the `make_step` function in the `buildbot/master/files/config/factories.py` file that checks for `./mach package` and uses steps.Compile to create that step. This a) is a closer semantic match (we're "compiling" the APK), and b) will stop the build if the packaging fails, so Buildbot doesn't try the S3 upload step.

This can simply check for `package` as an argument to mach instead of using a more complicated regex as the other heuristics do, because currently there is only one type of packaging: Android APK.

lucasloisp:
I would like to implement this. If I understand correctly, we should check that `mach_arg` re.matches "package"` and assign `step_class = steps.Compile`. Right?
I'll get to it later today

aneeshusa:
@lucasloisp that's correct! I've assigned this to you, leave a comment if you have any questions.

Note: there are currently two copies of `make_step`, make sure to update both of them.

jdm:
Fixed by #503.

PR #503, opened 2016-10-05
`mach package` treated as a steps.Compile

lucasloisp:
@aneeshusa: as per issue #372, both `make_step` now have a heuristic for checking that the argument passed to mach (`mach_arg`) is `'package'` and assigns `step_class = steps.Compile`

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/503)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -106,6 +106,8 @@ class DynamicServoFactory(ServoFactory):
                 step_desc = [mach_arg]
                 if re.match('build(-.*)?', mach_arg):
                     step_class = steps.Compile
+                elif re.match('package', mach_arg):
+                    step_class = steps.Compile
                 elif re.match('test-.*', mach_arg):
                     step_class = steps.Test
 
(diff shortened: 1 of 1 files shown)


buildbot/master/files/config/factories.py
```diff
@@ -106,6 +106,8 @@ def make_step(self, command):
                 step_desc = [mach_arg]
                 if re.match('build(-.*)?', mach_arg):
                     step_class = steps.Compile
+                if re.match('package', mach_arg):
```
aneeshusa: nit: use `elif` instead of `if`

buildbot/master/files/config/factories.py
```diff
@@ -251,6 +253,8 @@ def make_step(self, command):
...
```
aneeshusa: nit: use elif instead of if

buildbot/master/files/config/factories.py
lucasloisp: I would have sworn I used `elif`. I'll change it

aneeshusa:
Commit message looks much better :) A few more nits:
- We actually have `mach package` for more platforms now, not just Android, so amend the beginning to something like `Even when this command fails, Buildbot still goes on to run the next step, ...`.
- The parenthetical with the issue reference is unneeded as GitHub already will make it easy to get to the issue from the commit, so remove it.
- grammar nit: `check for it on the` -> `check for it in the`

Also it looks like a merge commit sneaked in there, so go ahead and squash everything into one commit.

lucasloisp:
@aneeshusa Anything I should change or is the change ready to be approved?

aneeshusa:
@lucasloisp Sorry, I didn't get a notification that you had pushed new changes for some reason! This looks great, thanks for the PR :)

@bors-servo r+

PR #504, opened 2016-10-06
Remove old DynamicServoFactory which reads from saltfs

### LINKED ISSUE #316: Move buildbot logic to scripts in `/etc/ci`

aneeshusa:
Now that the new dynamic factory is working (which reads the steps
configuration from the main servo repo), remove the old one.
However, re-use the old name (DynamicServoFactory instead of
DynamicServoYAMLFactory) for concision and clarity.

Fixes #316. Blocked on servo/servo#13611.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/504)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -1,5 +1,4 @@
 import copy
-import os.path
 import re
 
 from buildbot.plugins import steps, util
@@ -15,12 +14,12 @@ SERVO_REPO = "https://github.com/servo/servo"
 
 
 class ServoFactory(util.BuildFactory):
-    """
+    """\
     Build factory which checks out the servo repo as the first build step.
     """
 
     def __init__(self, build_steps):
-        """
+        """\
         Takes a list of Buildbot steps.
         Prefer using DynamicServoFactory to using this class directly.
         """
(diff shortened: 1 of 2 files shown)


aneeshusa:
FYI I did this at 2am, so I'd like to take another look in the morning :)

aneeshusa:
servo/servo#13611 has merged, so this is no longer blocked.

aneeshusa:
`steps.yml` was changed, now blocked on a [follow-up PR](https://github.com/servo/saltfs/pull/508#issuecomment-253203920).

jdm:
The followup PR merged. It would be really great to get this reviewed and throw the switch before we change steps.yml again.

larsbergstrom:
This looks good. I think we can land this (even though it'll bitrot my MSVC PR). @aneeshusa Is this good to go?

My only worry is that now to add/change builders we have to do coordinated commits AND buildbot restarts across servo/servo and servo/saltfs and that's certainly going to cause breakage every time we have to do it.

aneeshusa:
Yes, this should be ready. This will optimize for the common case (changing steps for pre-existing builders), and I don't think having to make an extra PR to add/remove a builder (less common) is too onerous.

larsbergstrom:
This has been deployed.

larsbergstrom:
Oh, hrm, should we have removed steps.yml from saltfs, too?

aneeshusa:
I made #513 for that, good call.

larsbergstrom:
What's the easiest way to back this out via salt? The `salt-run` stuff that I did to stage the new mac?

aneeshusa:
If #514 doesn't work, easiest is probably to `git revert` the commits from before this PR up to current master, run a regular highstate, and prepare another PR (with a test builder on Windows!) at our leisure.

The other option is using the override folder and checking out an old git revision, notes at https://github.com/servo/saltfs/blob/b64ff04d219d1fe34c653ae5939f43fc4311d30b/salt/files/master/ADMIN_README.

PR #505, opened 2016-10-06
WIP: Automatically restart Buildbot master on configuration change

aneeshusa:
This should allow us to use Upstart to restart Buildbot cleanly, instead
of having to `su` to servo and restart Buildbot manually.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/505)

### DIFF
--- a/buildbot/master/files/buildbot-master-autostart.conf
+++ b/buildbot/master/files/buildbot-master-autostart.conf
@@ -1,0 +1,19 @@
+# Because the buildbot-master service is an instance job,
+# we must provide a reason to start an instance of it.
+# This Upstart task will automatically start one instance at bootup.
+# Additionally, if no instances are alive during a Salt deploy,
+# then this task will start a fresh instance.
+# Upstart will not create new instances otherwise;
+# additional instances can be created on-demand.
+# Each instance will automatically stop itself at shutdown via its `stop on`.
+
+start on (local-filesystems and net-device-up IFACE!=lo)
+
+task
+
+script
+  # Avoid restarting Buildbot if it's already running
+  if ! initctl list | grep "buildbot-master (" >/dev/null; then
+    start buildbot-master reason="$(date --utc --iso-8601=seconds)-autostart"
+  fi
+end script
--- a/buildbot/master/files/buildbot-master.conf
+++ b/buildbot/master/files/buildbot-master.conf
@@ -1,9 +1,20 @@
-exec /usr/local/bin/buildbot start --nodaemon {{ common.servo_home }}/buildbot/master
+# Use multiple instances to model multiple process:
+# - (1) currently running Buildbot master
+# - (n) queued restarts
+# Each 'restart' process will wait for the existing master to finish pending
+# builds, then take over as its own long-lived process.
+# Runnign them via Upstart provides process supervision and monitoring,
+# obviating the need to run restarts in tmux or screen (or disown them).
+#
+# Note that we don't actually use the $reason variable,
+# except to provide distinct instance IDs for Upstart.
+instance $reason
+
+exec /usr/local/bin/buildbot restart --clean --nodaemon $HOME/buildbot/master
 
 setuid servo
 setgid servo
 
-start on (local-filesystems and net-device-up IFACE!=lo)
 stop on runlevel [016]
 
 env HOME={{ common.servo_home }}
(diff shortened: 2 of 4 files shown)


aneeshusa:
I'm not 100% sure about how Upstart works, so we may want to try this manually on `servo-master1` before merging.

r? @larsbergstrom @edunham

metajack:
Upstart terminates the process and starts another on restart. The only difference between calling stop then start versus restart is that post-start and pre-stop and such logic is skipped for restart and the job control config is not reread.

There is also respawn, which restarts the job automatically when it exits. http://upstart.ubuntu.com/cookbook/#respawn

I don't think this patch is correct. What is the behavior you are observing and trying to fix?

aneeshusa:
Hmm, it does sound this patch won't work. Essentially, I want to teach Upstart how to restart Buildbot gracefully, so instead of running the steps at https://github.com/servo/servo/wiki/Buildbot-administration#dealing-with-troubles manually, we can always do `service buildbot-master restart`.

Since I don't see us upgrading Buildbot soon, most of the time we only need to reload the configuration, so another option would be cleanly reloading Buildbot. Maybe using `service reload buildbot-master` would work? Builbot [handles SIGHUP](https://docs.buildbot.net/0.8.12/manual/cmdline.html?highlight=sighup#sighup), but it's not clear to me if it handles it cleanly (without interrupting ongoing builds).

metajack:
There are other job types you could set up for one off things like this. What does --clean do anyway?

aneeshusa:
Buildbot's `--clean` parameter makes it shut down (and then possibly restart) cleanly, `allowing all running builds to complete first`.  This was [introduced in Buildbot 0.8.8](https://github.com/buildbot/buildbot/blob/master/master/docs/relnotes/0.8.8.rst#features).

The main point is to 1) avoid dropping builds 2) require as little admin effort as possible to update the Buildbot config (if you force-kill it it has to be cleaned up by hand to restart).

aneeshusa:
I looked into it more deeply and it seems Buildbot's reloading functionality has a lot of gotchas and edges cases it doesn't handle well. I also spent some more time with the Upstart Cookbook and I think I found another approach that may work - @metajack, does this look like a reasonable idea? (Still WIP.)

metajack:
Looks good, though I'm confused where the value of `$reason` comes from.

aneeshusa:
The `$reason` variable is just to give Upstart unique ids for each instance; the `cmd.run` state will automatically set it to `"salt-$(date)"`, while it will need to be specified manually for manual starts.

Earlier I was looking into making it work without specifying a reason, but it looks like that causes other problems, so I think I need to add another Upstart job to start an instance at boot-up.

metajack:
Oh, I see it now. r=me if you think this is ready to go.

aneeshusa:
First commit now fixes #552, and I included a bunch more changes too beyond the original PR. This should really make Buildbot much nicer and automated!

r? @larsbergstrom or @edunham on the Salt/Buildbot changes
r? @metajack on the Upstart changes

Also, let me know what docs I should add with this PR! Most things should be handled automatically by Salt now.

I highly recommend trying this out in Vagrant with the `servo-master1` VM. Use
`initctl list | grep buildbot` to look at the relevant jobs, and look inside the `/var/log/upstart` directory for relevant logs, for both `buildbot-master` and `buildbot-master-autostart`.
A few scenarios to try:
- Fresh deploy (`vagrant destroy --force servo-master1 && vagrant up servo-master1`): DB should be created and Buildbot should be (re)started and stay alive
- Reprovision: (`vagrant provision servo-master1`): no changes, should see the same Buildbot instance still running
- VM Restart: (`vagrant reload servo-master1`): Should see an autostarted Buildbot instance
- Upgrade (change buildbot version from 0.8.12 to 0.8.14 in `buildbot/master/init.sls` and run `vagrant provision servo-master1`): DB upgrade, Buildbot restart
- Start from stopped: Force stop Buildbot (`vagrant ssh servo-master1`, then run `sudo buildbot stop --clean /home/servo/buildbot/master` via SSH, confirm Buildbot is stopped, then run `vagrant provision servo-master1`): Buildbot should be started via autostart

FYI, strange output is due to https://github.com/saltstack/salt/issues/35989 and can be fixed by a Salt update.

aneeshusa:
Also, there is some manual migration required: Make sure to delete the existing `.pyc` files, e.g. with `$ find /home/servo/buildbot/master -name '*.pyc' -delete`. When poking inside Vagrant you can also leave a `watch -d find /home/servo/buildbot/master "'*.pyc'"` running to ensure no new `*.pyc` files get created, or use `watch -d sh -c "'initctl list | grep buildbot'"` to watch the running buildbot related Upstart jobs.

PR #508, opened 2016-10-12
Add release stylo unit tests

Manishearth:
https://github.com/servo/servo/pull/13700 moved some important binding layout tests to test-stylo (they used to be run by the regen script).

Currently test-stylo is only run in debug mode. This runs it in release mode as well.

Since linux-rel-css already contains a `./mach build-geckolib --release`, and `test-stylo` just needs a rebuild of style and geckolib (which takes half a minute), this shouldn't cause much of an issue.

However, linux-rel-css is one of our slowest builders. Perhaps both `build-geckolib --release` and `test-stylo --release` should be put in linux-dev instead?

r? @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/508)

### DIFF
--- a/buildbot/master/files/config/steps.yml
+++ b/buildbot/master/files/config/steps.yml
@@ -65,6 +65,7 @@ linux-rel-css:
   - ./mach test-css --release --processes 16 --log-raw test-css.log --log-errorsummary css-errorsummary.log
   - ./mach build-cef --release
   - ./mach build-geckolib --release
+  - ./mach test-stylo --release
   - bash ./etc/ci/lockfile_changed.sh
   - bash ./etc/ci/manifest_changed.sh
 


larsbergstrom:
@bors-servo r+

Let's see how the timing works out in production and we can shuffle things around as-needed. I'd probably be OK with either moving the "stable" rust stuff on top of `linux-dev` or creating a `linux-rel-stable` builder and doing the work there, on the assumption that we will eventually build+test more of Servo's tree on stable.

In general I somewhat prefer more builders doing different things than glomming unrelated things into the same builder except in the case where they can share artifacts in the `CARGO_TARGET_DIR`.

Manishearth:
Note that release geckolib does not share artefacts with release servo. So there's no advantage to keeping it in a `-rel` builder aside from consistent naming. The disadvantage is that the crates (which may have different versions) will get downloaded again, as will rust stable (though that will only happen every now and then)

aneeshusa:
Please file a follow up PR to servo/servo to sync the steps.yml again in anticipation of #504. (Also see servo/servo#13611).

Issue #473, opened 2016-08-25
Get A+ on Mozilla Observatory for servo.org


aneeshusa:
> Observatory by Mozilla is a project designed to help developers, system administrators, and security professionals configure their sites safely and securely. 

Scan results: https://observatory.mozilla.org/analyze.html?host=servo.org

This should mostly entail adding headers via Ngnix.

aneeshusa:
Something that caught my eye is that we are failing the cookie test, even though we ourselves don't set any cookies. It turns out CloudFlare is adding a `__cfuid` cookie which 1) cannot be disabled, 2) does not have the `Secure` flag set, and 3) is sent even on `http` connections, not just `https` connections.

aneeshusa:
Oops, misclick.

aneeshusa:
Bonus points if you utilize the actual [`http observatory code`](https://github.com/mozilla/http-observatory) for the test instead of just checking headers.

Coder206:
@aneeshusa I would like to help out here.

jarondl:
I have started investigating this issue, and this is my current understanding:

1. [servo.org](servo.org) , [starters.servo.org](starters.servo.org) , and [download.servo.org](download.servo.org)  are all hosted on github pages (on separate repos) with cloudflare in-front.
 For them, the nginx configuration file is irrelevant. Maybe the headers can be added via cloudfront page rules, but i.d.k

2. [build.servo.org](build.servo.org) is using the `nginx` server defined using saltstack. It is not even https, so it cannot pass the observatory requirements. We should first convert it to https (getting a certificate etc.), and only then can we try to get high scores on the observatory.

Are there other subdomains I haven't found?

aneeshusa:
@jarondl The only other subdomain I know of is `dl.servo.org` (which currently gets TLS errors because we don't have SANs on our `download.servo.org` cert); maybe @edunham or @larsbergstrom knows about more about that or adding headers in cloudfront/cloudflare.

(I think download.servo.org is on S3 with some redirection magic, not GH pages.)

We are planning to add https to build.servo.org; see #423 for the tracking issue. Even while that is in progress, we can still work on adding the appropriate headers to Nginx; it's not necessary to go from ~0 to 100% in one step. I can assign you if you would like to work on this.

jarondl:
There is also `blog.servo.org`, which is also gh-pages with cloudflare.
Yes, I would like to work on this, please assign me.

aneeshusa:
[...]

PR #511, opened 2016-10-17
Add security related headers to build.servo.org

jarondl:
Add some headers to `build.servo.org` to improve our Mozilla Observatory
score. The helps the cause of #473, by raising our score from F (0) to a
whopping B- (65). The testing is still basic, the docker-observatory
based testing will go in a different PR.

To get to an A+ we must first implement https.

This is the local observatory report:

```
Score Rule                       Description
      -20 redirection                Does not redirect to an https site.
      -20 strict-transport-security  HTTP Strict Transport Security (HSTS)
        header cannot be set for sites not available over https.
       -5 contribute                 Contribute.json file missing from root
        of website.
        5 content-security-policy    Content Security Policy (CSP)
        implemented without 'unsafe-inline' or 'unsafe-eval'.
        5 x-frame-options            X-Frame-Options (XFO) implemented via
        the CSP frame-ancestors directive.

    Score: 65
    Grade: B-

```

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/511)

### DIFF
--- a/nginx/https_headers.conf
+++ b/nginx/https_headers.conf
@@ -1,0 +1,15 @@
+# The following headers were added to pass the mozilla http observatory
+# test suit.
+# https://wiki.mozilla.org/Security/Guidelines/Web_Security
+
+# All of them are very simple in this case, because both homu and buildbot
+# do not use external resources.
+
+# CSP is the most important one
+add_header Content-Security-Policy "default-src 'self'; frame-ancestors 'none'";
+# Block site from being framed
+add_header X-Frame-Options DENY;
+# Block pages from loading when they detect reflected XSS attacks
+add_header X-XSS-Protection "1; mode=block";
+# Prevent browsers from incorrectly detecting non-scripts as scripts
+add_header X-Content-Type-Options nosniff;
--- a/nginx/init.sls
+++ b/nginx/init.sls
@@ -14,6 +14,15 @@ nginx:
     - watch_in:
       - service: nginx
 
+/etc/nginx/conf.d/https_headers.conf:
+  file.managed:
+    - source: salt://nginx/https_headers.conf
+    - user: root
+    - group: root
+    - mode: 644
+    - watch_in:
+      - service: nginx
+
 /etc/nginx/sites-enabled/default:
   file.symlink:
     - target: /etc/nginx/sites-available/default
--- a/tests/sls/nginx/headers.py
+++ b/tests/sls/nginx/headers.py
@@ -1,0 +1,30 @@
+"""
+Test nginx's security related headers
+
+Test the headers very naively - their existence and value. Only check the
+top-most '/' domain.
+"""
+import urllib.request
+
+from tests.util import Failure, Success
+
+
+def run():
+    expected_headers = [
+        ('Content-Security-Policy',
+            "default-src 'self'; frame-ancestors 'none'"),
+        ('X-Frame-Options', 'DENY'),
+        ('X-XSS-Protection', '1; mode=block'),
+        ('X-Content-Type-Options', 'nosniff')]
+    with urllib.request.urlopen('http://localhost/') as local_open:
+        actual_headers = local_open.getheaders()
+        failures = []
+        for header in expected_headers:
+            if header not in actual_headers:
+                failures.append('Missing or changed header - {}:{}'.format(*header))
+
+    if len(failures) > 0:
+        return Failure('nginx is serving wrong securirty headers',
+                       '\n'.join(failures))
+    else:
+        return Success('nginx is serving the correct security headers')


aneeshusa:
@jarondl Can you point me to some documentation about the `contribute.json` file? I haven't heard of this and didn't see it on the observatory results.

jarondl:
Yes, http://contribute.json/  is quite obscure. I have found about it only when I've browsed through the observatory code to resolve some bug.
According to the observatory, only Mozilla related websites must have it, see for example: 

    https://observatory.mozilla.org/analyze.html?host=mozilla.org

The list of mozilla related sites is hard coded here: https://github.com/mozilla/http-observatory/blob/7dd04ff5b51e39698a8b683df33ef430b4ed4213/httpobs/conf/httpobs.conf#L33
and servo is not there, but in all fairness it should be. We should submit a PR to add it, and degrade us from 70 to 65.

jarondl:
I believe the tests have failed because buildbot is down in travis. We can circumvent this by letting nginx serve 200 on some page (say http://buildbot.servo.org/generate200/ ), but I don't really like this solution. We should have something more similar to real life. Also, we plan to use the observatory, and it also check the contents of the page.
@aneeshusa, could you explain why buildbot and homu don't work during testing? Is there anything we can do about that?

aneeshusa:
Buildbot immediately stops on Travis after being started (due to our fake test credentials not validating IIRC), which I don't have an easy fix for. In the meantime, I believe this is why I added the `make_cors_request` function in #375, since Nginx should still add the headers even though the status code indicates error (since Buildbot is down). Try adding/using that in this test.

nginx/https_headers.conf
```diff
@@ -0,0 +1,15 @@
+# The following headers were added to pass the mozilla http observatory
```
aneeshusa: nit: mozilla -> Mozilla

nginx/https_headers.conf
```diff
@@ -0,0 +1,15 @@
...
+# test suit.
```
aneeshusa: spelling: suit -> suite

nginx/https_headers.conf
```diff
@@ -0,0 +1,15 @@
...
+# https://wiki.mozilla.org/Security/Guidelines/Web_Security
+
+# All of them are very simple in this case, because both homu and buildbot
+# do not use external resources.
+
+# CSP is the most important one
```
aneeshusa: The options we're setting are fairly basic and IMO the headers are readable enough as is that we don't need an expository comment for each one.

nginx/init.sls
```diff
@@ -14,6 +14,15 @@ nginx:
     - watch_in:
       - service: nginx
 
+/etc/nginx/conf.d/https_headers.conf:
+  file.managed:
+    - source: salt://nginx/https_headers.conf
```
aneeshusa: Use `salt://{{ tpldir }}/files/https_headers.conf`, and make a `files` subdir to put the config file in.

I'm not sure how well you know Salt, but we'll also want the `/etc/nginx/conf.d` state from #375 and some requisites to order things correctly/restart nginx automatically on config change. I can also take care of this if you'd like.

tests/sls/nginx/headers.py
```diff
@@ -0,0 +1,30 @@
...
+            "default-src 'self'; frame-ancestors 'none'"),
+        ('X-Frame-Options', 'DENY'),
+        ('X-XSS-Protection', '1; mode=block'),
+        ('X-Content-Type-Options', 'nosniff')]
+    with urllib.request.urlopen('http://localhost/') as local_open:
+        actual_headers = local_open.getheaders()
+        failures = []
+        for header in expected_headers:
+            if header not in actual_headers:
+                failures.append('Missing or changed header - {}:{}'.format(*header))
```
aneeshusa: This line is failing tidy because it is too long.

tests/sls/nginx/headers.py
```diff
@@ -0,0 +1,30 @@
...
+from tests.util import Failure, Success
+
+
+def run():
+    expected_headers = [
+        ('Content-Security-Policy',
+            "default-src 'self'; frame-ancestors 'none'"),
+        ('X-Frame-Options', 'DENY'),
+        ('X-XSS-Protection', '1; mode=block'),
+        ('X-Content-Type-Options', 'nosniff')]
```
aneeshusa: style nit: leave a trailing comma after this last tuple, and put the closing `]` by itself on the next line.

tests/sls/nginx/headers.py
```diff
@@ -0,0 +1,30 @@
...
+    with urllib.request.urlopen('http://localhost/') as local_open:
+        actual_headers = local_open.getheaders()
+        failures = []
+        for header in expected_headers:
+            if header not in actual_headers:
+                failures.append('Missing or changed header - {}:{}'.format(*header))
+
+    if len(failures) > 0:
+        return Failure('nginx is serving wrong securirty headers',
```
aneeshusa: spelling: securirty -> security

aneeshusa:
Also, please open a separate issue about `contribute.json` to keep this PR about headers.

jarondl:
Hi @aneeshusa , thanks for reviewing, I will work on it.

aneeshusa:
I'm going to close this as it's been inactive for a while; anyone is welcome to pick this work up again.

PR #514, opened 2016-10-19
Fix steps reading on Windows

aneeshusa:
`cat` is not available, so use `type` which does the same thing.

r? @larsbergstrom 
I spent about ~10 seconds searching for this, but it looks like it'll work...

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/514)


larsbergstrom:
Cool, I'll roll it out and push to the servers and retry something after this. No queue to speak of right now, so not creating a huge backlog yet :-)

larsbergstrom:
This passed Travis. Merging so I can roll it out.

PR #517, opened 2016-10-20
Fix buildbot_steps.yml reading on Windows

UK992:
This should fix reading `buildbot_steps.yml` on Windows.

cc @larsbergstrom @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/517)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -187,8 +187,11 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
         self.is_windows = re.match('windows.*', self.builder_name) is not None
         try:
             show_cmd = "cat" if not self.is_windows else "type"
+            native_yaml_path = self.yaml_path
+            if self.is_windows:
+                native_yaml_path = native_yaml_path.replace('/', '\\')
             cmd = yield self.makeRemoteShellCommand(
-                command=[show_cmd, "{}".format(self.yaml_path)],
+                command=[show_cmd, native_yaml_path],
                 collectStdout=True
             )
             yield self.runCommand(cmd)


aneeshusa:
I've put in a few partial fixes so far, let me do a try run to see if we need this as well.

buildbot/master/files/config/factories.py
```diff
@@ -187,8 +187,9 @@ def run(self):
         self.is_windows = re.match('windows.*', self.builder_name) is not None
         try:
             show_cmd = "cat" if not self.is_windows else "type"
+            yaml_file_path = self.yaml_path.replace('/', '\\') if self.is_windows else self.yaml_path
```
aneeshusa: This line is too long; consider
```
native_yaml_path = self.yaml_path
if self.is_windows:
    native_yaml_path = native_yaml_path.replace('/', '\\')
```

UK992: done

buildbot/master/files/config/factories.py
```diff
@@ -187,8 +187,11 @@ def run(self):
         self.is_windows = re.match('windows.*', self.builder_name) is not None
         try:
             show_cmd = "cat" if not self.is_windows else "type"
+            native_yaml_path = self.yaml_path
+            if self.is_windows:
+                native_yaml_path = native_yaml_path.replace('/', '\\')
             cmd = yield self.makeRemoteShellCommand(
-                command=[show_cmd, "{}".format(self.yaml_path)],
+                command=[show_cmd, "{}".format(native_yaml_path)],
```
aneeshusa: While you're at it, can you get rid of the useless `format` call and just use `native_yaml_path` directly? I missed that on my last PR which I did in a rush.

UK992: done

PR #529, opened 2016-10-29
Update to a newer version of the Android NDK

larsbergstrom:
r? @aneeshusa 

Comments from @mmatyas welcome :-)

There are two parts to this change.
1) Update the NDK. Google has both changed the DL URL format and removed the excutable bit in favor of a normal zipfile.
2) We no longer use the android standalone toolchain stuff, as that is not compatible with the LLVM libc++ that SpiderMonkey now requires us to use on Android.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/529)

### DIFF
--- a/servo-build-dependencies/android.sls
+++ b/servo-build-dependencies/android.sls
@@ -28,6 +28,7 @@ android-dependencies:
       - lib32z1
       - libstdc++6
       - libgl1-mesa-dev
+      - unzip
     - refresh: True
   pip.installed:
     - pkgs:
@@ -79,25 +80,18 @@ android-sdk-current:
 
 
 android-ndk:
-  file.managed:
-    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin
-    - source: https://dl.google.com/android/ndk/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin
+  archive.extracted:
+    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}
+    - source: https://dl.google.com/android/repository/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
     - source_hash: sha512={{ android.ndk.sha512 }}
+    - archive_format: zip
+      # Workaround for https://github.com/saltstack/salt/pull/36552
+    - archive_user: servo
     - user: servo
     - group: servo
-    - mode: 744
-    - dir_mode: 755
-    - makedirs: True
+    - if_missing: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}
     - require:
       - user: servo
-  cmd.run:
-      # Need to filter log output to avoid hitting log limits on Travis CI
-    - name: '{{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin | grep -v Extracting'
-    - runas: servo
-    - cwd: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}
-    - creates: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}
-    - require:
-      - file: android-ndk
 
 android-ndk-current:
   file.symlink:
@@ -106,5 +100,4 @@ android-ndk-current:
     - user: servo
     - group: servo
     - require:
-      - cmd: android-ndk
-
+      - android-ndk
--- a/servo-build-dependencies/android.sls
+++ b/servo-build-dependencies/android.sls
@@ -99,15 +99,6 @@ android-ndk:
     - require:
       - file: android-ndk
 
-android-toolchain:
-  cmd.run:
-    - name: bash {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}/build/tools/make-standalone-toolchain.sh --platform=android-{{ android.platform }} --toolchain=arm-linux-androideabi-4.8 --install-dir='{{ common.servo_home }}/android/toolchain/{{ android.ndk.version }}/android-toolchain' --ndk-dir='{{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}'
-    - runas: servo
-    - creates: {{ common.servo_home }}/android/toolchain/{{ android.ndk.version }}/android-toolchain
-    - require:
-      - cmd: android-ndk
-
-# Toolchain depends on NDK so update the symlinks together
 android-ndk-current:
   file.symlink:
     - name: {{ common.servo_home }}/android/ndk/current
@@ -116,14 +107,4 @@ android-ndk-current:
     - group: servo
     - require:
       - cmd: android-ndk
-      - cmd: android-toolchain
 
-android-toolchain-current:
-  file.symlink:
-    - name: {{ common.servo_home }}/android/toolchain/current
-    - target: {{ common.servo_home }}/android/toolchain/{{ android.ndk.version }}/android-toolchain
-    - user: servo
-    - group: servo
-    - require:
-      - cmd: android-ndk
-      - cmd: android-toolchain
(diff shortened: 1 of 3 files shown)


aneeshusa:
It sounds like those two changes are unrelated, so please split this into two commits, and add those justifications to the commit messages please :)

servo-build-dependencies/android.sls
```diff
@@ -92,22 +93,13 @@ android-ndk:
       - user: servo
   cmd.run:
       # Need to filter log output to avoid hitting log limits on Travis CI
-    - name: '{{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin | grep -v Extracting'
+    - name: 'unzip {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip'
```
aneeshusa: `archive.extracted` supports unzipping, so please use that instead of manually running unzip.

servo-build-dependencies/android.sls
```diff
@@ -28,6 +28,7 @@ android-dependencies:
       - lib32z1
       - libstdc++6
       - libgl1-mesa-dev
+      - unzip
```
aneeshusa: It looks like Salt sometimes uses the binary and sometimes Python's zipfile module, so let's keep this even with `archive.extracted`.

servo-build-dependencies/map.jinja
```diff
@@ -7,8 +7,8 @@
       'sha512': '96fb71d78a8c2833afeba6df617edcd6cc4e37ecd0c3bec38c39e78204ed3c2bd54b138a56086bf5ccd95e372e3c36e72c1550c13df8232ec19537da93049284'
     },
     'ndk': {
-      'version': 'r10e',
-      'sha512': '8948c7bd1621e32dce554d5cd1268ffda2e9c5e6b2dda5b8cf0266ea60aa2dd6fddf8d290683fc1ef0b69d66c898226c7f52cc567dbb14352b4191ac19dfb371'
+      'version': 'r12b',
```
mmatyas: Personally I'm using r12, does it work with r12b too?

mmatyas:
Another problem that might occur is that everything works fine if you only have Java 7 or older installed, but if you also have Java 8 or 9, then certain libs want to build using the latest version, causing version difference and parsing errors during `mach package`. I'm not sure what kind of system the bots run on, but in case you get such errors, thats's the cause.

larsbergstrom:
Hrm, we just use whatever the `default-jdk` is on Ubuntu (https://github.com/servo/saltfs/blob/master/servo-build-dependencies/android.sls#L23 ). Perhaps I should finally update us to use the more modern Android build system tools once this is fully landed :-)

mmatyas:
`default-jdk` points to `openjdk-8-jre` on my Ubuntu 16.04, but on older systems, it was Java 7, and I had to manually `add-apt-repository ppa:openjdk-r/ppa` and install `openjdk-7-jdk` to make it work. Even after setting it as the default Java, the build system (probably one of the dependencies) still calls J8 somehow. But as long as you have Java 7 only, it should work fine.

servo-build-dependencies/android.sls
aneeshusa: To be more clear, please use the Salt `archive.extracted` state (with `archive_type: zip`) instead of manually downloading and extracting for `android-ndk`. (`android-sdk` has an example of this in the same file, and the docs are [here](https://docs.saltstack.com/en/2016.3/ref/states/all/salt.states.archive.html#salt.states.archive.extracted).) Also keep the `unzip` package dependency, as Salt sometimes uses it internally.

larsbergstrom:
@aneeshusa OK, does that look better?

larsbergstrom:
OK, passing CI now! Squashed. r?  :-)

servo-build-dependencies/android.sls
```diff
@@ -79,51 +80,24 @@ android-sdk-current:
...
-    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin
-    - source: https://dl.google.com/android/ndk/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin
+  archive.extracted:
+    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
+    - source: https://dl.google.com/android/repository/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
     - source_hash: sha512={{ android.ndk.sha512 }}
+    - archive_format: zip
     - user: servo
     - group: servo
     - mode: 744
```
aneeshusa: Remove `mode` and `dir_mode`, which aren't accepted by `archive.extracted.`

servo-build-dependencies/android.sls
```diff
@@ -79,51 +80,24 @@ android-sdk-current:
...
 android-ndk:
-  file.managed:
-    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin
-    - source: https://dl.google.com/android/ndk/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin
+  archive.extracted:
+    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
+    - source: https://dl.google.com/android/repository/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
     - source_hash: sha512={{ android.ndk.sha512 }}
+    - archive_format: zip
     - user: servo
```
aneeshusa: Also pass `archive_user` (see the `android-sdk` state for why, feel free to copy the justification comment down here as well.)

servo-build-dependencies/android.sls
```diff
@@ -79,51 +80,24 @@ android-sdk-current:
...
     - dir_mode: 755
     - makedirs: True
     - require:
       - user: servo
-  cmd.run:
-      # Need to filter log output to avoid hitting log limits on Travis CI
-    - name: '{{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.bin | grep -v Extracting'
-    - runas: servo
-    - cwd: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}
-    - creates: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}
```
aneeshusa: I haven't checked in Vagrant, but I'm assuming the `archive.extracted` state will also create this in the FS; convert this to a `if_missing` arg for `archive.extracted`.

servo-build-dependencies/android.sls
```diff
@@ -79,51 +80,24 @@ android-sdk-current:
...
+  archive.extracted:
+    - name: {{ common.servo_home }}/android/ndk/{{ android.ndk.version }}/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
+    - source: https://dl.google.com/android/repository/android-ndk-{{ android.ndk.version }}-linux-x86_64.zip
     - source_hash: sha512={{ android.ndk.sha512 }}
+    - archive_format: zip
     - user: servo
     - group: servo
     - mode: 744
     - dir_mode: 755
     - makedirs: True
```
aneeshusa: Remove this as well, as `archive.extracted` always makes the necessary parent directories and so doesn't accept `makedirs`.

larsbergstrom:
Also updated commit messages :-)

aneeshusa:
OK, tested this in Vagrant and it looks OK.

r=me after a few more commit message nits:
- Don't have the first commit message all on one line; use "Remove usage of ... for building Android" as the first line (title), then make the rest into the body. 
- Leave off the periods at the end of the commit title (first line)

Don't forget to clean out the old toolchain/ dirs on the builders, as well as the old `r10e` NDK.

larsbergstrom:
This has been deployed to `servo-linux-cross*`

larsbergstrom:
Hrm, it appears that salt did not preserve the execute bit on some of these files?
```
servo@ip-172-31-34-111:~/android/ndk$ ls -al ./r12b/android-ndk-r12b/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-gcc
-rw-r--r-- 1 servo servo 784904 Nov  9 20:51 ./r12b/android-ndk-r12b/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/bin/arm-linux-androideabi-gcc
```

aneeshusa:
The permissions not being extracted properly should be fixed by #639. (I didn't forget about this! :smile:)

Issue #530, opened 2016-10-29
Buildbot sometimes checks out the wrong revision for try runs


jdm:
https://github.com/servo/servo/pull/13957#issuecomment-257116847 links to a log with a compilation error that is actually from https://github.com/servo/servo/pull/13956#issuecomment-257115718 . This really confused me!

aneeshusa:
I'm unable to reproduce - when I click through to the `stdio` and `warnings` logs for those comments, they have different contents - different lengths, file paths, headers, etc. (although they both end in the same error). Can you provide direct links to the logs you are seeing or maybe upload some screenshots?
There may be a caching problem somewhere, although I'm doubtful as one is for a `windows-dev` build and the other is for a `linux-dev` build.

jdm:
However, you're absolutely right that http://build.servo.org/builders/windows-dev/builds/1770/steps/compile/logs/stdio and http://build.servo.org/builders/windows-dev/builds/1772/steps/compile/logs/stdio are different logs with different contents. This may be a problem of the code on the try branch not being updated correctly?

jdm:
http://build.servo.org/builders/windows-dev/builds/1770/steps/git/logs/stdio and http://build.servo.org/builders/windows-dev/builds/1772/steps/git/logs/stdio show the same hash being used when updating the branch, which would explain why this occurred.

jdm:
http://build.servo.org/builders/windows-dev/builds/1770 and http://build.servo.org/builders/windows-dev/builds/1772 show that different revisions should have been used.

aneeshusa:
OK, that makes more sense and matches what I'm seeing, thanks for looking again.
http://build.servo.org/builders/windows-dev/builds/1772 has different values for `revision` (the revision triggering the build request, in this case `a06cb1ca`) and `got_revision` (the revision that was actually used during the build, in this case `59e651b7`). Checking the git log, it seems like Buildbot fetched the `try` branch, tried to check out the commit from `revision`, did both of those steps again, then gave up and just used the current value of `try` as the `got_revision` branch.

I can see two potential problems here:
- Something force-pushing to `try` such that `a06cb1ca` was not available - this may be due to multiple restarts of Buildbot soon after each other when fixing #524 yesterday.
- Buildbot accepting whatever commit it found on the try branch (aka `59e651b7`) instead of erroring out.

I'll look into some potential solutions.

aneeshusa:
[...]

PR #531, opened 2016-10-30
Add step to ensure Buildbot checks out right commit

aneeshusa:
We are using `retryFetch=True` to have Buildbot retry git fetches if
they error out, as we often encounter intermittent errors (e.g. network
problems).

However, Buildbot will allow not only the first fetch + checkout to
fail, but also the second (final) fetch + checkout to fail as well
(which I consider a bug in Buildbot), making it possible for Buildbot
to run a build on the wrong revision, causing confusion.

Retrying failed fetches is too useful to disable.
Hence, to work around this, add a custom step to all builds that checks
if the commit we requested (the `revision` property) matches the
actual commit Buildbot checks out, and fail the build if not the case.

---

This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/531)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -14,6 +14,36 @@ import environments as envs
 SERVO_REPO = "https://github.com/servo/servo"
 
 
+class CheckRevisionStep(buildstep.BuildStep):
+    """\
+    Step which checks to ensure the revision that triggered the build
+    is the same revision that we actually checked out,
+    and fails the build if this is not the case.
+    """
+
+    haltOnFailure = True
+    flunkOnFailure = True
+
+    def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


aneeshusa:
cc @jdm 
Will cause Buildbot to complain loudly if #530 happens again.

r? @larsbergstrom

larsbergstrom:
This is awesome! Thanks for doing this. I'm willing to give this a try - should really help! Just one issue:
```
  ./buildbot/master/files/config/factories.py:37:16: E711 comparison to None should be 'if cond is not None:'
```

aneeshusa:
Fixed the nit. I haven't seen this happen too many times in the past (I'm hoping it didn't just silently happen!) so I'm not sure how helpful this is, but it does give me more peace of mind.

PR #536, opened 2016-11-13
homu: Add ability to add per-repo reviewers.

emilio:
I've been thinking about adding @fitzgen to the reviewers in servo/rust-bindgen.

I think he can perfectly get access to full reviewer privileges, but there are
also a few contributors that I've been considering giving access to, and those
don't need full Servo reviewer access.

This adds a way to add per-repo reviewers to the travis repos, something I guess
it's also useful for rust-url, where there are more people with write
permissions.

The way this is organized makes easier to add "out-of-the-org" repos too.

r? @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/536)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -15,55 +15,154 @@ port = 54856
 secret = "{{ pillar["homu"]["web-secret"] }}"
 
 {% set travis_repos = [
-    ('servo', 'angle'),
-    ('servo', 'app_units'),
-    ('servo', 'cgl-rs'),
-    ('servo', 'cocoa-rs'),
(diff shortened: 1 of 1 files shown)


larsbergstrom:
@bors-servo r+

Thanks - this looks great to me!

aneeshusa:
This is fine, but I would have preferred something like this:

```
travis_repos = {
...
"osmesa-src": {}.
"rust-azure": {},
"rust-bindgen": {
    "extra_reviewers": [ "fitzgen" ]
},
"rust-cssparser": {},
...
}
```

Jinja is pretty powerful :)

PR #322, opened 2016-04-16
use bash to run the mach stuff

larsbergstrom:
r? @aneeshusa @edunham 

Ugh, I forgot that we need to execute bash.exe in order to run the mach command successfully. See:
https://github.com/servo/servo/blob/master/appveyor.yml#L45

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/322)

PR #537, opened 2016-11-15
Export Windows Python on Windows builders

UK992:
This _should_ fix problem on Windows builders in https://github.com/servo/servo/pull/14039
r? @aneeshusa 
cc @metajack

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/537)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -176,11 +176,9 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
 
         command = command.split(' ')
 
-        # Add `bash -l -c` and export Windows Python before every command
-        # on Windows builders
-        bash_args = "export PATH=/c/python27:/c/python27/scripts:$PATH; "
-        command_win = ["bash", "-l", "-c", bash_args + " ".join(command)]
-        step_kwargs['command'] = command_win if self.is_windows else command
+        # Add `bash` before every command on Windows builders
+        bash_args = ["bash"] if self.is_windows else []
+        step_kwargs['command'] = bash_args + command
         if self.is_windows:
             step_env += envs.Environment({
                 # Set home directory, to avoid adding `cd` command every time
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -176,9 +176,11 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
 
         command = command.split(' ')
 
-        # Add `bash -l` before every command on Windows builders
-        bash_args = ["bash", "-l"] if self.is_windows else []
-        step_kwargs['command'] = bash_args + command
+        # Add `bash -l -c` and export Windows Python before every command
+        # on Windows builders
+        bash_args = "export PATH=/c/python27:/c/python27/scripts:$PATH; "
+        command_win = ["bash", "-l", "-c", bash_args + " ".join(command)]
+        step_kwargs['command'] = command_win if self.is_windows else command
         if self.is_windows:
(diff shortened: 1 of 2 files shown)


buildbot/master/files/config/factories.py
```diff
@@ -176,9 +176,12 @@ def make_step(self, command):
...
 
-        # Add `bash -l` before every command on Windows builders
-        bash_args = ["bash", "-l"] if self.is_windows else []
-        step_kwargs['command'] = bash_args + command
+        # Add `bash -l -c` and export Windows Python before every command
+        # on Windows builders
+        if self.is_windows:
+            bash_args = "export PATH=/c/python27:/c/python27/scripts:$PATH; "
+            command_win = ["bash", "-l", "-c", bash_args + " ".join(command)]
+        step_kwargs['command'] = command_win if self.is_windows else command
```
metajack: Do you not need `command_win = ""` above the if? This looks like odd scoping.

metajack:
r=me with comment answered or addressed

buildbot/master/files/config/factories.py
UK992: i removed above if.

buildbot/master/files/config/factories.py
```diff
@@ -176,9 +176,11 @@ def make_step(self, command):
 
         command = command.split(' ')
 
-        # Add `bash -l` before every command on Windows builders
-        bash_args = ["bash", "-l"] if self.is_windows else []
-        step_kwargs['command'] = bash_args + command
+        # Add `bash -l -c` and export Windows Python before every command
+        # on Windows builders
+        bash_args = "export PATH=/c/python27:/c/python27/scripts:$PATH; "
```
aneeshusa: We have an `Environment` class which makes it easy to add things to environment. Is there some reason you can't update `step_env` instead (i.e. does the environment not get inherited through bash)?

UK992: yes, `PATH` environment variable does not get inherited through bash.

aneeshusa: Alternately, @larsbergstrom is there a `.profile` or `.bash_profile` that would reset the `$PATH` when we run bash?

metajack:
I added the path stuff to `.bashrc` on the two windows builders before my last retry. It didn't seem to help, but I haven't had to chance to investigate.

aneeshusa:
@metajack `.bashrc` isn't loaded for login shells: https://shreevatsa.wordpress.com/2008/03/30/zshbash-startup-files-loading-order-bashrc-zshrc-etc/

Try `~/.profile` instead.

metajack:
`~/.profile` doesn't work but `~/.bash_profile` seemed to. Testing that now.

UK992:
With removing `-l` argument, bash inherit `PATH` environment variable, so this should work without any additional hacks.

aneeshusa:
This approach looks fine to me. @larsbergstrom, do you remember why you used a login shell initially in #322?
Also, we should check for `.bashrc` files on the builders, as that will likely override any settings here if we switch to non-login shells.

metajack:
Note that I have tried to solve the problem by modifying `/etc/profile`. Buildbot doesn't run as admin, and it's home dir is the build dir, so there's no way to locally override PATH there.

aneeshusa:
@metajack Considering we haven't Salted the Windows builders, I like the approach in this PR since the PATH change is stored in version control, vs. editing arbitrary files on the builders directly. If the `/etc/profile` modification fixes the build then that works too, but it would be nice to write down somewhere (wiki?).

aneeshusa:
Closing since servo/servo#14039 was merged and Windows builds are working.

PR #541, opened 2016-11-29
update_brew doesn't exist anymore

paulrouget:
follow up https://github.com/servo/servo/pull/14403

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/541)


aneeshusa:
We'll need to pass the token to the `upload_nightly.sh` script instead. Best way is to move the token into the `upload_nightly` `Environment` in `environments.py`.

Normally we would copy the token to both places, then remove the old copy after removing `update_brew.sh`, but since it's currently busted anyways I think we can skip that.

aneeshusa:
Didn't see you had updated this, LGTM. @bors-servo r+

PR #542, opened 2016-11-29
pin python 2.7.9 to make SNI work

edunham:
see http://stackoverflow.com/questions/18578439/using-requests-with-tls-doesnt-give-sni-support/18579484#18579484 for more info but really it's the Right Thing to make sure all the builders are on the same python version anyways

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/542)

### DIFF
--- a/buildbot/slave/files/config/buildbot.tac
+++ b/buildbot/slave/files/config/buildbot.tac
@@ -1,4 +1,3 @@
-
 import os
 
 from buildslave.bot import BuildSlave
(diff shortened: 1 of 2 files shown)


buildbot/slave/files/config/buildbot.tac
```diff
@@ -1,4 +1,3 @@
-
```
aneeshusa: I believe the `.tac` files are generated by Buildbot, so I would prefer to leave them untouched.

PR #546, opened 2016-12-07
Use non-broken hash for Salt fileserver cache checking

aneeshusa:
The default is md5, so upgrade to a better hash.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/546)

### DIFF
--- a/salt/map.jinja
+++ b/salt/map.jinja
@@ -11,6 +11,7 @@
             },
             'gitfs_remotes': [ 'https://github.com/servo/saltfs.git' ],
             'gitfs_env_whitelist': [ 'base' ],
+            'hash_type': 'sha384',
             'pillar_roots': {
               'base': [ '/srv/pillar' ]
             }


aneeshusa:
FYI, this will require a manual `salt-master` restart after deployment; make sure to clean out the Salt master cache (`/var/cache/salt/master` I believe) when restarting as well to get rid of old md5 hashes.

larsbergstrom:
I think you might need to rebase to pick up the newer intermittent tracker thing? The Travis stuff failed. After that, r=me :-)

aneeshusa:
It shouldn't matter since homu will test after merging into master, but rebased anyways. @bors-servo r=larsbergstrom

larsbergstrom:
This has been deployed, the cache obliterated, and logs checked to ensure that after a restart teh salt-master service still appeared to be running.

Issue #437, opened 2016-07-11
Create + document Salt highstate log creation/retention policy


aneeshusa:
Follow up from servo/servo#12314.
cc @larsbergstrom @edunham and anyone else who runs highstates

There are multiple people who run highstates, and our saltfs testing story is still not that great, so it would be nice to have logs for highstates in case something goes wrong. I'd be +1 for logging all highstate runs + retention for at least 2 days (should be more than enough for any issues to pop up).

I like to test and run highstates as so (second time without `test=True`):
```
salt --force-color '*' state.highstate test=True | tee apply_prnumberhere_output | less -R
```
We may also want to capture `stderr`. If we do this, we should choose a directory to store these logs in.

larsbergstrom:
Is there any way to set up the salt master / minion services to log verbosely automatically? I'd be inclined to roll that out and just rely on normal logrotate, etc. instead of having us try to remember to capture all the output to files, if it's possible.

PR #547, opened 2016-12-07
Enable unlimited Salt job logging

aneeshusa:
Add some docs in tree as well about this feature,
to avoid forgetting to update the wiki when the PR lands.

This is a potential (partial?) fix for #437, with two caveats:
- I'm not sure of a way to "tag" a highstate deployment with the revision/PR being deployed; we'll still need to keep track of that manually
- The `keep_jobs` config option is a number of hours (the default is 24), so we can set it a concrete number instead of 0 for unlimited if you'd like.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/547)

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,8 @@ infrastructure. See [the tutorials](https://docs.saltstack.com/en/2016.3/topics/
 
 ## Contributing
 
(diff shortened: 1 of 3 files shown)


larsbergstrom:
Same as other, r=me after rebase so that it has the non-broken intermittent tracker revision :-)

Thanks!

Issue #545, opened 2016-11-30
We're gating on Android


Ms2ger:
Somehow #539 got deployed before its time.

On `servo-master1`, I cloned 70fe5103dd9846cf5859dc48e8add5fcb73c9c2f into `/tmp/salt-testing-root/saltfs` and ran `salt 'servo-master1' state.highstate 2>&1`. Waiting to see if that helped.

Ms2ger:
CC @aneeshusa @larsbergstrom @metajack @Manishearth

Ms2ger:
CC @edunham

Ms2ger:
That seems to have helped as a temporary band-aid.

Some notes for the post-mortem:
* I didn't find docs on how saltfs code is deployed now; if that had existed I could probably have reverted in the repo and followed the docs
* https://github.com/servo/servo/wiki/Buildbot-administration still mentions /srv/salt/, which no longer exists
* We landed https://github.com/servo/saltfs/pull/539 before we were able to deploy it, putting saltfs master in a known-bad state
* There is no log of deploys (that I could find), which would have made it easier to verify that a bad deploy had happened
* There is no obvious / documented way to revert a deployment
* The changes were deployed, apparently without someone watching that everything still worked after
* No issue was filed when the problem was first discovered

Manishearth:
Bad deploy was mine; I was testing out the intermittent tracker stuff. It's much harder to deploy from branch now, though.

larsbergstrom:
It's my fault for landing the Android gating assuming it would be "ready to go shortly" and then taking forever to get mozjs building on Android.

We do try to log (see https://github.com/servo/servo/wiki/Buildbot-administration#applying-changes), but there's no default verbose / logged mode for salt highstates, at least that I know how to turn on. It's something we've definitely wanted!

aneeshusa:
#549 covers documentation, and #437 tracks logging, which is addressed by #547. We're still gating on Android but recent builds are all green, so I'm closing this.

PR #549, opened 2016-12-08
Move most Salt docs in tree

aneeshusa:
This also includes Vagrant information,
and should cover all the steps of a normal Salt workflow.

This does not (yet) include information about
setting up new Salt masters/minions,
or how to restart various services cleanly after a deploy.

r? @edunham @larsbergstrom 
cc @Ms2ger per [this comment](https://github.com/servo/saltfs/issues/545#issuecomment-263861942)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/549)

### DIFF
--- a/docs/salt.md
+++ b/docs/salt.md
@@ -1,6 +1,244 @@
 # Notes on Salt
 
-## Looking up previous runs
+## Salt overview
+
+We use [SaltStack](https://saltstack.com/) (Salt for short)
+to configure our infrastructure machines.
+
+We're currently on the Salt 2016.3 release branch, so make sure to look at the
+[right version of the docs](https://docs.saltstack.com/en/2016.3/contents.html).
+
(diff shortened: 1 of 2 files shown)


larsbergstrom:
I think I'm generally OK with this.

@edunham what do you think?

edunham:
As long as we replace the corresponding wiki docs with pointers to the docs in tree, I'm fine with this. 

My only concern is that if we end up with parallel sets of docs on the same topic, one or the other will fall out of date and deciding which to follow becomes an unnecessary burden.

aneeshusa:
Yes, I'll replace those sections on the wiki with references to the in-tree docs after this merges.

Thanks for looking over this! @bors-servo r=edunham

aneeshusa:
I updated the Buildbot and SaltStack Administration pages on the wiki to remove the duplicate information and add pointers to the in-tree docs.

Issue #553, opened 2016-12-13
Error with CheckRevisionStep


larsbergstrom:
I discovered that factories.py had not been getting recompiled & we were using an old version. Once buildbot started using it, if `CheckRevisionStep` is enabled, then we get:
http://build.servo.org/builders/arm32/builds/4619/steps/generic/logs/err.text

For now, I've just commented out adding it to the list, as it's not clear to me how to fix the issue.

cc @aneeshusa @edunham

aneeshusa:
@larsbergstrom, the build you linked seems to be an example of `CheckRevisionStep` working properly by failing the build when the `git` step doesn't succeed sucessfully - Buildbot seems to handle the Exception cleanly, but let me know if it didn't.

However, the behavior in all of the the linked bugs is a bug, which I've filed #554 for.

aneeshusa:
One further idea here might be to re-invoke the git step a bounded number of times if we notice it didn't work properly the first time instead of just failing straight away, but I will leave that for a follow-up.

aneeshusa:
Fix has been deployed and is working.

PR #554, opened 2016-12-13
Make CheckRevisionStep Twisted-compatible

aneeshusa:
The inlineCallbacks decorator requires that the function it wraps
(here, the run() method of CheckRevisionStep) is a generator, so that
is is able to suspend execution to require it to insert callbacks.

Make run() into a generator by yielding the Deferred return value.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/554)


aneeshusa:
Fixes #553; r? @larsbergstrom

PR #562, opened 2016-12-22
Use correct Windows GNU builder name for Homu

aneeshusa:
Fixup for #491. r? @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/562)


larsbergstrom:
@bors-servo r+

Clearly I should not be allowed around a keyboard this week.

PR #569, opened 2016-12-30
Use UTC timezone

charlesvdv:
Fix the issue #303 and follow up of #439 PR.

I'm not sure about the spot to put the ```sls.common``` tests.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/569)

### DIFF
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -72,4 +72,10 @@ else
     if [[ "${SALT_NODE_ID}" == "servo-master1" ]]; then
         ./test.py sls.buildbot.master sls.homu sls.nginx
     fi
+
+    # Salt doesn't support timezone.system on OSX
+    # See https://github.com/saltstack/salt/issues/31345
+    if [[ ! "${SALT_NODE_ID}" =~ servo-mac* ]]; then
+        ./test.py sls.common
+    fi
 fi
--- a/common/init.sls
+++ b/common/init.sls
@@ -45,6 +45,11 @@ servo:
     - shell: /bin/bash
     - home: {{ common.servo_home }}
 
+{% if grains['os'] != 'MacOS' %}
+UTC:
+    timezone.system
+{% endif %}
+
 {% for hostname, ip in common.hosts.items() %}
 host-{{ hostname }}:
   host.present:
--- a/tests/sls/common/timezone.py
+++ b/tests/sls/common/timezone.py
@@ -1,0 +1,18 @@
+import subprocess
+
+from tests.util import Failure, Success
+
+
+def run():
+    ret = subprocess.run(
+        ['date'],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+
+    stdout = ret.stdout.decode('utf-8')
+
+    if ret.returncode == 0 and 'UTC' in stdout:
+        return Success('Date is in UTC')
+    else:
+        return Failure('Date is not in UTC: ', stdout)
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ matrix:
       os: linux
       sudo: required
       dist: trusty
+      language: python
+      python: 3.5
     - env:
         - SALT_NODE_ID=servo-mac1
         - SALT_FROM_SCRATCH=true
@@ -25,6 +27,8 @@ matrix:
       os: linux
       sudo: required
       dist: trusty
+      language: python
+      python: 3.5
     - env:
         - SALT_NODE_ID=servo-master1
         - SALT_FROM_SCRATCH=true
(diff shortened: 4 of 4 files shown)


charlesvdv:
Looks like timezone is not supported for macOS : https://github.com/saltstack/salt/issues/31345

.travis/dispatch.sh
```diff
@@ -72,4 +72,5 @@ else
     if [[ "${SALT_NODE_ID}" == "servo-master1" ]]; then
         ./test.py sls.buildbot.master sls.homu sls.nginx
     fi
+    ./test.py sls.common
```
aneeshusa: We should only run this on the non-macOS machines for now, so we can check the `${SALT_NODE_ID}` isn't a macOS node; easiest way is probably using the `=~` bash operator.

common/init.sls
```diff
@@ -45,6 +45,9 @@ servo:
     - shell: /bin/bash
     - home: {{ common.servo_home }}
 
+UTC:
```
aneeshusa: Use Jinja to render this state conditionally, only on non-macOS, using an `{% if ... %}` block.

tests/sls/common/timezone.py
```diff
@@ -0,0 +1,17 @@
+import subprocess
+
+from tests.util import Failure, Success
+
+
+def run():
+    command = 'date'
+    ret = subprocess.run(command,
```
aneeshusa: Inline the command and pass as an array so we control word-splitting, etc.
```
subprocess.run(['date'], ...)
```

aneeshusa: Also, prefer to use block indent:
```
ret = subprocess.run(
    ['date'],
    stdout=subprocess.PIPE,
    stderr=subprocess.PIPE
)
```

.travis/dispatch.sh
```diff
@@ -72,4 +72,10 @@ else
     if [[ "${SALT_NODE_ID}" == "servo-master1" ]]; then
         ./test.py sls.buildbot.master sls.homu sls.nginx
     fi
+
+    # Salt doesn't support timezone.system on OSX
+    # See https://github.com/saltstack/salt/issues/31345
+    if [[ "${TRAVIS_OS_NAME}" != "osx" ]]; then
```
aneeshusa: In the future the list of tests to run will be auto-generated from the Salt top file, so I'd prefer to check `${SALT_NODE_ID}` instead of `${TRAVIS_OS_NAME}`. You can use e.g. 
```
if [[ ! "${SALT_NODE_ID}" =~ servo-mac* ]]; then
   ...
fi
```

charlesvdv: Sorry about that. I thought it would be less error-prone (because the servo-mac* could change at some point) but if it needs to be travis independant I totally get that.

tests/sls/common/timezone.py
```diff
@@ -0,0 +1,18 @@
...
+    ret = subprocess.run(
+        ['date'],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+
+    stdout = ret.stdout.decode('utf-8')
+
+    if ret.returncode == 0 and 'UTC' in stdout:
+        return Success('Date is in UTC')
```
aneeshusa: I was reading an article about testing yesterday which reminded me of expected-failure tests, which is IMO a nicer way to handle the macOS situtation. Roughly, instead of just returning `Success` here, use `sys.platform` to check if we're on macOS, and if so, return `Failure` in that case with an appropriate message; and vice-versa for the `Failure`. This inverts the expectations and lets us run the `sls.common` test unconditionally.

Note that in the .travis.yml file, I believe `python` isn't supported on macOS, so don't update those entries to add the python3.5 spec. However, we may need to install Python 3.5 manually with `brew` on the OS X builders; make a separate commit and see if it's necessary before adding that.


It may be a little tricky to add so I'll make it optional, for bonus points :)

charlesvdv: Should I put the brew python install inside a saltscript or inside the .travis/install...sh ?

aneeshusa: Let's not add the brew install unless the test suite fails to run on macOS on Travis (i.e. check first), but if we need it I'd probably put it in the dispatch script before we print the Python version.

aneeshusa:
If you go for the bonus points, in the `.travis/dispatch.sh` script please also print the Python3 version unconditionally before running the test suite instead of just for the `test` node.

charlesvdv:
The macOS test fails before testing because of the ```set -o errexit``` in dispatch.sh

aneeshusa:
I believe installing Python 3.5 on the Mac builders actually takes a bit of time, so let's skip the xfail stuff for now. Everything else looked good, so discard the last commit and squash the first five, and we'll get this landed.

aneeshusa:
Also, please rebase on top of the latest master to pull in #573.

aneeshusa:
Thanks for the patch @charlesvdv! @bors-servo r+

Issue #443, opened 2016-07-20
Ensure all Buildbot builders listed for servo repo in Homu configuration are valid


aneeshusa:
The Homu configuration (`homu/cfg.toml`) lists a number of Buildbot repos that are used for gating/try builds, in the `builders` and `try_builders` fields of `repo.servo.buildbot`. Recently, the `windows` builder [got renamed](https://github.com/servo/saltfs/commit/ac587293db495213cf9fbf16881206f5cb42e88a) to `windows-dev` in the Buildbot configuration without updating the Homu configuration, causing breakage. We should ensure that all the builders listed are valid builders.

Files:
- Make a new test file `tests/sls/homu/valid_builders.py`.

The test should read in the templated, deployed Homu configuration using the `toml` library, and examine those two fields, checking them against the list of builders specified in the deployed Buildbot configuration to make sure they all exist.

Note: successfully reading the deployed Buildbot configuration is not that easy, because we need to import the `master.cfg` file, but it doesn't have a `.py` extension! If you look at how Buildbot 8 [does this](https://github.com/buildbot/buildbot/blob/2e5c5ced05615ec80f03b40d7bfe80322cf04f9f/master/buildbot/config.py#L121), it looks like this requires usage of `exec`. (I don't see anything in either `importlib` or `imp` that would help here.) I consider this a necessary evil for now.

Note 2: there may be problems due to the tests using Python 3 and the Buildbot configuration using Python 2; it is likely that a Python 2 subprocess will be needed to extract the list of builders from the configuration.

Tip: if you rename the `master.cfg` file to `master.py`, you can import the module from a Python REPL to look at the available bindings to get to the list of builders.

loganmhb:
I am interested in giving this a try.

aneeshusa:
@loganmhb Great! Let us know if you have any questions.

loganmhb:
Just starting out with this, but it looks like the imp.load_source() function might enable us to load master.cfg without the use of exec, if that seems preferable. (See http://stackoverflow.com/questions/2601047/import-a-python-module-without-the-py-extension)

aneeshusa:
@loganmhb If you can make it work without `exec`, I'd be overjoyed. Thanks for double-checking the docs!

aneeshusa:
@loganmhb How's this going? Do you have any more questions?

loganmhb:
[...]

PR #571, opened 2017-01-02
Add test to verify if all homu builders are listed in the buildbot config

charlesvdv:
That should resolve the #443.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/571)

### DIFF
--- a/tests/sls/homu/get_buildbot_cfg.py
+++ b/tests/sls/homu/get_buildbot_cfg.py
@@ -1,0 +1,30 @@
+from __future__ import print_function
+
+import imp
+import json
+import os.path
+import sys
+
+BUILDBOT_MASTER_PATH = '/home/servo/buildbot/master/'
+
+
+def main():
+    sys.path.append(BUILDBOT_MASTER_PATH)
+
+    config = imp.load_source(
+        'config',
+        os.path.join(BUILDBOT_MASTER_PATH, 'master.cfg')
+    )
+
+    for sched in config.c['schedulers']:
+        if sched.name == 'servo-auto':
+            out = {'builders': sched.builderNames}
+            print(json.dumps(out))
+            return 0
+
+    print('error: "servo-auto" scheduler not found', file=sys.stderr)
+    return 1
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/tests/sls/homu/valid_builders.py
+++ b/tests/sls/homu/valid_builders.py
@@ -1,0 +1,41 @@
+import json
+import os.path
+import subprocess
+import toml
+
+from tests.util import Failure, Success
+
+
+def run():
+    homu_cfg = toml.load('/home/servo/homu/cfg.toml')
+    homu_builders = homu_cfg['repo']['servo']['buildbot']
+
+    # We need to invoke a new process to read the Buildbot master config
+    # because Buildbot is written in python2.
+    scriptpath = os.path.join(os.path.dirname(__file__), 'get_buildbot_cfg.py')
+    ret = subprocess.run(
+        ['/usr/bin/python2', scriptpath],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    if ret.returncode != 0:
+        return Failure(
+            'Unable to retrieve buildbot builders:', ret.stderr
+        )
+
+    buildbot_builders = json.loads(ret.stdout.decode('utf-8'))['builders']
+
+    failure_msg = ''
+    for builder_set in ['builders', 'try_builders']:
+        diff = set(homu_builders[builder_set]) - set(buildbot_builders)
+        if diff:
+            failure_msg += 'Invalid builders for "{}": {}\n' \
+                .format(builder_set, diff)
+
+    if failure_msg:
+        return Failure(
+            "Homu config isn't synced with Buildbot config:",
+            failure_msg
+        )
+
+    return Success('Buildbot and homu configs are synced')


charlesvdv:
The tests are failing because buildbot is installed in the python3 library folder and not the python2 but buildbot is a python2 package which is required to import ```master.cfg```. I don't really know what solutions could be the best:
- symlink
- install buildbot with ```pip2``` (is Salt even support this ? With a quick look at the doc I don't see any ways to change the pip version)
- add a sys.path.append before the dynamic import

Or maybe another idea I didn't think of ?

tests/sls/homu/get_buildbot_cfg.py
```diff
@@ -0,0 +1,16 @@
+import imp
+import json
+import sys
+
+if __name__ == '__main__':
+    sys.path.append('/home/servo/buildbot/master/')
```
aneeshusa: Pull this path out into a constant.

tests/sls/homu/get_buildbot_cfg.py
```diff
@@ -0,0 +1,16 @@
...
+
+    config = imp.load_source(
+        'config',
+        '/home/servo/buildbot/master/master.cfg'
```
aneeshusa: Use `os.path.join` to join the constant and `master.cfg`.

tests/sls/homu/get_buildbot_cfg.py
```diff
@@ -0,0 +1,16 @@
...
+    )
+
+    for sched in config.c['schedulers']:
+        if sched.name == 'servo-auto':
+            out = {'builders': sched.builderNames}
+            print(json.dumps(out))
```
aneeshusa: Add a `return 0` after printing to indicate success and exit early. After the loop (meaning `servo-auto` was not found, print an error message to sys.stderr (e.g. `error: 'servo-auto' scheduler not found) and add a `return 1`.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,40 @@
+import json
+import subprocess
+import toml
+
+from tests.util import Failure, Success
+
+
+def is_builders_valid(buildbot, homu):
```
aneeshusa: grammar: `is_builders_valid` -> `are_builders_valid`

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,40 @@
...
+            return False
+    return True
+
+
+def run():
+    homu_cfg = toml.load('/home/servo/homu/cfg.toml')
+    homu_buildbot = homu_cfg['repo']['servo']['buildbot']
+
+    ret = subprocess.run(
+        ['python2', 'tests/sls/homu/get_buildbot_cfg.py'],
```
aneeshusa: You'll need to use `/usr/bin/python2`. This is because Travis's only has one "mega" Trusty image, on which it installs all its Pythons, Rubies, etc., and the `$PATH` always contains all of those Pythons before the main `/usr/bin`, etc. entries. Hence, just `python2` invokes a Python from `/opt`, not the main Python that we want.

aneeshusa: Also, right now the rest of the test suite is agnostic to CWD, so I'd prefer to refer to the `get_buildbot_cfg.py` file differently, likely using `__file__`.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,40 @@
...
+
+    buildbot_builders = json.loads(ret.stdout.decode('utf-8'))['builders']
+
+    if not is_builders_valid(buildbot_builders, homu_buildbot['builders']):
+        return Failure(
+            'Homu config is not in sync with the buildbot config', ''
+        )
+    if not is_builders_valid(buildbot_builders, homu_buildbot['try_builders']):
+        return Failure(
+            'Homu config is not in sync with the buildbot config', ''
```
aneeshusa: Have this and the previous message be different (ie. reference `builders` vs `try_builders`) to the error message is immediately helpful.

aneeshusa: Additionally, we should make use of the second argument to Failure. To check validity, cast the homu builders and buildbot builders to lists, then subtract the sets to perform a diff. If the `len()` of the diff is non-zero, then pretty-print the mismatched (i.e. remaining in the diff) values to use for the second argument to `Failure()`.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,40 @@
...
+        )
+
+    return Success('Buildbot and homu configs are synced.')
```
aneeshusa: nit: No period at the end

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+                'value': fail
+            }
+
+    return {
+        'success': True
+    }
+
+
+def get_script_path():
+    dirname = os.path.dirname(__file__)
```
aneeshusa: This can be inlined into the next line (should fit within 80 char limit).

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+import os.path
+import subprocess
+import toml
+
+from tests.util import Failure, Success
+
+
+def are_builders_valid(buildbot, homu_cfg, buildertype):
+    for builder in homu_cfg[buildertype]:
+        if builder not in buildbot:
```
aneeshusa: The set differencing subsumes the need to loop over each builder and check if it is present or not, so these first two lines aren't needed.

tests/sls/homu/valid_builders.py
aneeshusa: Also, the whole method can then be inlined into `run()`.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
+import json
+import os.path
+import subprocess
+import toml
+
+from tests.util import Failure, Success
+
+
+def are_builders_valid(buildbot, homu_cfg, buildertype):
```
aneeshusa: Inline `are_builders_valid` into `run()`, which allows removing a bunch of duplicated error checking.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+    for builder in homu_cfg[buildertype]:
+        if builder not in buildbot:
+            diff = list(set(homu_cfg[buildertype]) - set(buildbot))
+            diff_print = ''
```
aneeshusa: This line is unnecessary.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+import subprocess
+import toml
+
+from tests.util import Failure, Success
+
+
+def are_builders_valid(buildbot, homu_cfg, buildertype):
+    for builder in homu_cfg[buildertype]:
+        if builder not in buildbot:
+            diff = list(set(homu_cfg[buildertype]) - set(buildbot))
```
aneeshusa: No need to cast to a list.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+            diff_print = ''
+            if len(diff) != 0:
+                diff_print = 'Difference: {}'.format(diff)
```
aneeshusa: This can be inlined into the `Failure` constructor.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+
+            fail = Failure(
+                'Homu "{}" config isn\'t sync with buildbot config'
```
aneeshusa: nit: sync -> synced
Also, using double quotes for this string will make it easier to read (no need to escape the apostrophe).

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+def get_script_path():
+    dirname = os.path.dirname(__file__)
+    return os.path.join(dirname, 'get_buildbot_cfg.py')
+
+
+def run():
+    homu_cfg = toml.load('/home/servo/homu/cfg.toml')
+    homu_buildbot = homu_cfg['repo']['servo']['buildbot']
```
aneeshusa: I think `homu_builders` is more clear.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,62 @@
...
+        stderr=subprocess.PIPE
+    )
+    if ret.returncode != 0:
+        return Failure(
+            'Unable to retrieve buildbot builders name:', ret.stderr
+        )
+
+    buildbot_builders = json.loads(ret.stdout.decode('utf-8'))['builders']
+
+    resp = are_builders_valid(buildbot_builders, homu_buildbot, 'builders')
```
aneeshusa: You can use a loop to cut down on some duplication:
```
for builder_type in ['builders', 'try_builders']:
    diff = ....
    if diff:
        return Failure(...)

return Success(..)
```

tests/sls/homu/valid_builders.py
aneeshusa: It would also be nice to add a comment about why we need to invoke a Python2 subprocess to run the separate script.

charlesvdv:
@aneeshusa This should be OK now! 

Sorry for the long review process 😞

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,37 @@
...
+        stderr=subprocess.PIPE
+    )
+    if ret.returncode != 0:
+        return Failure(
+            'Unable to retrieve buildbot builders name:', ret.stderr
+        )
+
+    buildbot_builders = json.loads(ret.stdout.decode('utf-8'))['builders']
+
+    for buildertype in ['builders', 'try_builders']:
```
aneeshusa: nit: `buildertype` -> `builder_set`

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,37 @@
...
+        diff = set(homu_builders[buildertype]) - set(buildbot_builders)
+        if diff:
+            return Failure(
+                "Homu {} config isn't synced with buildbot config"
```
aneeshusa: nits:
- add a colon at the end
- capitalize Buildbot

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,37 @@
...
+                .format(buildertype),
+                'Difference: {}'.format(diff)
```
aneeshusa: Before the for loop, initialize a string var (e.g. `failure_msg`) to an empty string. In the loop, if the `if diff:` condition is true, append information about the diff to `failure_msg`.

After the loop, `if failure_msg` then return a `Failure` passing `failure_msg` for the second argument. This way if there invalid builders for `builders` and `try_builders` they're both reported.

Prose changes:
  - Keep the same title (first arg) for the Failure, but don't mention the builder set
   - Use a more descriptive prefix than `Difference`, maybe `Invalid builders for {builder_set}:` (this gets added to `failure_msg` iteratively)

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,37 @@
...
+    # because buildbot is written in python2.
+    scriptpath = os.path.join(os.path.dirname(__file__), 'get_buildbot_cfg.py')
+    ret = subprocess.run(
+        ['/usr/bin/python2', scriptpath],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE
+    )
+    if ret.returncode != 0:
+        return Failure(
+            'Unable to retrieve buildbot builders name:', ret.stderr
```
aneeshusa: nit: Use `'Unable to retrieve Buildbot builders:'`

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,37 @@
...
+
+from tests.util import Failure, Success
+
+
+def run():
+    homu_cfg = toml.load('/home/servo/homu/cfg.toml')
+    homu_builders = homu_cfg['repo']['servo']['buildbot']
+
+    # We need to invoke a new process to read the buildbot master config
+    # because buildbot is written in python2.
```
aneeshusa: nit: capitalize both instances of Buildbot

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,42 @@
...
+
+    buildbot_builders = json.loads(ret.stdout.decode('utf-8'))['builders']
+
+    failure_msg = ''
+    for builder_set in ['builders', 'try_builders']:
+        diff = set(homu_builders[builder_set]) - set(buildbot_builders)
+        if diff:
+            if failure_msg:
+                failure_msg += '\n'
+            failure_msg += 'Invalid builders for "{}": {}'.format(builder_set, diff)
```
aneeshusa: Linter says this line is too long.

tests/sls/homu/valid_builders.py
```diff
@@ -0,0 +1,42 @@
...
+        )
+
+    buildbot_builders = json.loads(ret.stdout.decode('utf-8'))['builders']
+
+    failure_msg = ''
+    for builder_set in ['builders', 'try_builders']:
+        diff = set(homu_builders[builder_set]) - set(buildbot_builders)
+        if diff:
+            if failure_msg:
+                failure_msg += '\n'
```
aneeshusa: You can actually include the `\n` unconditionally as part of the format string - the outputter is smart and will strip the last trailing newline.

aneeshusa:
Thanks for tackling this! @bors-servo r+

PR #572, opened 2017-01-03
Upload MSVC Nightlies from MSYS environment

UK992:
This should fix uploading MSVC nightlies, this also need updated step in `buildbot_steps.yml` to prefix `bash -l`.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/572)


aneeshusa:
I don't know our Windows environments very well. r? @larsbergstrom

larsbergstrom:
@bors-servo r+

Yes, this should work - we have the same machines, so we can just rely on MSYS to do the upload as usual. Great catch, @UK992!

PR #573, opened 2017-01-03
Install autoconf@2.13 from main homebrew-core repo

aneeshusa:
Homebrew has been making a number of changes towards improved support
for installing multiple versions side by side. As part of this change,
autoconf213 has been renamed to autoconf@2.13,
and has been migrated from the `homebrew-versions` repository
(which is being deprecatetd) to the `homebrew-core repository`,
meaning `homebrew-versions` no longer needs to be tapped.
Additionally, the improved side-by-side installation support means
both versions cleanly install together, meaning the custom script
to handle linking is no longer needed.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/573)


aneeshusa:
r? @larsbergstrom @edunham 

This breaking change by Homebrew means no PRs can be landed until this one is.

Issue #560, opened 2016-12-20
Create Dockerfile for Linux TC builds


aneeshusa:
- [ ] Create basic Dockerfile
- [ ] Partition Salt rules to avoid installing Buildbot deps in our Docker containers
- [ ] Figure out image building/pushing story
- [ ] Move all Docker-related code to main servo repo to allow auto-build on change

PR #577, opened 2017-01-12
Refactor to enable mach bootstrap (and leaner Docker builds)

aneeshusa:
This refactor will make it possible to run the states installing Salt build dependencies
without making other system level changes, for use in mach bootstrap.

This will also allow skipping unnecessary states for use in Docker image creation #560, for faster image builds and reduced image size.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/577)


aneeshusa:
OK, this is good enough to allow `./mach bootstrap` for Ubuntu at least!

r? @larsbergstrom

larsbergstrom:
This looks pretty good to me. I think the machine <-> IP files might be a bit out of date (missing some macs, listing one that we don't use anymore) but otherwise it should be good to go. Thanks!

aneeshusa:
Oh, I totally forgot about that, I pulled in the code from #369 so we can get rid of those unused static hosts. Actually tested this in a VM, added some fixes to make it work and updated it to make it ready to add new OS support, r? @larsbergstrom (for real this time)

PR #580, opened 2017-01-22
Quote repo names to ensure valid TOML

aneeshusa:
Otherwise repos like `download.servo.org`
which contain dots in their name are not parsed correctly;
using quotes ensures the repo name is parsed as a single string.

Discovered while trying to merge https://github.com/servo/download.servo.org/pull/16.
I have little TOML experience, so I'm not sure if this actually works or will break our `cfg.toml` entirely.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/580)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -252,16 +252,16 @@ password = "{{ pillar["homu"]["buildbot-http-pass"] }}"
 
 {% for repo in travis_repos %}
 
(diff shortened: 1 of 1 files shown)


notriddle:
Unless the repo name contains a `"`, this should be right.

notriddle:
See <https://github.com/toml-lang/toml#table>

@bors-servo r+

Issue #462, opened 2016-08-17
Update build machines to use Ubuntu 16.04


glennw:

glennw:
cc @larsbergstrom @edunham

aneeshusa:
This is going to be non-trivial, I can think of a few hurdles at least:
- Python 3 is the new default, Python 2is not installed by default, and Salt as of right now is still Python 2 only (although working on Py3 compat). This may affect bootstrapping and our other services (Buildbot, Homu).
- The init system has been changed from Upstart to Systemd, so we'll need to rewrite all of our init scripts.
- Some of our bootstrapping scripts and Salt code conflates Ubuntu and Trusty, which will need to get ironed out/updated for Xenial.

That being said, I agree this is a desirable change.
@glennw Are there any reasons in particular you'd like to upgrade?

aneeshusa:
Also, SSH is updated to a new version which disables DSA keys by default, so we'll need to replace the existing DSA keys. (Currently only @michaelwu, I believe.)

metajack:
It shouldn't take much to rewrite the init scripts. I wrote the current ones and have written both upstart and systemd ones in the past.

Installing python2 seems easy enough.

We want to upgrade in order to test WebRender. Basically need need a version of mesagl that supports OpenGL 3.3. The one in Trusty does not, but the one in Xenial does.

aneeshusa:
OK, I think the easiest way to do it would be adding a new build machine that's running 16.04 and to update the Salt code to handle Trusty and Xenial in parallel, then shift everything over to 16.04 once the kinks are ironed out.

Salt has a built-in grain (`init`) we can branch on to do Upstart things vs Systemd things, and it looks like Python 2 (and 3 for Homu) are already Salted.

metajack:
First issue I've run into is that https://repo.saltstack.com/apt/ubuntu/16.04/amd64/archive/ doesn't have 2015.5.8. Are any of those versions there compatible with our current master?

aneeshusa:
Unfortunately, no.

[...]

Issue #417, opened 2016-06-30
Manage `sources.list` and make bootstrap script purge lists before updating


edunham:
Ran into `SystemError: E:Unable to parse package file /var/lib/apt/lists/us-west-2.ec2.archive.ubuntu.com_ubuntu_dists_trusty-updates_main_i18n_Translation-en (1), E:The package lists or status file could not be parsed or opened.` again today. We've been running into similar apt issues every week or two on the linux slaves. 

@aneeshusa, @larsbergstrom, what are your thoughts on adding a salt state to `rm -r /var/lib/apt/lists/; sudo apt-get update` before applying other states on the Linux hosts, to circumvent future instances of this issue?

aneeshusa:
I did some searching and this appears to be a problem with EC2's apt mirroring. In the interest of being more reproducible, here is what I think is a better solution:
  - Use Salt to manage the `/etc/apt/sources.list` file, like we already manage files in `/etc/apt/sources.list.d/`, so that it never points at EC2 mirrors.
  - Potentially update the bootstrap script to replace the `sources.list` file and run `rm -r /var/lib/apt/lists/*` and `sudo apt-get update` to make sure new machines don't exhibit this problem.
  - Manually run `rm -r /var/lib/apt/lists/*` and `sudo apt-get update` on the existing machines.

edunham:
Sounds good to me. This would be an appropriate first bug for someone with intermediate Ubuntu familiarity :)

choudhary001:
I would like give this a go. Where should i start with this?

aneeshusa:
@choudhary001 Great, it's yours :)

Are you familiar with `apt`s `/etc/apt/sources.list` file? A good starting point would be adding a new [`file.managed`](https://docs.saltstack.com/en/2015.5/ref/states/all/salt.states.file.html#salt.states.file.managed) state to the `ubuntu/init.sls` file which locks down the contents of the `/etc/apt/sources.list` file on our Ubuntu builders, using official Ubuntu mirrors instead of EC2 mirrors.

choudhary001:
@aneeshusa I am familiar with /etc/apt/sources.list file. I looked into 'https://docs.saltstack.com/en/latest/ref/states/all/salt.states.file.html' as well. But I didn't understand the locking down contents of the sources.list file. Could explain it a bit more?

aneeshusa:
[...]

PR #582, opened 2017-01-22
Use an up-to-date llvm.

emilio:
This allows us to update OsMesa, see:

 * https://github.com/servo/osmesa-src/pull/15
 * https://github.com/servo/servo/pull/14584

Please wait for ci, since I don't have any Debian-based system.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/582)

### DIFF
--- a/servo-build-dependencies/init.sls
+++ b/servo-build-dependencies/init.sls
@@ -30,6 +30,7 @@ servo-dependencies:
       - libgles2-mesa-dev
       - libosmesa6-dev
       - libssl-dev
+      - llvm-3.5-dev
       - xorg-dev
       - xpra
       - xserver-xorg-input-void


aneeshusa:
FYI, you can [use Vagrant to test locally](https://github.com/servo/saltfs/blob/master/docs/salt.md#testing-with-vagrant), although we also test the Salt configs on Travis.

Is this blocked on the other two PRs or standalone? Changes LGTM either way, so r=me when this is ready.

emilio:
This is standalone, the other two PRs are blocked on this one.

Thanks for reviewing this @aneeshusa! :)

aneeshusa:
FYI, I don't think `llvm-3.8-dev` is available in the default repos for Trusty (14.04), although `llvm-3.5-dev` (via `security`) and llvm-3.4-dev should be.

emilio:
Those should definitely be enough. Travis builds seem green, though I think they use a newer distro version?

Anyway, 3.5 should be enough, so will change to that.

aneeshusa:
Travis doesn't use the standard sources lists I believe, see #417, so it might be adding some extra repos. We're using Trusty for now on Travis, as Xenial isn't supported yet, see my latest comment at #462.

PR #583, opened 2017-01-23
Install llvm in OSX builders

emilio:
Same deal than the previous PR. One build at a time.

r? @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/583)


emilio:
Gah, forgot about alphabetic ordering

PR #590, opened 2017-01-30
Add additional reviewers to webrender repositories

jdm:
This seems like the state of the world right now.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/590)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -169,8 +169,7 @@ secret = "{{ pillar["homu"]["web-secret"] }}"
         "extra_reviewers": [ "jrmuizel", "vvuk" ]
     },
     {
-        "name": "webrender_traits",
-        "extra_reviewers": [ "jrmuizel", "vvuk" ]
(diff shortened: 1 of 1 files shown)


homu/files/cfg.toml
```diff
@@ -165,10 +165,12 @@ secret = "{{ pillar["homu"]["web-secret"] }}"
...
     },
     {
-        "name": "webrender"
+        "name": "webrender",
+        "extra_reviewers": [ "jrmuizel", "vvuk" ]
     },
     {
-        "name": "webrender_traits"
+        "name": "webrender_traits",
+        "extra_reviewers": [ "jrmuizel", "vvuk" ]
```
emilio: This is probably useless because wr-traits now effectively lives in the wr repo.

emilio:
I'd approve this myself, but I think I can't, so maybe r? @glennw

jdm:
Rebased in #669.

PR #591, opened 2017-02-02
Use host 0.0.0.0

Manishearth:
bottle doesn't like empty hosts anymore

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/591)


aneeshusa:
I'd prefer this change be made in Homu itself for the default setting.

PR #596, opened 2017-02-07
Prepare for Salting Windows

aneeshusa:
Avoid running states on Windows minions so that salt '*' state.highstate does not cause spurious errors while our states haven't been ported to Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/596)


aneeshusa:
r? @larsbergstrom 

I'd like to have the top file changes in place before actually trying to install Salt on the Windows machines.

Issue #376, opened 2016-05-19
Run nightly tests of the saltfs repo


aneeshusa:
We should run the full saltfs suite (from scratch) on Travis every night to catch changes in upstreams (e.g. #314).

@edunham I think you like to use [nightli.es](https://nightli.es/) for this purpose? We should also have some form of alerting on failure, but I'm not sure of a good way to do that.

edunham:
We can configure Travis to send email or IRC notifications of build status, https://docs.travis-ci.com/user/notifications/. I think emails will be the best communication channel for me. @larsbergstrom, would / where would you like notifications?

larsbergstrom:
I agree that email is probably the best of the options. The awesomest would be automatically opening a GH issue in the case of failure, but I don't think that's an available option :-)

aneeshusa:
Ping on this, would have helped catch #483.

larsbergstrom:
@edunham could we use an alias on servo.org? e.g., github-saltfs@servo.org

aneeshusa:
I opened #376 for this, but I see a few problems with this solution:
- I am unaware of any way to filter notifications by branch. Because `homu` always uses the `auto` branch, even though Travis does not send notifications for PRs this will send extra emails when going through `bors-servo`, causing lots of spam.
- Having an (internal?) email list adds more state, somewhere that is not 1) openly visible 2) likely not under version control of some kind. What does the process look like for changing that list?
- If we add this to other repos, we'd start wanting one list per repo, which is a lot of lists to keep track of (e.g. which lists am I on? did we remember to enable notifications on [new repo] yet?)

I think it may be time to invoke Zawinski's Law and add email features to homu, where we can add a few more smarts and standardize this a bit, making it available to all repos easily.

aneeshusa:
Apparently Travis has [a new feature called Crons](https://github.com/travis-ci/beta-features/issues/1) which allows natively scheduling nightly builds. It looks like this is still in Beta but on the cusp of going GA - I believe it can be enabled through the Travis console, possibly after a support request.

I think IRC notifications would be fine to start with, although I suppose an email alias would work as well.

edunham:
Nightly builds are now enabled and #598 adds notification in the bots channel on all builds.

aneeshusa:
@edunham Do you know if this is working? Just got hit by a new failure (see #601), but I'm not sure if Travis ran a nightly build yesterday.

edunham:
[...]

PR #598, opened 2017-02-09
notify travis status in servo-bots channel

edunham:
this is the other half of fixing https://github.com/servo/saltfs/issues/376 

i have already turned on nightly builds via travis's experimental cron feature

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/598)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,3 +80,5 @@ script: .travis/dispatch.sh
 
 notifications:
   webhooks: http://build.servo.org:54856/travis
+  irc: "irc.mozilla.org#servo-bots"
+


aneeshusa:
What is the `#servo-bots` channel? Is there someone (a human) monitoring this so we find out about failed nightly builds?

edunham:
#servo-bots is a channel where all of Servo's bots dump information, so that we can use our IRC clients to filter those notifications which we care about. Jack and I idle in there, and others are welcome to as well. 

Dropping notifications into the bots channel basically just makes it easy to "subscribe" to whatever subset of notifications we each find relevant, while making sure we don't clutter the channels which exist for human conversation.

aneeshusa:
OK, the only other suggestion I had was maybe using `on_success` and `on_failure` to pre-filter which notifications Travis sends to IRC, but if you're filtering in your client it's not necessary. r=me with or without that.

Issue #581, opened 2017-01-22
Add test to check all repos in Homu's cfg.toml are valid


aneeshusa:
This will prevent typos in repo names as well as issues like #580.

Files:
- Make a new test file `tests/sls/homu/valid_repos.py`

The test should read the deployed `cfg.toml` file for Homu (should be at `/home/servo/homu/cfg.toml`) using the `toml` Python library. Each subkey under the `repo` key of the configuration should be queried for existence on GitHub as a valid repo.

Bonus points if you query the GitHub API for the webhooks of each repo to check that an appropriate webhook is set up (see https://github.com/servo/servo/wiki/Adding-a-repo-to-Homu), although I think this might require an new access token with permissions we don't necessarily want to provide.

aneeshusa:
This can be tested locally using the `servo-master1` VM [in Vagrant](https://github.com/servo/saltfs/blob/master/docs/salt.md#testing-with-vagrant); look at other tests under `tests/sls/homu` for similar examples that read the deployed configuration.

aneeshusa:
It looks like we'd need the `read:repo_hook`, which only allows read and ping abilities for webhooks on public and private repos, per [the GitHub v3 API docs](https://developer.github.com/v3/oauth/). I think this is a reasonably small scope to provide to Travis to integrated into CI, since Travis already knows our webhook secret.

rwaweber:
Hey @aneeshusa, I think I may give this a shot tonight. 

Also, mentioned it in #servo, but since I had a tough time grabbing the dependencies (at least on El Capitan) for the vagrant environment, I setup this cask: https://github.com/rwaweber/homebrew-servo-saltfs.

aneeshusa:
Great @rwaweber, ask if you have any questions.

BTW, did you try installing Vagrant 1.8.1 from [the Hashicorp releases](https://releases.hashicorp.com/vagrant/) or via [an older Cask URL](https://github.com/caskroom/homebrew-cask/blob/b43b81d80661f38b6fe94c3f6262c862942d6fb3/Casks/vagrant.rb) (at least, I think you can do `brew cask install <some_url>`...)?

rwaweber:
I've got a WIP [here](https://github.com/servo/saltfs/pull/606).

A couple things:
[...]

PR #606, opened 2017-02-18
Check all repos in Homu's cfg.toml are valid

rwaweber:
Enumerates the inconsistencies between repositories configured in homu
vs those configured in the servo github organization.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/606)

### DIFF
--- a/tests/sls/homu/valid_repos.py
+++ b/tests/sls/homu/valid_repos.py
@@ -1,0 +1,36 @@
+from urllib.error import URLError
+from urllib.request import Request, urlopen
+
+import toml
+
+from tests.util import Failure, Success
+
+
+def repo_exists(url):
+    '''
+    Checks if the given repo exists on GitHub
+    '''
+    try:
+        request = Request("https://github.com/{}".format(url), method='HEAD')
+        with urlopen(request) as conn:
+            return conn.status == 200
+    except URLError:
+        return False
+
+
+def run():
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    homu_repos = (
+        "{}/{}".format(repo['owner'], repo['name'])
+        for repo in repo_cfg.values()
+    )
+    missing_repos = [
+        repository for repository in homu_repos
+        if not repo_exists(repository)
+    ]
+    if len(missing_repos) > 0:
+        return Failure(
+            'Some repos set up for Homu do not exist on GitHub:',
+            "\n".join(" - {}".format(repo) for repo in missing_repos)
+        )
+    return Success('All repos in the Homu config exist on Github')


jdm:
@aneeshusa or @edunham Could you take a look at this?

edunham:
The logic looks correct to me, and the whole thing superficially looks like it *should* work. However, both the master tests fail on Travis. 

Could you please add a test in the repo_pager to bail out with a descriptive error message if the auth credentials passed into it didn't work? 

(the following is my notes from debugging it -- I might want them later, but you're free to ignore them)

So as you mentioned in your comment on #581, we're sticking a string into the gh-access-token [here](https://github.com/servo/saltfs/blob/master/.travis/test_pillars/homu.sls#L2) where we should actually be reading an *env var* named `TEST_HOMU_GH_ACCESS_TOKEN` after having set the `TEST_HOMU_GH_ACCESS_TOKEN` to a working token at https://travis-ci.org/servo/saltfs/settings . So then the string which is not in fact a token at all gets blindly pulled in to the config without being validated as a token [here](https://github.com/servo/saltfs/blob/master/homu/files/cfg.toml), and is then (i *think*) successfully read out as if it were the token on line 38 of the new test. 

So then the actual error we get is [string indices must be integers](https://travis-ci.org/servo/saltfs/jobs/212499025#L2455). I *think* that means that the req_list is getting filled with strings instead of dicts that you can sanely ask for a html_url out of, and that's happening up in the repo_pager.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,59 @@
...
+import requests
+from tests.util import Failure, Success
+
+
+def repo_pager(endpoint, repoblock, auth):
+    """
+    recursively consumes an endpoint and an empty (self-populating) list
+    it will exit when it fails to find the 'next' key in the links dict
+    """
+    req = requests.get(endpoint, headers=auth)
```
edunham: If the endpoint returns anything about failed auth, please bail out right here with a descriptive error message.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,59 @@
...
+        access_token = homu_cfg['github']['access_token']
+        auth = {'Authorization': 'token '+access_token}
+        req_list = []
+        repo_pager('https://api.github.com/orgs/servo/repos', req_list, auth)
+        # grok out the html_url's
+        # when the github api fails to authenticate, the html_url key will not
+        # exist in the req dicts
+        gh_repos = [req['html_url'] for req in req_list]
+    except TypeError as e:
+        return Failure('Unable to authorize to github API', str(e))
```
rwaweber: After https://github.com/servo/saltfs/pull/606/commits/d83bedf34bb3baa3e30104a8c916e384044f89fb I feel that the need to catch this exception is significantly diminished, since the authorization check is now handled in the repo_pager function.

rwaweber:
@aneeshusa Thanks for the review! 

> just make a HEAD request for each repo to see if it exists, e.g.requests.head('https://github.com/servo/servo').status_code == 200

Agreed! I think that would be a great deal simpler. Would also allow me to remove the hackery around the `repo_pager` function necessary to get through the github api, I must've misunderstood the original request in the issue statement.

Just for the sake of clarity, we only want to make sure that all of the repositories in the homu configuration are configured on github, correct?

Not sure about the rate limiting risk here, as there would now be a request for every repo configured. But, github /probably/ wouldn't rate limit on there http frontend and with ~50 requests. This is something to figure out with testing.

Would it also be preferred to consider using `urllib` instead of introducing an extra dependency in the form of `requests`? I'm perfectly happy to rewrite it that way.

aneeshusa:
Yes, the intent is to ensure all the repositories in the Homu config exist on GitHub.

I didn't think about rate limiting earlier, I thought the token was to get read access to the list of repos under the organization. I do think we'll need a token to work around rate limiting since IIRC there's a limit of 60 requests an hour to the API. However, we can use a separate token for the test suite and just read it from the environment instead of gathering it from the Homu config (see [my comment](https://github.com/servo/saltfs/pull/627#issuecomment-293394570) on the PR for the token). Of course, if it works without a token, that's even better!

It would be nice to use urllib and avoid the dependency (especially since it's more ergonomic in Python 3), but I'm also ok with using requests.

rwaweber:
Lemme know if there's anything you'd want me to change with the above! 

From what I can tell(by checking the `rate_limit` endpoint), these requests are not affected by the 60 requests-an-hour limit imposed by the github api. Downside is, that it makes the `homu.sls` tests run a good 10-15 seconds slower, which is sort of annoying. Though, getting around that limitation may end up being more trouble than it's worth.

jdm:
@aneeshusa review ping!

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
+import toml
```
aneeshusa: Split up the imports into three groups: Python modules (the urllib imports), external modules (toml) and internal modules (tests.util). Separate each group of imports by a blank line, and alphabetize in each group.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+    '''
+    try:
+        response = urllib.request.urlopen(url).status
+    except URLError:
+        response = url
+    return response
+
+
+def run():
+    # repository configuration dictionary from homu
```
aneeshusa: nit: Remove comment

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    VCS = "https://github.com/"
+    # extracting owner and repo from the configuration dict
```
aneeshusa: nit: Remove comment

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
+import toml
+import urllib.request
+from urllib.error import URLError
+from tests.util import Failure, Success
+
+
+def getStatus(url):
```
aneeshusa: Let's make this a `repo_exists` method, which takes a repo identifier string in the form `owner/name`, e.g. `servo/saltfs`, and returns a boolean. Update the doc string as appropriate.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+    '''
+    Consumes a url string and returns the status code of a GET request
+    '''
+    try:
+        response = urllib.request.urlopen(url).status
```
aneeshusa: Use HEAD requests to reduce bandwidth usage - you can create a `urllib.request.Request` object with `method='HEAD'`, and pass that to `urlopen`.
Use a with statement for `urlopen` to automatically close the response.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def run():
+    # repository configuration dictionary from homu
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    VCS = "https://github.com/"
+    # extracting owner and repo from the configuration dict
+    # and formatting it to more easily form a url to submit a request to
+    homu_repos = [repo_cfg[repo_title]['owner']+'/'+repo_title
+                  for repo_title in repo_cfg.keys()]
```
aneeshusa: Iterating over `repo_cfg.values()` and using the `'owner'` and `'name'` keys for each repo will make this a little cleaner.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+    return response
+
+
+def run():
+    # repository configuration dictionary from homu
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    VCS = "https://github.com/"
+    # extracting owner and repo from the configuration dict
+    # and formatting it to more easily form a url to submit a request to
+    homu_repos = [repo_cfg[repo_title]['owner']+'/'+repo_title
```
aneeshusa: I prefer to use string formatting, e.g. `'{}/{}'.format(repo['owner'], repo['name'])`

aneeshusa: Also, this can be a generator expression instead of making a temporary list.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+                  for repo_title in repo_cfg.keys()]
+    failed_responses = [repository for repository in homu_repos
+                        if getStatus(VCS+repository) != 200]
+    failed_resp_str = " \n".join(failed_responses)
```
aneeshusa: Inline this inside the `Failure`constructor to avoid constructing it if not necessary.
Use `'\n'` to join the responses (no preceding space), and precede each entry with a `- `, which makes a nice list.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+
+def run():
+    # repository configuration dictionary from homu
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    VCS = "https://github.com/"
+    # extracting owner and repo from the configuration dict
+    # and formatting it to more easily form a url to submit a request to
+    homu_repos = [repo_cfg[repo_title]['owner']+'/'+repo_title
+                  for repo_title in repo_cfg.keys()]
+    failed_responses = [repository for repository in homu_repos
```
aneeshusa: Rename this to `missing_repos`.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,33 @@
...
+        response = urllib.request.urlopen(url).status
+    except URLError:
+        response = url
+    return response
+
+
+def run():
+    # repository configuration dictionary from homu
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    VCS = "https://github.com/"
```
aneeshusa: Inline this the `urllib.request.Request` constructor in `repo_exists`.

tests/sls/homu/valid_repos.py
rwaweber: ++ on the generator recommendation, I've never used those before so that was neat to play around with them!

rwaweber:
Sorry about the double push! Didn't see the message about the capitalized error message until it was too late. 😅

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
+from urllib.error import URLError
+from urllib.request import Request, urlopen
+
+import toml
+
+from tests.util import Failure, Success
+
+
+def repoExists(identifier):
```
aneeshusa: It's more Pythonic to use snake_case: `repo_exists`.

Also, I think `repo` is more clear than `identifier`.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+    '''
+    Consumes a repo identifier string in the form of owner/name and returns a
```
aneeshusa: No need to repeat all the details of how the checking is done, the docstring only needs to describe the overall goal of the method, e.g. `Checks if the given repo exists on GitHub`.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+    boolean indicating whether the request to check if the repository exists
+    on github was successful (200) or not
+    '''
+    try:
+        endpoint = "https://github.com/{}".format(identifier)
+        requester = Request(endpoint, method='HEAD')
```
aneeshusa: nit: `request` instead of `requester`

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+        with urlopen(requester) as conn:
+            if conn.status != 200:
```
aneeshusa: You can `return conn.status == 200` directly.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+                response = False
+            else:
+                response = True
+    except URLError:
+        response = False
```
aneeshusa: You can `return False` directly.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def repoExists(identifier):
+    '''
+    Consumes a repo identifier string in the form of owner/name and returns a
+    boolean indicating whether the request to check if the repository exists
+    on github was successful (200) or not
+    '''
+    try:
+        endpoint = "https://github.com/{}".format(identifier)
```
aneeshusa: nit: `endpoint` -> `url`. This can also be inlined into the `Request` constructor if you want.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+            else:
+                response = True
+    except URLError:
+        response = False
+    return response
+
+
+def run():
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    # formatting to more easily form a url to submit a request to
```
aneeshusa: Comment is unnecessary, this style is fairly common anyways.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+    homu_repos = ("{}/{}".format(repo['owner'], repo['name'])
```
aneeshusa: Prefer block indent, i.e.
```
homu_repos = (
    "{}/{}".format(...)
    for repo in repo_cfg.values()
)
```

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+                  for repo in repo_cfg.values())
+    missing_repos = ["- {}".format(repository) for repository in homu_repos
```
aneeshusa: Use block indent.
Do the `"- {}"` formatting inside the if block, `missing_repos` can just be the list of repositories itself.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+                     if not repoExists(repository)]
+    if len(missing_repos) > 0:
+        return Failure('Repos in homu not on github: ',
+                       "\n".join(missing_repos))
+    else:
+        return Success('All repos in homu config on github')
```
aneeshusa: This can be at the top level instead of inside an `else`, as the `if` case does an early return.

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,39 @@
...
+
+def run():
+    repo_cfg = toml.load('/home/servo/homu/cfg.toml')['repo']
+    # formatting to more easily form a url to submit a request to
+    homu_repos = ("{}/{}".format(repo['owner'], repo['name'])
+                  for repo in repo_cfg.values())
+    missing_repos = ["- {}".format(repository) for repository in homu_repos
+                     if not repoExists(repository)]
+    if len(missing_repos) > 0:
+        return Failure('Repos in homu not on github: ',
```
aneeshusa: Use block indent.
No need for a space after the colon.
I'd make the message a little more descriptive, e.g. `Some repos configured for Homu do not exist on GitHub:`.

tests/sls/homu/valid_repos.py
aneeshusa: grammar: capitalize Homu, GitHub, `in homu config` -> `in the Homu config`, `on github` -> `exist on GitHub`

rwaweber:
Hey @aneeshusa, just lemme know if you want me to look at anything else. If not, I'm happy to get down to squashing!

tests/sls/homu/valid_repos.py
```diff
@@ -0,0 +1,37 @@
...
+    homu_repos = (
+        "{}/{}".format(repo['owner'], repo['name'])
+        for repo in repo_cfg.values()
+    )
+    missing_repos = [
+        repository for repository in homu_repos
+        if not repo_exists(repository)
+    ]
+    if len(missing_repos) > 0:
+        return Failure('Some repos set up for Homu do not exist on GitHub:',
```
aneeshusa: one last nit: use block indent for this. This also means the second argument will all fit on one line, i.e.:
```
    return Failure(
        'Some repos...on GitHub:',
        '\n'.join(' - {}'.format(repo) for repo in missing_repos)
    )
```

edunham:
@aneeshusa review ping :)

aneeshusa:
@rwaweber, all the code looks good! Please update the commit message to have a descriptive title for the change and include some motivation/details as a body. I like https://chris.beams.io/posts/git-commit/ as a guide to good commit messages.

aneeshusa:
Looks great, thanks for the PR and your patience, @rwaweber!

@bors-servo r+

rwaweber:
Thanks for taking the time to teach, @aneeshusa. Looking forward to the next one!

PR #615, opened 2017-03-09
Use separate homu user for Homu-related states

### LINKED ISSUE #422: Use separate homu user for Homu-related states
Homu has no need to interact with the Buildbot-related files under servo, and neither should Buildbot be able to see Homu's file. We should be using a separate `homu` user for Homu-related states, which can often be done via a `user: homu` parameter to those states. We also will need to add a state that adds the required `homu` user.

Part of #421.

Files to change:
- `homu/init.sls` is where to change the Salt states.
- Add a tests in the `tests/sls/homu` directory to check file permissions. It should check a) all the files under the Homu deployment path are owned by Homu and b) that none of those files are world-readable, or at least none of the config files are world-readable.

ALikhachev:
Fixes #422 

- Added 'homu' user
- Added parameters to run pip and virtualenv under 'homu' user
- Changed homu config owner and permissions to prevent it to be read by strangers :)
- Added code that tests these changes

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/615)

### DIFF
--- a/homu/init.sls
+++ b/homu/init.sls
@@ -9,8 +9,13 @@ homu-debugging-packages:
       - sqlite3
 
 homu:
+  user.present:
+    - fullname: Homu
+    - shell: /bin/bash
+    - home: /home/servo/homu
   virtualenv.managed:
     - name: /home/servo/homu/_venv
+    - user: homu
     - venv_bin: virtualenv-3.5
     - python: python3
     - system_site_packages: False
@@ -18,6 +23,7 @@ homu:
       - pkg: python3
       - pip: virtualenv
   pip.installed:
+    - user: homu
     - pkgs:
       - git+https://github.com/servo/homu@{{ homu.rev }}
       - toml == 0.9.1  # Please ensure this is in sync with requirements.txt
@@ -36,9 +42,9 @@ homu:
   file.managed:
     - source: salt://{{ tpldir }}/files/cfg.toml
     - template: jinja
-    - user: servo
-    - group: servo
-    - mode: 644
+    - user: homu
+    - group: homu
+    - mode: 640
 
 /etc/init/homu.conf:
   file.managed:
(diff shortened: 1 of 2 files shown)


homu/init.sls
```diff
@@ -9,15 +9,21 @@ homu-debugging-packages:
       - sqlite3
 
 homu:
+  user.present:
+    - fullname: Homu
+    - shell: /bin/bash
+    - home: /home/servo/homu
```
aneeshusa: Homu should have its own home directory completely separate from servo, i.e. `/home/homu`. We'll also need to have some manual deployment steps for this to move the homu db to the right place by hand.

aneeshusa: We should also create a `homu` group as the primary group of the `homu` user.

tests/sls/homu/config_permissions.py
```diff
@@ -0,0 +1,26 @@
...
+    return pwd.getpwuid(os.stat(filename).st_uid).pw_name
+
+
+def is_world_readable(filename):
+    st = os.stat(filename)
+    return bool(st.st_mode & stat.S_IROTH)
+
+
+def run():
+    for root, directories, filenames in os.walk('/home/servo/homu/'):
```
aneeshusa: This will also check files like `.profile` for the homu user. I think it's OK if we only check permissions on Homu's `cfg.toml` file (and the database if it exists).

homu/init.sls
```diff
@@ -36,9 +42,9 @@ homu:
   file.managed:
     - source: salt://{{ tpldir }}/files/cfg.toml
     - template: jinja
-    - user: servo
-    - group: servo
-    - mode: 644
+    - user: homu
+    - group: homu
+    - mode: 640
```
aneeshusa: Since this file is no longer world-readable, we'll need to run the tests with `sudo` from `.travis/dispatch.sh`. Please put this change in a separate commit first to make sure it doesn't break anything else.

mrobinson:
Closing as homu is no more.

PR #617, opened 2017-03-13
Update the jdk version

fabricedesre:
This is needed for the gradle based build system (see https://github.com/servo/servo/pull/15773#issuecomment-285882132)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/617)

### DIFF
--- a/servo-build-dependencies/android.sls
+++ b/servo-build-dependencies/android.sls
@@ -23,7 +23,7 @@ android-dependencies:
       - libc6:i386
       - libstdc++6:i386
       {% endif %}
-      - default-jdk
+      - openjdk-8-jdk
       - ant
       - expect
       - gcc


edunham:
If this has spent more than an hour or so after landing without being deployed, ping me on IRC and I'll roll it out. I'll try to check back on it through the afternoon.

edunham:
I just highstated the cross builders. The change should be deployed, unless the Salt gitfs wizardry has failed us.

larsbergstrom:
Ugh, Ubuntu 14.04.5 LTS does not have `openjdk-8-jdk`, but this did not fail in testing or on deploy. Only when I try to `apt-get install` on the machine do I get:
```
E: Unable to locate package openjdk-8-jdk
```

larsbergstrom:
@aneeshusa Does ^ seem familiar to you? Is there something I'm doing wrong that could have caught this?

KiChjang:
The package name is apparently `openjdk-8-jre`, according to this http://openjdk.java.net/install/.

larsbergstrom:
I think we may need to do more work, given that we're on 14.01 - http://askubuntu.com/a/666481

PR #618, opened 2017-03-13
Update android build tools

MortimerGoro:
This is needed for the gradle based build system too. See servo/servo#15773 and https://github.com/servo/saltfs/pull/617

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/618)


aneeshusa:
See also #629 for a fixed version of #617.

PR #620, opened 2017-03-15
Remove windows-gnu-dev builds

nox:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/620)


nox:
We are currently deprecating these builds, as discussed with @metajack.

PR #626, opened 2017-03-22
Add clang dependency required for rust-bindgen

MortimerGoro:
This allows us to use crates which depend on rust-bindgen.

Required for https://github.com/servo/servo/pull/16059. There are more rust-bindgen based dependencies coming soon (https://crates.io/crates/gvr-sys)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/626)


MortimerGoro:
clang-3.5 should be enough for now (right @emilio ?)

We can update to clang-3.8 when the 16.04 based machines are ready: cc @larsbergstrom @fabricedesre @metajack

PR #627, opened 2017-03-23
Set up taking a GitHub access token 

edunham:
Duct tape fix to #606. I have created a token with no special perms from my account and stuck it in the appropriate slot in the Travis UI to expose it as an env var to this repo's builds. We can change the contents of that env var through the UI at any time so it'll be easy enough to switch it to a token with different scopes, and eventually to one on Homu's account, in the process of debugging this.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/627)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -76,7 +76,11 @@ matrix:
       language: python
       python: 3.5
 
-script: .travis/dispatch.sh
+before_script:
+    - echo $GHTOKENLINE >> .travis/test_pillars/homu.sls
+
+script:
+    - .travis/dispatch.sh
 
 notifications:
   webhooks: http://build.servo.org:54856/travis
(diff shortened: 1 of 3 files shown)


aneeshusa:
Closing in favor of an approach that uses HEAD requests to GitHub to query repo existence, which shouldn't require a token, as mentioned in my review. If we ever update the test to also check the configured webhooks, we can set up a token at that point.

aneeshusa:
Oh, I see this is to get around rate-limiting instead, we'll probably need this.

aneeshusa:
Instead of conflating this with Homu's GitHub access token, I would prefer to have a separate token for the test suite to use when making requests to GitHub. This means we can use the environment variable directly and don't have to sneak it through Salt.

aneeshusa:
Also, it seems like Travis doesn't like the existing token either and complains that it is malformed: https://travis-ci.org/servo/saltfs/jobs/221121527

aneeshusa:
We didn't need a token for #606 (just used HEAD requests); we can revisit this later if needed.

PR #628, opened 2017-03-27
Replace shared root login with user accounts

edunham:
These changes: 

* Create individual user account for everyone who has an SSH key
* Add each SSH key to its person's account instead of to root
* Forbid root login at ssh config level (OSX and Ubuntu)
* Permit passwordless sudo for each user who has an SSH key. This is not great but it is no worse than the current state of affairs. The options for moving away from passwordless sudo will all potentially break peoples' workflows, so I'd like to write and test them as a separate set of changes from this one. 

r? @aneeshusa to catch my inevitable Salt room-for-improvement

cc @larsbergstrom for thoughts on if/how we should actively disallow root login on Windows

Labeling please-don't-merge because I'm 90% sure something in here will break Travis.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/628)


edunham:
Oh actually I had a couple old PRs about this which were broken for various reasons. I'll integrate the suggestions from them before consuming anyone's review time.

larsbergstrom:
From the logs, it looks like the creation step needs to be depended upon by the keyfile installation step? There are a bunch of OSX errors about `[ERROR   ] Failed to add the ssh key. Is the home directory available, and/or does the key file exist?`

admin/files/sshd_config
```diff
@@ -0,0 +1,28 @@
...
+
+# Root login is not allowed for auditing reasons. This is because it's difficult to track which process belongs to which root user:
+#
+# On Linux, user sessions are tracking using a kernel-side session id, however, this session id is not recorded by OpenSSH.
+# Additionally, only tools such as systemd and auditd record the process session id.
+# On other OSes, the user session id is not necessarily recorded at all kernel-side.
+# Using regular users in combination with /bin/su or /usr/bin/sudo ensure a clear audit track.
+PermitRootLogin No
+
+# Use kernel sandbox mechanisms where possible in unprivilegied processes
```
aneeshusa: spelling: unprivileged

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
...
+    {% elif grains['os'] == 'Ubuntu' %}
+    - group: root
+    {% endif %}
+    - mode: 644
+    - source: salt://{{ tpldir }}/files/sshd_config
+
+
+{% for ssh_user in admin.ssh_users %}
+{{ ssh_user }}:
+    user.present:
```
aneeshusa: Before `user.present`, we should also use `group.present` to create a group for each user with the same name. We should require the group state from the user state and set `gid_from_name: True` in the user state.

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
...
+        - empty_password: True
+        - optional_groups:
```
aneeshusa: We should use `groups` instead of `optional_groups` so Salt complains if this doesn't work.

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
...
+            - wheel
```
aneeshusa: Might be nice to add a `group.present` state for the `wheel` group itself.

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
...
+    ssh_auth.present:
```
aneeshusa: This should `require` the `user.present` state.

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
...
+        - user: {{ ssh_user }}
+        - source: salt://{{ tpldir }}/files/ssh/{{ ssh_user }}.pub
+
 {% endfor %}
+
+# FIXME This is just as bad as all sharing root login.
+/etc/sudoers:
+    file.append:
```
aneeshusa: I've personally never been a fan of `file.append`, I prefer to use `file.managed` and control the whole contents instead of hoping that Salt is always smart enough to do the right thing with `file.append`.

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
...
+        - text:
+            {% for ssh_user in admin.ssh_users %}
```
aneeshusa: We could provide permissions to the `wheel` group instead of each user individually.

admin/init.sls
```diff
@@ -27,9 +27,36 @@ UTC:
     - mode: 644
     - source: salt://{{ tpldir }}/files/hosts
 
-{% for ssh_user in admin.ssh_users %}
-sshkey-{{ ssh_user }}:
-  ssh_auth.present:
+# FIXME we should also explicitly AllowUsers usera userb userc
```
aneeshusa: We could also use `AllowGroups wheel`, I think.

aneeshusa:
It might also be nice to switch to a non-standard port to reduce log spam, since we've set the log level to verbose.

aneeshusa:
It might also be cool to integrate https://github.com/mozilla/ssh_scan as a test.

aneeshusa:
If we want to implement #657, I suggest the following changes:
- Don't provide sudoers access to these per-user accounts; we'll implement the Publisher ACL in a separate PR
- Keep root SSH access enabled for now

After testing Publisher ACL, we can enable sudoers access if necessary and drop root SSH access.

admin/init.sls
edunham: I believe we want to end up with as few people having root access as possible, so most users who can access the hosts to examine logs or whatever will not be in `wheel`. Thus they'll either need to be individually added with `AllowUsers`, or all added to a group which in turn is added using `AllowGroups`. While a dedicated group would have the same effect as individually allowing, it adds a layer of indirection when trying to figure out which users have access compared to just looking at `sshd_config`. 

So my FIXME was pretty unclear, but I'm pretty sure it's an issue that `AllowGroups wheel` does not adequately address.

edunham:
Feedbacks addressed; merging in master and fixing the conflicts now

admin/init.sls
```diff
@@ -22,22 +22,37 @@ UTC:
...
+    group.present:
+        - members:
+            - {{ ssh_user }}
+    user.present:
+        - empty_password: True
+        - gid_from_name: True
+        - groups:
+            - wheel
+        - require:
+            - group:
```
aneeshusa: Unless this is a new feature, I don't think this works and we'll need:
```
- group: wheel
- group: {{ ssh_user }}
```
as two separate lines.

admin/init.sls
```diff
@@ -22,22 +22,37 @@ UTC:
...
+    - mode: 644
+    - source: salt://{{ tpldir }}/files/sshd_config
+
+wheel:
+    group.present
+
+{% for ssh_user in admin.ssh_users %}
+{{ ssh_user }}:
+    group.present:
+        - members:
```
aneeshusa: We need to create the group first, and then the user, so this state shouldn't have a `members` key as that would be a circular dependency. The `gid_from_name` parameter in the `user.present` state will take care of this.

aneeshusa: FYI, since the `group.present` will be empty at that point, just set it to an empty list as so to get the YAML to parse correctly: https://docs.saltstack.com/en/2016.3/topics/troubleshooting/yaml_idiosyncrasies.html#yaml-does-not-like-double-short-decs

admin/init.sls
```diff
@@ -22,22 +22,37 @@ UTC:
...
+    - user: root
+    {% if grains['os'] == 'MacOS' %}
+    - group: wheel
+    {% elif grains['os'] == 'Ubuntu' %}
+    - group: root
+    {% endif %}
+    - mode: 644
+    - source: salt://{{ tpldir }}/files/sshd_config
+
+wheel:
```
aneeshusa: I agree with your comment about [only providing some users root access](https://github.com/servo/saltfs/pull/628#discussion_r147214123). To prevent confusion, let's rename this group to `sshusers`, and reserve the `wheel` group for users that will have root perms.

admin/init.sls
```diff
@@ -22,22 +22,37 @@ UTC:
...
+    user.present:
+        - empty_password: True
+        - gid_from_name: True
+        - groups:
+            - wheel
+        - require:
+            - group:
+                - wheel
+                - {{ ssh_user }}
+    ssh_auth.present:
```
aneeshusa: We'll need to keep the two-state solution of `file.managed` to continue being able to handle SSH key rotation and revocation.

admin/init.sls
```diff
@@ -22,22 +22,37 @@ UTC:
...
 
-sshkeys-dir:
-  file.directory:
-    - name: {{ root.home }}/.ssh
-    - user: {{ root.user }}
-    - group: {{ root.group }}
-    - mode: 700
-
-sshkeys:
+/etc/sshd_config:
```
aneeshusa: We should keep the states that set up the SSH keys for the root user, as we don't want to lock ourselves out (and at least for now want to keep installing root SSH keys on new machines).

aneeshusa:
Also, since it [doesn't look like the Publisher ACL feature panned out as I was expecting](https://github.com/servo/saltfs/issues/657), feel free to bring back the sudoers-related states. A rebase would also be helpful for cleaning up the history a bit.

admin/init.sls
edunham: oops, yeah. Fixed.

edunham:
@aneeshusa Should be more-right. I un-clobbered your SSH states, restricted it to not trying to do Mac because getting account creation right on OSX is :man_shrugging:, but kept the sshd config addition. 

Looking ready for squash?

admin/init.sls
edunham: Fixed :)

Issue #601, opened 2017-02-11
Pin all dependencies when installing Python packages via pip


aneeshusa:
This ensures we use known good versions and won't have random failures when new versions of packages are uploaded to pypi.

aneeshusa:
Currently, it looks like #600 is failing because [Twisted `17.1.0` was released today on pypi](https://pypi.python.org/pypi/Twisted).

PR #630, opened 2017-04-12
Pin Twisted version for builders

### LINKED ISSUE #588: Add a third linux-cross build machine
Android and Linux ARM builds have become a frequent bottleneck in Servo's buildbot capacity.  Adding a third buildslave to run these jobs would let them run in parallel, reducing the total turnaround time.  CC @larsbergstrom and @edunham.

aneeshusa:
I was following up on #588 and saw that a `servo-linux-cross3` machine has been provisioned, but the buildmaster doesn't know about it. Checking the logs, it appears that the buildbot-slave service doesn't start successfully, and the stack trace seems to be the same reason as https://github.com/servo/saltfs/issues/601#issuecomment-279163471, namely a version of Twisted that is too new. This PR is similar to #604, but pins the Twisted version for builders instead of the buildmaster.

I also included some more requisites.
This should fix #588 once deployed.

r? @edunham @larsbergstrom

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/630)

### DIFF
--- a/buildbot/master/init.sls
+++ b/buildbot/master/init.sls
@@ -11,7 +11,7 @@ buildbot-master:
       - txgithub == 15.0.0
       - boto == 2.38.0
       - pyyaml == 3.11
-      - twisted == 16.6.0
+      - twisted == 16.6.0  # NOTE: keep in sync with buildbot-slave sls
     - require:
       - pkg: pip
   service.running:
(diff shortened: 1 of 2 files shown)


larsbergstrom:
This is awesome - thanks much!

@bors-servo r+

aneeshusa:
I updated this to work around a Salt behavior that caused the build to fail previously.

PR #631, opened 2017-04-12
WIP: Add mach bootstrap test

aneeshusa:
This is a new test to ensure saltfs changes don't break `./mach bootstrap` for Servo.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/631)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,77 +6,77 @@ language: cpp
 # sudo: required and dist: trusty enable usage of Trusty
 matrix:
   include:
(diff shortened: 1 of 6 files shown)


aneeshusa:
It seems the Travis builders are detected as `'debian'` by `mach bootstrap`, so I've added Debian support here as well, which is the same as the Ubuntu case. This is now blocked on servo/servo#16389.

aneeshusa:
Now blocked on servo/servo#16505.

PR #629, opened 2017-04-11
Use openjdk-r PPA for OpenJDK8 on Ubuntu 14.04

aneeshusa:
This makes #617 work properly by using the openjdk-r PPA for the OpenJDK 8 packages.
r? @larsbergstrom 
cc @fabricedesre

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/629)

PR #638, opened 2017-04-17
Set JAVA_HOME env var to OpenJDK 8 for Android builds

aneeshusa:
The new gradle builds require Java 8,
and the existing ant builds also work with Java 8.

This is easier than running many `update-alternatives` calls from Salt.
Moreover, this allows keeping Java 7 installed together with Java 8.

Needed for servo/servo#15773.
Follow-up to to #617 and #629.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/638)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -79,6 +79,8 @@ build_linux = build_common + Environment({
 build_android = build_linux + Environment({
     'ANDROID_NDK': '{{ common.servo_home }}/android/ndk/current/',
     'ANDROID_SDK': '{{ common.servo_home }}/android/sdk/current/',
+    # TODO(aneeshusa): Template this value for e.g. macOS builds
+    'JAVA_HOME': '/usr/lib/jvm/java-8-openjdk-amd64',
     'PATH': ':'.join([
         '/usr/local/sbin',
         '/usr/local/bin',


aneeshusa:
@MortimerGoro can you please confirm that JAVA_HOME works with the ant build, so we don't break the current Android build?

MortimerGoro:
@aneeshusa I ran some tests in my machine:

**Gradle**
- Compiles ok with JAVA_HOME set to Java8 (/usr/lib/jvm/java-8-openjdk-amd64)
- Fails as expected with JAVA_HOME set to Java 7 (/usr/lib/jvm/java-7-openjdk-amd64): Unsupported major.minor version 52.0

**Ant**
- Compiles ok with JAVA_HOME set to Java8 (/usr/lib/jvm/java-8-openjdk-amd64)
- Compiles ok with JAVA_HOME set to Java 7 (/usr/lib/jvm/java-7-openjdk-amd64)
- Fails as expected when setting an invalid JAVA_HOME folder: Error: JAVA_HOME is not defined correctly. This confirms that ant takes JAVA_HOME into account.

**My environment**:
- Both java7 and java8 installed on Ubuntu 16.04. 
- Both android-18 and android-25 APIs installed.
- Android tools **v25.2.5**. Important: Android tools >= 25.3.x breaks the old build system because they deprecated `android` cli command.

MortimerGoro:
Based on these tests setting JAVA_HOME env variable seems safe.

aneeshusa:
Thanks so much @MortimerGoro! This is ready for review, r? @larsbergstrom

larsbergstrom:
This means that we're probably not going to support cross-compiling from OSX, right? @fabricedesre Is that OK with you?

I know that the OSX cross builds have been broken for a little while (don't test => eventually breaks), but this seems to explicitly move us down a road of Linux-only cross builds. Unless that directory should also be present on OSX, too?

Otherwise, I'm fine with this, assuming it passes a `try`!

fabricedesre:
I don't personally care about cross compiling from OSX, being a linux user. go ahead!

aneeshusa:
@larsbergstrom `JAVA_HOME` is not required in general, we're only using it to point to Java 8 instead of Java 7 on the builders. macOS builds should still work if only Java 8 is installed, or else I believe there are similar directories `JAVA_HOME` can be set to.

It does make sense to template this in the future, though, so I've added a comment to that effect.
@bors-servo r=larsbergstrom

PR #639, opened 2017-04-19
Install android tools r25.2.3 concurrently

aneeshusa:
Install the new Android tools concurrently with the old Android tools, including automatically cleaning up the old version of the tools and handling the directory layout change. (Keep the default tool version the same.)
Backport the `archive.py` state module to fix zip file permissions extraction.
Fix ownership setting on extracted files.
Add testing for the Android SDK due to repeated issues and the `android` tool always exiting with a 0 return code even in case of error.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/639)

### DIFF
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -58,12 +58,13 @@ else
         run_salt 'old'
         set -o errexit
 
+        git checkout "${TRAVIS_COMMIT}"
+
         travis_fold_start "salt.invalidate_cache" 'Invalidating the Salt cache'
-        rm -rf /var/cache/salt/minion/files/base/*
+        salt_call 'saltutil.clear_cache'
         salt_call 'saltutil.sync_all'
         travis_fold_end "salt.invalidate_cache"
 
-        git checkout "${TRAVIS_COMMIT}"
         run_salt 'upgrade'
     fi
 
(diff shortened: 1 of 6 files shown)


aneeshusa:
I got this working! r? @larsbergstrom 
Also cc @MortimerGoro @fabricedesre in case I've made any bad assumptions about how the various Android tools are packaged

jdm:
Amazing!

aneeshusa:
BTW, I would appreciate a highstate/deploy to roll out #638 and confirm that it's working before we apply this PR, to make it easier in case we need to revert something.

MortimerGoro:
lgtm!

larsbergstrom:
@aneeshusa Deployed and confirmed set in:
http://build.servo.org/builders/android/builds/6435

larsbergstrom:
@aneeshusa Wow, this is an epic PR! I think I'm ready to `r+` this if the logs from the run above look good to you. We'll probably have to be a little careful with the rollout to the cross builders and make sure we have folks on call if something goes south. Maybe morning/afternoon tomorrow (Friday), US time? I have some actual meeting-free slots then :-)

aneeshusa:
@larsbergstrom Looks like the `JAVA_HOME` PR is working (or at least didn't break anything). I'm free after 2PM EST tomorrow for a rollout. I also recommend playing with this in Vagrant to see how it changes the filesystem - AFAICT everything under the `sdk/current` symlink (which is the only place Buildbot looks) looks the same before and after.

PR #642, opened 2017-04-21
WIP: Support python nightly upload script

aneeshusa:
Needed to fix servo/servo#16560.

Note in particular the first Salting on Windows: using Salt to install Python 3.
I haven't tested that at all, so I'd recommend rolling that out using the temporary Salt root.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/642)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -210,22 +210,11 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
 
             # Provide environment variables for s3cmd
             elif (
-                arg == './etc/ci/upload_nightly.sh' or
-                arg == r'.\etc\ci\upload_nightly.sh'
+                arg == './etc/ci/upload_nightly' or
+                arg == r'.\etc\ci\upload_nightly'
             ):
                 step_kwargs['logEnviron'] = False
                 step_env += envs.upload_nightly
-                if self.is_windows:
-                    # s3cmd on Windows GNU does not work in MINGW
-                    step_env['MSYSTEM'] = 'MSYS'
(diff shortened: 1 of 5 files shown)


larsbergstrom:
This all looks good, code-wise - will do the test deployment stuff early next week when folks are around, in case I burn something down :-)

aneeshusa:
The second commit will likely need to be updated to add the Python3 bits to the Windows $PATH in the `build_windows_msvc` var in `buildbot/master/files/config/environment.py`. I'm not sure exactly what the folder will be, so I'll wait until the test deployment to finalize that.

metajack:
@larsbergstrom Do you want me to do this?

larsbergstrom:
@metajack That would help! I'm all day in meetings through Friday afternoon, but will get to it then if you can't.

metajack:
python3 package not found on windows

aneeshusa:
I looked into this a little more and I don't think Python3 on Windows is the best direction for this, I will redo this with a different approach.

metajack:
Doh. Do you have more details?

aneeshusa:
A few reasons:
- Salt has support for installing Windows packages via MSIs, which actually works. However, it fails when checking to see if the installation succeeds - it either looks for the wrong key or in the wrong list of packages, I don't remember. Hopefully this is better in more up-to-date versions of Salt (we're over a year behind).
- Python3 on Windows doesn't install a `python3` executable, only a `python` executable, which makes it trickier to make it available on the `PATH`. There is a `py` launcher that apparently can be used to invoke the right Python, but it seems a bit messy.

This may be worth revisiting down the road, but in the meantime I fixed up the servo PR to not requires saltfs changes.

Issue #597, opened 2017-02-07
Allow mach command executed by buildbot to know the PR being tested


jdm:
I'm looking at some useful improvements we could make to our automation, but I'm getting stuck on how to get some information from the buildbot environment into the mach commands being invoked as part of buildbot_steps.py. I'm specifically interested in making available:
* the builder in use
* the platform it's running
* the PR being tested

What's the best way to do this, @aneeshusa? Environment variables?

aneeshusa:
I think environment variables is the best way to go; AFAIK you should be able to use [Buildbot properties](https://docs.buildbot.net/0.8.12/manual/cfg-properties.html#properties), which can be [interpolated](https://docs.buildbot.net/0.8.12/manual/cfg-properties.html#interpolate) at runtime. I believe the builder being used and platform should be easy to expose; however, IIRC I think Homu just sends Buildbot a specific commit to test, not necessarily a PR number - this would likely need to be parsed out manually from the commit message or something.

BTW, can you expand on the commands you'd like to add? I'd like to avoid coupling `mach` commands too tightly to Buildbot, especially with TC somewhere on the horizon.

jdm:
I want to extend the filter-intermittents command to report the intermittents that were filtered out to a database so we can track them over time.

zachncst:
@jdm @aneeshusa  I've been investigating this for working on https://github.com/servo/servo/wiki/Tracking-intermittent-failures-over-time-project as a part of a NCSU masters project. The easiest means of tracking what PR the change is associated with is using this api command `https://api.github.com/search/issues?q=<hash>`. If we just used the buildbot properties to set environmental variables that the mach commands can read then it could decouple the two and make for easy transferring to TC (very easy to set build environment variables there, I've got experience). 

That would mean we'd need 3 ENV variables: `PLATFORM`, `TEST_MACHINE`, `GIT_SHA`. If you're in agreement, what's the best way for myself to implement these into the current buildbot system?

aneeshusa:
@zachncst That API looks like a good find! I've assigned this to you.

[...]

PR #648, opened 2017-04-27
Adding gitSha as a step kwarg and env var

zachncst:
To be able to find the pull request for intermittent issues, we need to add the gitSha into the build steps so the PR can be queried from github. More information here: #597. In support of: https://github.com/servo/servo/wiki/Tracking-intermittent-failures-over-time-project

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/648)

### DIFF
--- a/buildbot/master/files/config/factories.py
+++ b/buildbot/master/files/config/factories.py
@@ -230,6 +230,11 @@ class StepsYAMLParsingStep(buildstep.ShellMixin, buildstep.BuildStep):
             else:
                 step_desc += [arg]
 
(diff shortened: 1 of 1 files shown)


aneeshusa:
Thanks for the PR! Now that I think about it some more, getting the PR can be done entirely from `mach`. `mach` can call `git rev-parse HEAD` to get the current commit (since the full git history is available during Buildbot builds), then hit the GitHub API you found to find the relevant PR number. I'd like to keep as much code in `servo/servo` as possible, so I'm going to close this in favor of that approach.

PR #649, opened 2017-04-27
Use Docker for testing on Travis, start on 16.04 support

aneeshusa:
Use pinned Docker images to test our states on multiple OSes.
Currently, only `servo-linux1` is tested, on Ubuntu 14.04 and 16.04.
Also note that some states are not tested since they don't make
sense in a Docker container, mainly the `service` states;
these still need to be updated for systemd.

This starts updating the Salt states for Ubuntu 16.04 compatibility,
and thus helps with #462.

Also, make the tests run on Python 3.4, which is the default Python 3 on 14.04.
See commit messages for some other miscellaneous changes.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/649)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,23 @@ matrix:
       os: linux
       sudo: required
       dist: trusty
+    # Salt inside Docker
+    - env:
+        - SALT_NODE_ID=servo-linux1
+        - SALT_FROM_SCRATCH=true
+          # ubuntu/14.04
+        - SALT_DOCKER_IMAGE=ubuntu@sha256:edf05697d8ea17028a69726b4b450ad48da8b29884cd640fec950c904bfb50ce
+      os: linux
+      sudo: required
+      dist: trusty
+    - env:
+        - SALT_NODE_ID=servo-linux1
+        - SALT_FROM_SCRATCH=true
+          # ubuntu/16.04
+        - SALT_DOCKER_IMAGE=ubuntu@sha256:f3a61450ae43896c4332bda5e78b453f4a93179045f20c8181043b26b5e79028
+      os: linux
+      sudo: required
+      dist: trusty
     # Not a Salt node, runs test suite instead
     - env:
         - SALT_NODE_ID=test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
-# Python is not supported on OS X yet
-# But a system python is installed for all workers
+# Handle Python installation and dependencies ourselves
 language: cpp
 
 # NOTE: Make sure the matrix covers all node types in top.sls
@@ -14,8 +13,6 @@ matrix:
       os: linux
       sudo: required
       dist: trusty
-      language: python
-      python: 3.5
     - env:
         - SALT_NODE_ID=servo-mac1
         - SALT_FROM_SCRATCH=true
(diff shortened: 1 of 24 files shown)


.travis.yml
```diff
@@ -57,24 +48,35 @@ matrix:
...
+      sudo: required
+      dist: trusty
+    - env:
+        - SALT_NODE_ID=servo-linux1
+        - SALT_FROM_SCRATCH=true
+          # ubuntu/16.04
+        - SALT_DOCKER_IMAGE=ubuntu@sha256:f3a61450ae43896c4332bda5e78b453f4a93179045f20c8181043b26b5e79028
+      os: linux
+      sudo: required
+      dist: trusty
```
metajack: It's no longer Trusty but Xenial, no? Or are we still doing xenial testing within docker on trusty travis machines?

.travis/install_salt.sh
```diff
@@ -43,11 +46,17 @@ configure_salt() {
...
-    sudo rm -rf /etc/salt
-    sudo mkdir -p /etc/salt
-    sudo cp "${FORCE_FLAG}" -- "${TEMPORARY_CONFIG_DIR}/minion" /etc/salt/minion
+    ${SUDO} rm -rf /etc/salt
+    ${SUDO} mkdir -p /etc/salt
+    ${SUDO} cp "${FORCE_FLAG}" -- \
+        "${TEMPORARY_CONFIG_DIR}/minion" /etc/salt/minion
 }
 
+SUDO=""
```
metajack: Instead of duplicating this code, maybe we need a common.bash that these files include.

test.py
```diff
@@ -7,19 +7,20 @@
...
 def run_tests(tests):
     any_failures = False
 
     for test_spec in tests:
         test_dir = os.path.join(project_path(), 'tests', *test_spec.split('.'))
-
-        python_scripts = filter(is_python_script, os.scandir(test_dir))
-        tests = sorted([entry.name for entry in python_scripts])
-
+        # TODO(aneeshusa): switch back to scandirr
```
metajack: I'd probably add a ntoe that this is needed because scandir is from python 3.5. Otherwise this knowledge of what the TODO item is will be lost.

buildbot/master/init.sls
```diff
@@ -14,6 +14,7 @@ buildbot-master:
       - twisted == 16.6.0  # NOTE: keep in sync with buildbot-slave sls
     - require:
       - pkg: pip
+  {% if grains.get('virtual_subtype', '') != 'Docker' %}
```
metajack: Why can't we get systemd running inside docker? Not having service.running seems not great, and certainly pollutes the states.

.travis.yml
aneeshusa: Travis doesn't have Xenial machines, so we're using Xenial Docker containers on their Trusty VMs.

.travis/install_salt.sh
aneeshusa: I thought about doing that, but I don't think it's worth it (yet) since it's only a few lines of duplication.
Having another file also breaks the `curl install.sh | bash` workflow, which is nice to setup another machine.

buildbot/master/init.sls
aneeshusa: The `service.running` states underneath call various commands that seem to want to do IPC to a running instance of the int system (Upstart or Systemd). In the container, there is no init process running (just the `dispatch.sh` inside bash), so the states fail to run. There are purportedly ways to make it all work, but I haven't found a clear guide yet, and I'd like to add at least a basic smoke test now with the option to add `service.running` support later.

I agree this isn't ideal, so I'm open to other ways to handle this.

metajack:
@bors-servo r+

Please file a bug about fixing the systemd issue. I'd like to make sure we eventually remove all those ifdefs :)

PR #651, opened 2017-04-27
Revert a51202bd953e01ce12781d93ff9726ebe7db9a59.

jdm:
When we don't clobber the target directory, we run into problems like https://github.com/servo/servo/issues/16602 and https://github.com/servo/servo/issues/16632 on our builders. Let's revert that until we solve it properly.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/651)


aneeshusa:
I've deployed this and restarted Buildbot.

Issue #654, opened 2017-05-04
Discussion: Should we upgrade pip more aggressively?


edunham:
servo-linux3 got into a bad state and needed its android dir blown away and replaced by a highstate today. That's fine. What's not fine is that the highstate didn't work until after I ran `pip install --upgrade pip` on the host. 

Error from Salt before upgrading Pip:
```
----------
          ID: virtualenv
    Function: pip.installed
      Result: False
     Comment: Error installing 'virtualenv == 14.0.6': Exception:
              Traceback (most recent call last):
                File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 122, in main
                  status = self.run(options, args)
                File "/usr/lib/python2.7/dist-packages/pip/commands/freeze.py", line 74, in run
                  req = pip.FrozenRequirement.from_dist(dist, dependency_links, find_tags=find_tags)
                File "/usr/lib/python2.7/dist-packages/pip/__init__.py", line 286, in from_dist
                  assert len(specs) == 1 and specs[0][0] == '=='
              AssertionError
              
              Storing debug log for failure in /root/.pip/pip.log
     Started: 20:14:15.620240
    Duration: 256.634 ms
     Changes:   
----------
          ID: buildbot-slave-dependencies
    Function: pip.installed
      Result: False
     Comment: Error installing 'buildbot-slave == 0.8.12': Exception:
              Traceback (most recent call last):
                File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 122, in main
                  status = self.run(options, args)
                File "/usr/lib/python2.7/dist-packages/pip/commands/freeze.py", line 74, in run
                  req = pip.FrozenRequirement.from_dist(dist, dependency_links, find_tags=find_tags)
                File "/usr/lib/python2.7/dist-packages/pip/__init__.py", line 286, in from_dist
                  assert len(specs) == 1 and specs[0][0] == '=='
              AssertionError
              
              Storing debug log for failure in /root/.pip/pip.log
     Started: 20:14:15.877244
    Duration: 254.463 ms
     Changes:   
```

Contents of logfile on linux3: 

```
root@(servo-linux3):~# cat /root/.pip/pip.log 
------------------------------------------------------------
/usr/bin/pip2 run on Thu May  4 20:14:16 2017
Exception:
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/pip/basecommand.py", line 122, in main
    status = self.run(options, args)
[...]

PR #655, opened 2017-05-05
add froydnj to the list of try pushers

froydnj:
This would be handy for working on Servo.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/655)


aneeshusa:
Looks like this is hitting a similar case to #654, I'll look into this more closely.

aneeshusa:
Hmm, the Travis cron job succeeded, maybe this was a transient failure.

@bors-servo retry

aneeshusa:
I've just rolled this out, Homu should recognize your new powers now @froydnj :)

Issue #658, opened 2017-05-08
Enable choosing specific build machine for Buildbot force builds


aneeshusa:
We need to add the `BuildslaveChoiceParameter` to our `ForceScheduler`s.
docs: https://docs.buildbot.net/0.8.12/manual/cfg-schedulers.html#buildslavechoiceparameter

edunham:
@aneeshusa, have you re-highstated the buildmaster to deploy this?

aneeshusa:
@edunham Not yet, there are builds in the queue: http://build.servo.org/homu/queue/all

I also made a new label (S-needs-deploy) to help keep track of this in the future.

edunham:
Thank you!

PR #659, opened 2017-05-08
Enable choosing build machine for force builds

aneeshusa:
This makes it easier to test if changes to a specific build machine are
working properly, e.g. the current impetus of seeing if Android-related
changes on servo-linux-cross3 are working.

Docs: https://docs.buildbot.net/0.8.12/manual/cfg-schedulers.html#buildslavechoiceparameter

Fixes: #658.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/659)

### DIFF
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -1,4 +1,5 @@
 from buildbot.plugins import buildslave, changes, schedulers, status, util
+from buildbot.schedulers.forcesched import BuildslaveChoiceParameter
 from buildbot.status import html, status_push, web, words
 
 import environments as envs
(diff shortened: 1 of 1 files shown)


aneeshusa:
The only thing I was concerned about was the correct import paths, but I checked in a REPL and I think those should be OK. r? @edunham

edunham:
@bors-servo r+ 

Looks good to me. I'll keep an eye on the builds after we roll this out -- if the first one refuses to start, we back this out and try again.

aneeshusa:
I've just rolled this out and confirmed that the web UI now allows choosing a particular build machine when running a force build.

PR #660, opened 2017-05-08
Set a maximum priority for builds.

edunham:
This should be in our configs at or before the time we deploy the changes to
get Homu to read the max priority setting.

The whole thing about having non-infinite priorities is to facilitate
sherriffing as discussed in https://github.com/servo/homu/issues/111

I pulled 9001 out of thin air because I consider it easy to memorize.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/660)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -10,6 +10,9 @@ access_token = "{{ secrets['gh-access-token'] }}"
 app_client_id = "{{ secrets['app-client-id'] }}"
 app_client_secret = "{{ secrets['app-client-secret'] }}"
 
+# Priority values above max_priority will be refused.
+max_priority = 9001
+
 [web]
 host = "0.0.0.0"
 port = 54856
--- a/homu/map.jinja
+++ b/homu/map.jinja
@@ -1,6 +1,6 @@
 {%
   set homu = {
     'db': '/var/homu/main.db',
-    'rev': '83689f0d52e64292a4eb50def23d0e9028f8aba1'
+    'rev': '62f0c9e5081e54c178fc75222e66c026a2a69f25'
   }
 %}


aneeshusa:
I'd prefer to have the Homu version bump and config change in the same PR, if possible.

edunham:
@aneeshusa Thanks! I completely neglected to bump the Homu version in the Homu changes, as well. I'll update both patches.

edunham:
@aneeshusa I may be misreading this, but it looks like we're pulling a Homu git revision from the pillar in https://github.com/servo/saltfs/blob/master/homu/init.sls#L22 rather than working from its version at all?

aneeshusa:
@edunham  `{{ homu.db }}` on that line is a Jinja substitution; on the first line of that file we're importing the `homu` dict from the `map.jinja` file in the same folder. The version is set [in that file](https://github.com/servo/saltfs/blob/master/homu/map.jinja).

aneeshusa:
You are correct that we're currently directly pinning the git revision instead of the PyPi version, however. I think using PyPi versions would be better!

edunham:
I think that we may not be using the PyPi version because we are not publishing our Homu to pypi with any sort of regular cadence or automation. For now I think we should land the Homu changeset then I'll update this PR with the new Homu rev of the merge, and I'll add fixing the pypi stuff to the stack of fixups Homu needs :)

edunham:
Fixed it. Sorry about the delay there. r? @aneeshusa

aneeshusa:
r=me after squash; please also clean up the commit message, particularly the bit about landing before the Homu bump, as Salt handles the ordering correctly already.

aneeshusa:
@bors-servo retry

- Intermittent failure connecting to repo.saltstack.com

Issue #661, opened 2017-05-08
Android builds are failing on servo-linux-cross3


aneeshusa:
We've tried re-highstating servo-linux-cross3; jury is still out on the last highstate, but previous attempts have not fixed the problem.

I'm not well versed with Android, but I have a new idea about what might be going wrong:
- The relevant builds (e.g. http://build.servo.org/builders/android/builds/6778/steps/compile/logs/stdio) are currently failing in the `/mach build --android --dev` step
- In [buildbot_steps.yml](https://github.com/servo/servo/blob/master/etc/ci/buildbot_steps.yml), we only override the `ANDROID_SDK` variable for the `./mach package` step, not the `./mach build` step, so the `current` SDK symlink takes precedence
- #644 has not yet been deployed to `-cross1` and `-cross2`, only `-cross3`:
  - Both `-cross1` and `-cross2` still have the older `r24.4.1` SDKs on disk, while all three have the new `r25.2.3` SDK on disk.
  - `-cross1` and `-cross2` have the old SDK as `current`, while `-cross3` has the new SDK as `current`.

Is it feasible that `./mach build --android --dev` is failing due to getting the new `r25.2.3` SDK on `-cross3`, while working OK with the `r24.4.1` SDK on `-cross1` and `-cross2`?

cc @larsbergstrom @MortimerGoro

aneeshusa:
OK, reproduced the failure on servo-linux-cross2 with the above test PR (see http://build.servo.org/builders/android/builds/6786), SDK 25.2.3 breaks `./mach build --android --dev`.

MortimerGoro:
This line in blurdroid dependency is the one causing the problem:

https://github.com/szeged/blurdroid/blob/master/build.rs#L64

The problem is that when SDK 25.2.3 is enabled, only android-25 platform is installed. When r24.4.1 is enabled only android-18 platform is installled. Blurdroid dependency has android-18 path hardcoded...

Possible solutions:
- Add an array of platforms to saltfs instead of using a unique android-platform
- Fix hardcoded path in blurdroid dependency

aneeshusa:
Making platforms into an array in saltfs is a fairly easy change. However, IMO it makes more sense to update blurdroid, then send a servo/servo PR with that change (and also fixing buildbot_steps.yml to pass the ANDROID_SDK var for `./mach build`) for a few reasons:
- Easier to test servo/servo changes since there's no need to revert as with saltfs changes
[...]

PR #663, opened 2017-05-08
Revert "Make r25.2.3 the default Android SDK"

aneeshusa:
This reverts commit a9e75da378eb55180681ab25d172334a607226d4.
`./mach build --android --dev` does not work with SDK 25.2.3.

SDK 25.2.3 is still available for the `./mach package step`.

This will address #661 for now to prevent all Android builds on `cross3` from failing.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/663)

### DIFF
--- a/servo-build-dependencies/map.jinja
+++ b/servo-build-dependencies/map.jinja
@@ -1,7 +1,12 @@
 {%
   set android = {
     'sdk': {
-      'current': 'r25.2.3',
+      'current': 'r24.4.1',
+      'r24.4.1': {
+        'build_tools': '23.0.3',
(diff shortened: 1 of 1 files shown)


aneeshusa:
Currently, I've just used the Buildbot Web UI to perform a "Graceful Shutdown" on the `servo-linux-cross3` builder, which stops the Buildbot service but keeps the machine alive (so we can still SSH in, run Salt, etc.).

aneeshusa:
No longer needed due to servo/servo#16812, which fixed the original issue.

PR #670, opened 2017-05-17
Add Behnam Esfahbod as a reviewer for unicode-bidi

mbrubeck:
@behnam is currently the most active developer of the unicode-bidi crate and has now written or rewritten large portions of the code.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/670)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -70,7 +70,9 @@ secret = "{{ secrets['web-secret'] }}"
     "skia": {},
     "string-cache": {},
     "tendril": {},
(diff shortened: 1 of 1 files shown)


Manishearth:
r+

Manishearth:
wait i forget what's our policy on adding reviewers again, do we need core approval?

jdm:
For our smaller repositories the process can be a bit more fluid. This case is interesting because it's a crate that until now has had very little active development, so we can't really link to any existing reviews. I think this makes sense, however.
@bors-servo: r+

PR #673, opened 2017-05-22
Change Wafflespeanut to wafflespeanut

wafflespeanut:
I'd recently changed my username (so that I have the same thing throughout the web). Apparently, homu is case-insensitive.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/673)


wafflespeanut:
Thanks!

PR #675, opened 2017-05-22
Provide secrets for `./mach upload-nightly`

aneeshusa:
Also remove s3cmd related code,
as we now use boto3 for uploads.

Goes along with servo/servo#16565.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/675)


aneeshusa:
cc @larsbergstrom @metajack

PR #688, opened 2017-06-05
Install sccache

aneeshusa:
Part of servo/servo#17106.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/688)

### DIFF
--- a/servo-build-dependencies/ci.sls
+++ b/servo-build-dependencies/ci.sls
@@ -1,0 +1,14 @@
+{% from 'common/map.jinja' import root %}
+{% from tpldir ~ '/ci-map.jinja' import sccache %}
+
+sccache:
+  file.managed:
+    - name: {{ sccache.destination }}
+    - source: https://s3.amazonaws.com/rust-lang-ci/rust-ci-mirror/{{ sccache.version }}-sccache-{{ sccache.platform }}
+    - source_hash: sha384={{ sccache.sha384 }}
+    - user: {{ root.user }}
+    {% if grains['os'] != 'Windows' %}
+    - group: {{ root.group }}
+    - mode: 755
+    {% endif %}
+    - makedirs: True
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -63,6 +63,7 @@ build_windows_msvc = build_common + Environment({
         r'C:\Program Files\Amazon\cfn-bootstrap',
         r'C:\Program Files\Git\cmd',
         r'C:\Program Files (x86)\WiX Toolset v3.10\bin',
+        r'C:\sccache',
     ]),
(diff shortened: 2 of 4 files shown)


aneeshusa:
r? @larsbergstrom 
cc @luser

servo-build-dependencies/ci.sls
```diff
@@ -0,0 +1,14 @@
+{% from 'common/map.jinja' import root %}
+{% from tpldir ~ '/map.jinja' import sccache %}
+
+sccache:
+  file.managed:
+    - name: {{ sccache.destination }}
+    - source: https://s3.amazonaws.com/rust-lang-ci/rust-ci-mirror/{{ sccache.version }}-sccache-{{ sccache.platform }}
```
larsbergstrom: Just one question - where does this come from? Do we want our own for Servo? cc @edunham

aneeshusa: Official binary releases are [not yet available](https://github.com/servo/servo/issues/17106#issuecomment-306148322) for sccache, so per suggestion I copied these links from [rust-lang/rust](https://github.com/rust-lang/rust) CI scripts.

larsbergstrom: Makes sense - sounds good to me! This is r=me, then. Looks like the travis failure is just atimeout to repo.saltstack.com?

aneeshusa: I wish, there's a real test failure on the cross build due to the Android SDK check I added, this looks like it might require some yak shaving. Investigating.

larsbergstrom:
Does this just need a rebase now that all the Android SDK build stuff has been updated?

aneeshusa:
@bors-servo retry
I've updated this with a workaround for the failing Android SDK test; I think there is a slightly better way possible but it requires a newer Salt and/or a bunch of backporting, so this is simpler.
This should clean up the builds that failed due to force pushing and being unable to find the previous SHA.
Looking at the saltfs build history, we also seem to be hitting https://github.com/travis-ci/travis-ci/issues/8363. I can look for a temporary workaround for that if we don't want to wait.

larsbergstrom:
hrm:
```
$ export GHTOKENLINE=[secure]

The previous command failed, possibly due to a malformed secure environment variable.

Please be sure to escape special characters such as ' ' and '$'.

For more information, see https://docs.travis-ci.com/user/encryption-keys.
```

Does that look familiar?

aneeshusa:
@larsbergstrom that's a separate problem; I thought I had an issue tracking that, but not yet it seems. We're not using this variable (I think it was for a PR we decided not to merge), so it should be unset at some point, but that isn't causing the test failure.

The actual test failure is caused by travis-ci/travis-ci#8363; we want to get Python 3.5 but are getting a weird Python 3.4 instead, causing the test failure. If you want sccache now we can add a workaround, or we can wait for Travis to fix this.

PR #692, opened 2017-06-07
Change key

Manishearth:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/692)


Manishearth:
(Old key had leaked, hence the forced merge. The commit is signed by the GPG key I often use to sign servo and rust commits so you know it's me.)

PR #694, opened 2017-06-07
Enable easy ssh key revocation

### LINKED ISSUE #693: We need a better story for SSH key rotation
Salt only adds keys, and doesn't remove them. We should be able to remove a key from the list here and have it get removed on the server.

I had to rotate a key recently and I had to change salt and also run some sed commands on all machines.

cc @edunham

aneeshusa:
Fully manage the `.ssh/authorized_keys` file for root,
so that keys removed from Salt are also removed from the file.

Note that the AWS (Linux) machines are configured to have two
additional keys in the root `authorized_keys` file,
namely `servo-aws` and `servo-aws-reserved-instances`,
but when used these keys will tell the user to log in as ubuntu,
so they don't provide real access.
Remove these keys as they are not useful.

Add documentation about gaining SSH access and revoking/rotating keys.

Fixes #693.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/694)

### DIFF
--- a/docs/admin.md
+++ b/docs/admin.md
@@ -1,0 +1,53 @@
+# Admin Tasks
+
+## SSH
+
+### Gaining SSH Access
+
+If you need access, create a PR against https://github.com/servo/saltfs/,
+including your account in the `admin/map.jinja` file
+and SSH pubkey in the `admin/ssh` folder.
+
+To access the machines, log in as root on Linux or macOS;
+there are not yet individual accounts on slaves.
+
+If you need to test something (e.g., a reftest failure),
+make sure to su - servo to simulate the space,
+and check the Buildbot config for any required environment variables.
+
+### SSH key revocation and rotation
+
+SSH key rotation can be performed via Salt;
+our Salt configs will both rotate in new keys
+and automatically remove old keys.
+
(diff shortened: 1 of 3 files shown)


aneeshusa:
r? @edunham 

cc @larsbergstrom @Manishearth @metajack 

Obviously, whoever deploys this should be careful and make sure to have an open SSH session (preferably with tmux or screen running) to avoid locking ourselves out.

docs/admin.md
```diff
@@ -0,0 +1,52 @@
...
+and SSH pubkey in the `admin/ssh` folder.
+
+To access the machines, log in as root on Linux or macOS;
+there are not yet individual accounts on slaves.
+
+If you need to test something (e.g., a reftest failure),
+make sure to su - servo to simulate the space,
+and check the Buildbot config for any required environment variables.
+
+### SSH key revocation and rotation
```
aneeshusa: Please take a look at these instructions in particular and let me know if they seem reasonable!

Manishearth:
WFM.

I'm not entirely clear on how to use the the salt testing dir.

aneeshusa:
It's explained in the `salt.md` docs, I added a link inline.

PR #703, opened 2017-06-22
Bump Salt timeout to 60 seconds

aneeshusa:
This provides slow machines (such as some of the Macs)
more time to respond.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/703)

### DIFF
--- a/salt/map.jinja
+++ b/salt/map.jinja
@@ -15,7 +15,8 @@
             'keep_jobs': 0,
             'pillar_roots': {
               'base': [ '/srv/pillar' ]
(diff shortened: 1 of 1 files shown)


edunham:
@bors-servo r+

Sorry for the delay here. 

Looks like I don't have perms to force push to your branch, or I'd've just overwritten it with the commit from this cherry-picked onto the present repo state.

aneeshusa:
Thanks for looking into this originally and reviewing @edunham!

Issue #506, opened 2016-10-06
Need to periodically clean up buildbot slave directories


larsbergstrom:
See:
https://github.com/servo/servo/pull/13597#issuecomment-252014690

PR #710, opened 2017-08-02
Remove all old logs, not just the zipped up ones.

edunham:
Fixes https://github.com/servo/saltfs/issues/506

The only stuff in these directories is the various logs put there by Buildbot, so we're safe to ditch them. 

5 days might be a bit to aggressive but it's what we have before to no major ill effects so I left it. If we want partial logs for a month, it'd be just as easy to run a separate cron job to delete the non-zipped files a bit less enthusiastically.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/710)

### DIFF
--- a/buildbot/master/init.sls
+++ b/buildbot/master/init.sls
@@ -90,7 +90,7 @@ buildbot-github-listener:
 
 remove-old-build-logs:
   cron.present:
-    - name: 'find {{ common.servo_home }}/buildbot/master/*/*.bz2 -mtime +5 -delete'
+    - name: 'find {{ common.servo_home }}/buildbot/master/*/* -mtime +5 -delete'
     - user: root
     - minute: 1
     - hour: 0


edunham:
Travis has failed twice now on assorted intermittents unrelated to this change.

buildbot/master/init.sls
```diff
@@ -90,7 +90,7 @@ buildbot-github-listener:
 
 remove-old-build-logs:
   cron.present:
-    - name: 'find {{ common.servo_home }}/buildbot/master/*/*.bz2 -mtime +5 -delete'
+    - name: 'find {{ common.servo_home }}/buildbot/master/*/* -mtime +5 -delete'
```
aneeshusa: Getting rid of the second `/*` at the end should probably do the trick. For the bz2 cron, we can add a `-name '*.bz2'` as well, and make a separate one for all logs.

Issue #645, opened 2017-04-22
Set UTF-8 locale


aneeshusa:
This is prompted by a warning in the homu log (`You need to set a locale compatible with unicode or homu will choke on Unicode in PR descriptions/titles.`), but in general is good hygiene.

PR #714, opened 2017-08-25
Set utf8 locale

edunham:
Fixes https://github.com/servo/saltfs/issues/645. Setting it on all hosts
because it's good practices to not just assume they have the right default
across diverse systems.

Others have hit quirks like https://github.com/saltstack/salt/issues/24827
with locale.present but the latest comment is saying it's ok now.

literally just https://docs.saltstack.com/en/latest/ref/states/all/salt.states.locale.html except with a tweaked state name

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/714)


edunham:
Restricted to ubuntu only since it works there, definitely doesn't on osx, not sure about windows.

Second commit is to see if it fixes the apparently intermittent failure stemming from missing /usr/share/i18n/locales, seen in https://travis-ci.org/edunham/saltfs/jobs/268434026 but not https://travis-ci.org/edunham/saltfs/jobs/268434017 . Checking on my system reveals i did it wrong; amendment incoming.

Issue #711, opened 2017-08-16
Upgrade linux cross builders to 16GB of RAM


larsbergstrom:
Required to revert the revert here: https://github.com/servo/servo/pull/18114

larsbergstrom:
Given that it was a runaway salt-minion process (1GB of RAM), this may not be required urgently. However, being within 1GB of the top limit seems risky, so probably something to think about in the longer term.

PR #715, opened 2017-09-07
2GB swapfile on Linux slaves

edunham:
to address https://github.com/servo/saltfs/issues/711

from example at
https://serverfault.com/questions/628531/how-to-enable-swap-with-salt-stack

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/715)


aneeshusa:
I would like to incorporate some of the suggestions from that same post about using fallocate instead of dd and maybe handling resizing properly.

PR #717, opened 2017-09-12
Try fixing CI intermittent

edunham:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/717)


aneeshusa:
If we want to work around this, we'll need to pin our Travis CI images to an older version. I believe the inverse of this diff should work: https://github.com/freeipa/freeipa/pull/1056/files

PR #724, opened 2017-09-29
Handle Homebrew renaming `mobile-shell` to `mosh`

aneeshusa:
Homebrew left in an alias for `mobile-shell` so the `mosh` package
actually installs correctly, but Salt's pkg.installed state then tries
to verify the installation by checking the installed packages.
That list only shows `mosh` and not the `mobile-shell` alias,
causing the state the fail.

Update to the new name, which is also consistent with everyone else.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/724)

### DIFF
--- a/admin/init.sls
+++ b/admin/init.sls
@@ -5,11 +5,9 @@ admin-packages:
   pkg.installed:
     - pkgs:
       - tmux
-      {% if grains['os'] != 'MacOS' %}
       - mosh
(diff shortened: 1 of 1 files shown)


aneeshusa:
The upstream change is at https://github.com/Homebrew/homebrew-core/pull/18636.

r? @larsbergstrom or @edunham

PR #725, opened 2017-09-30
Manage sshd_config file

aneeshusa:
This is based on Mozilla's Modern sshd_config from:
https://wiki.mozilla.org/Security/Guidelines/OpenSSH

Note that root login is still allowed,
because we have not yet set up per-user accounts.

Add a test to ensure the sshd_config file is properly parsed and
validated by the OpenSSH version on the machine to help guard
against this behavior.

Additionally, our new 10.11.6 macOS machines have Xcode 8.2 installed.
Use xcode8 on Travis, which maps to Xcode 8gm and macOS 10.11.
This is also important to help ensure tests like the sshd config test
match the behavior we'll see in the field.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/725)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,11 @@ matrix:
       os: osx
       osx_image: xcode8
     - env:
+        - SALT_NODE_ID=servo-mac2
+        - SALT_FROM_SCRATCH=true
+      os: osx
+      osx_image: xcode6.4
+    - env:
         - SALT_NODE_ID=servo-linux1
         - SALT_FROM_SCRATCH=true
       os: linux
@@ -50,6 +55,11 @@ matrix:
         - SALT_FROM_SCRATCH=false
       os: osx
       osx_image: xcode8
+    - env:
+        - SALT_NODE_ID=servo-mac2
+        - SALT_FROM_SCRATCH=false
+      os: osx
+      osx_image: xcode6.4
     - env:
         - SALT_NODE_ID=servo-linux1
         - SALT_FROM_SCRATCH=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ matrix:
         - SALT_NODE_ID=servo-mac1
         - SALT_FROM_SCRATCH=true
       os: osx
-      osx_image: xcode8.3
+      osx_image: xcode8
     - env:
         - SALT_NODE_ID=servo-linux1
         - SALT_FROM_SCRATCH=true
@@ -49,7 +49,7 @@ matrix:
         - SALT_NODE_ID=servo-mac1
         - SALT_FROM_SCRATCH=false
       os: osx
-      osx_image: xcode8.3
+      osx_image: xcode8
     - env:
         - SALT_NODE_ID=servo-linux1
         - SALT_FROM_SCRATCH=false
--- a/.travis/dispatch.sh
+++ b/.travis/dispatch.sh
@@ -141,6 +141,7 @@ else
         ./test.py sls.servo-build-dependencies.android
     fi
 
+    ./test.py sls.admin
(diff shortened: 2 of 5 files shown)


aneeshusa:
This is based on part of #628; @edunham, could you please take a look?

Do not merge this yet - we need to figure out how to bring the versions on `servo-mac1` inline with the other Macs, mostly due to the mismatch in the `sshd_config` location.
IMO, it would be nice to update all the Macs to OS X 10.11 and Xcode 8:
- We can use a modern sshd_config on all machines
- `servo-mac1` has been running this and builds [are green](http://build.servo.org/buildslaves/servo-mac1)
- This matches one of the possible [Travis configurations](https://docs.travis-ci.com/user/reference/osx/#OS-X-Version) for better test/prod parity

aneeshusa:
Also cc @larsbergstrom and @metajack about getting our versions to all match. I don't know how this works with MacStadium:
- Can we pick what version of macOS to have when provisioning a new machine, or is it standard?
- Are we supposed to handle macOS version upgrades, or is that a MacStadium support ticket item?

Another option is using the OpenSSH in Homebrew, but I'm a little wary of messing with the default `ssh.plist` file and getting locked out.

larsbergstrom:
1. We can pick the version of macOS at provisioning time and they will image the machine with it.
2. We can manually upgrade the macOS version upgrades but have to be ready for potential recovery steps - in my experience it will occasionally shut off both ssh and VNC and then you need to open a support ticket.

I'd support a move to 10.11.

metajack:
macOS is at 10.13 now. I think we should move to 10.12 if we are going to upgrade. 10.13 is really new, and some things may be getting ironed out still, but 10.12 should be plenty stable.

aneeshusa:
I'm +1 on standardizing on 10.12 (thanks for the suggestion @metajack), but I'd like to move just one builder to 10.12 first and let it run for a bit to make sure it won't cause any build breakage. It's probably easiest to upgrade `servo-mac1` to 10.12, but we could also add a new `servo-mac10` for a bit more safety.

It would also be nice to standardize the XCode version; I think one of the 8.x versions is probably good for now.

aneeshusa:
With the resolution of #731, 5/9 Mac builders now have macOS 10.11.6, so I think for now it'll be easiest to get all the Macs on 10.11.6 and think about upgrading to 10.12 later. Macs that still need to be upgraded:
- [ ] servo-mac2
- [ ] servo-mac3
- [ ] servo-mac4
- [ ] servo-mac7

I also looked at the `sw_vers` output and it looks like most of the 10.11.6 Macs are on `Build Version 15G1611`. I'm not sure if this matters, but extra consistency here can't hurt.

@larsbergstrom or @edunham Can you put in tickets with MacStadium to upgrade the remaining Macs as well?

aneeshusa:
Updated and simplified to assume all macOS machines are on macOS 10.11.

larsbergstrom:
I think that @edunham has gotten all the machines to 10.11 now,so this should theoretically be good to review and land :-)

aneeshusa:
@jdm I'm still interested in this PR and have rebased it, but we ought to do #811 before merging this, hence the lack of updates so far.

I also chose the Xcode version specifically to be `xcode8` instead of `xcode8.3` per the second commit message, which looks like got lost in between #807/#808. Was there a reason for that? I'd like to match our Travis builders to the actual machines in prod as closely as possible.

aneeshusa:
I just pushed a commit to add more Travis builders for 10.10 (so we test both 10.10 and 10.11 on Travis); if that looks OK we may be able to merge this without waiting for #811.

jdm:
We can revert to xcode8 if it works. I updated to xcode8.3 to make #807 work, but I didn't test anything in between.

PR #726, opened 2017-09-30
Re-enable service.running states in Docker

aneeshusa:
This appears to have started working.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/726)


aneeshusa:
Hmm, this somehow is working for the `sshd` service in #725, so I thought it might be working for other services as well. Seems that's not the case.

Issue #593, opened 2017-02-02
Make `pip.installed` actually upgrade Homu


aneeshusa:
It seems like the `pip.installed` state currently is failing to update Homu to newer versions when the git revision changes, requiring manual upgrades.

Temporary manual upgrade procedure:
- Stop Homu with `service homu stop`
- Back up the Homu database (currently located at `/home/servo/homu/main.db`)
- Blow away the Homu dir (`/home/servo/homu`)
- Run the Salt highstate to install the upgraded Homu
- Replace the Homu database

jdm:
This also affects the version of https://github.com/servo/intermittent-tracker/ that's in use; it seems like the only steps necessary there would be:
* Blow away the Homu dir (/home/servo/intermittent-tracker)
* Run the Salt highstate to install the upgraded intermittent-tracker

aneeshusa:
I haven't checked, but I think #623 may have fixed this.

jdm:
Pip installs fine from git sources when using the `--upgrade` flag. Adding the equivalent flag in salt makes this work.

jdm:
#623 definitely should have fixed this.

PR #742, opened 2017-10-25
Fix failure tracker

jdm:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/742)

### DIFF
--- a/intermittent-failure-tracker/init.sls
+++ b/intermittent-failure-tracker/init.sls
@@ -9,6 +9,10 @@ intermittent-failure-tracker:
     - venv_bin: virtualenv-3.5
     - python: python3
     - system_site_packages: False
+    - pip_pkgs:
+      # This package is required to install the non-code resources that are present
+      # in the tracker package's git repository.
+      - setuptools-git
     - require:
       - pkg: python3
       - pip: virtualenv
@@ -16,6 +20,7 @@ intermittent-failure-tracker:
     - pkgs:
       - git+https://github.com/servo/intermittent-failure-tracker@{{ tracker.rev }}
     - bin_env: /home/servo/intermittent-failure-tracker/_venv
+    - upgrade: True
     - require:
       - virtualenv: intermittent-failure-tracker
   {% if grains.get('virtual_subtype', '') != 'Docker' %}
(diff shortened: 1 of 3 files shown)


jdm:
This fixes a long-time issue with the intermittent-tracker where we never actually could update the installed package on the server. This also makes the new intermittent failure tracker server start up successfully; I hadn't tried it in python3 before.

intermittent-failure-tracker/init.sls
```diff
@@ -9,13 +9,16 @@ intermittent-failure-tracker:
     - venv_bin: virtualenv-3.5
     - python: python3
     - system_site_packages: False
+    - pip_pkgs:
+      - setuptools-git
```
aneeshusa: Can you add a comment about why this is needed?

intermittent-failure-tracker/map.jinja
```diff
@@ -1,5 +1,5 @@
 {%
   set tracker = {
-    'rev': '37ef73bb0059e0555a903e758dae8006c46a9916'
+    'rev': 'f2f3825b6b026b54769a995ae8a526b2d81ee5fd'
```
aneeshusa: It would be nice to publish all these packages to PyPI as well, as our Salt version does better with that than git installs.

intermittent-failure-tracker/init.sls
```diff
@@ -37,6 +40,12 @@ intermittent-failure-tracker:
     - group: servo
     - mode: 644
 
+/home/servo/intermittent-failure-tracker/intermittent_errors.json:
```
aneeshusa: Why do we need to create this file ourselves? Seems like something the package should create on-demand.

jdm: The current working directory is not owned by the user that is running the server, so it can't.

intermittent-failure-tracker/map.jinja
jdm: Better in what sense?

aneeshusa: https://github.com/servo/saltfs/issues/593 is what I'm aware of; this may have been fixed or improved since then but at least in our old Salt the git installs seem less well supported.

jdm: Per my latest comment in that issue, a2934d015b300086b1f6194f5b69d785ebaf4df0 is all that is necessary. I would rather not go through the effort of publishing these packages.

aneeshusa: Thanks for updating that issue, since it seems to be working then git dependencies are fine by me.

PR #756, opened 2017-11-22
Install aws-cli on linux builders.

jdm:
@asajeffrey says that this would be useful for the linux-nightly builder.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/756)


asajeffrey:
Indeed it would be! @bors-servo r+

PR #761, opened 2017-12-08
Make rustup update when specific version not present.

jdm:
This PR causes us to update to the most recent release of rustup, as a result.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/761)

### DIFF
--- a/servo-build-dependencies/ci-map.jinja
+++ b/servo-build-dependencies/ci-map.jinja
@@ -1,4 +1,10 @@
 {%
+  set rustup = {
+    'version': 'rustup 1.7.0 (813f7b7a8 2017-10-30)'
+  }
+%}
+
+{%
   set sccache = salt['grains.filter_by']({
       'defaults': {
         'destination': '/usr/local/bin/sccache',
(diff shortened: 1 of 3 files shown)


jdm:
r? @metajack 
cc @SimonSapin

aneeshusa:
This current solution will cause new machines to fail to bootstrap if rustup publishes a newer version than 1.7.0.
I would want to have versioned-locked download URLs for reproducibility.
The long term goal here is versioned uploads of rustup itself, published by the rustup.rs repo (either S3 or GH releases).
In the meantime, we can manually mirror selected versions to a servo-controlled S3 bucket (we have a couple of them).
I would prefer to download rustup itself directly if it is a static binary on all platforms (similar to sccache), but if that is not possible it would also be acceptable to have versioned URLs for the rustup installer.

SimonSapin:
I don’t think we should be in the business of hosting rustup binaries. We can parse the output of `rustup --version` to extract a tuple of integers, and check for a minimum version instead of an exact one.

SimonSapin:
Something like…

```diff
-    - unless: rustup --version | grep '{{ rustup.version }}'
+    - unless: rustup --version |  python -c 'import sys, re; [int(i) for i in re.match("rustup (\d+).(\d+).(\d+) ", sys.stdin.read()).groups()] >= [1, 7, 0] or sys.exit(1)'
```

Issue #755, opened 2017-11-15
Upgrade sccache


jdm:
This will necessitate running `cargo install --force sccache`, I believe.

jdm:
Nevermind, we download it directly: https://github.com/servo/saltfs/blob/e4a1d01330a5ecc1b41423e1a4e27560f9cca1d3/servo-build-dependencies/ci-map.jinja#L2

jdm:
This will require using the most recent version of sccache, and updating the download hashes in that file to match the hashes for each of the downloads.

jdm:
The other versions listed at https://crates.io/crates/sccache do not exist on rust-lang-ci/rust-ci-mirror.

jdm:
@acrichto @luser What is required to get a new version of sccache uploaded on https://s3.amazonaws.com/rust-lang-ci/rust-ci-mirror/?

alexcrichton:
@jdm right now that's just binaries used for Rust's own CI and we haven't needed to update it for quite some time so we haven't, otherwise sccache should [have support now](https://github.com/mozilla/sccache/blob/master/.travis.yml#L49-L59) for binary releases to github which can be mirrored to a custom bucket.

PR #763, opened 2017-12-14
Make sccache never time out

jdm:
This should fix https://github.com/servo/servo/issues/19495.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/763)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -51,6 +51,7 @@ doc = Environment({
 build_common = Environment({
     'RUST_BACKTRACE': '1',
     'BUILD_MACHINE': str(util.Property('slavename')),
+    'SCCACHE_IDLE_TIMEOUT': '0',
 })
 
 build_windows_msvc = build_common + Environment({


aneeshusa:
Are there docs for this somewhere? Please add a link to the commit message and also maybe a description of the problem (where is sccache timing out currently?) and/or link to a Buildbot log in this PR.

jdm:
Commit message updated.

aneeshusa:
Thanks for the links. Two questions I have:
- The `Jenkins.md` link recommends `start the server beside Jenkins as a system service` in addition to setting the environment variable in this PR. Do you know if this is required/what the effect is if we don't start sccache by ourselves as a system service (e.g. upstart or systemd)?
- We will need to upgrade sccache from time to time (e.g. #755). What is the story for handling these kinds of upgrades: do we need to perform manual steps, and if so, what are they? (Stopping the existing sccache process(es), starting new one(s), clearing cache data from disk, etc.)

jdm:
The first time a build is run with a timeout of 0, it will automatically start the long-lived server in the background. As part of upgrading, we should run `sccache --stop-server` before installing the new version. Otherwise we should not need to do anything else. This PR should be mergeable as-is.

jdm:
I'm going to go ahead and merge this change because https://github.com/servo/servo/issues/19495 is really frustrating and I want to have enough time to be able to look into anything that might go wrong.

@bors-servo: r+

PR #768, opened 2017-12-19
Add taskcluster yml

aneeshusa:
We need newer Docker images to be able to use TC for servo (see https://github.com/servo/servo/pull/19591). Use TC to build these images from the saltfs repository.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/768)


aneeshusa:
r? @edunham 
We also need to enable the TaskCluster webhook on this repository.
Once we have images building, we'll need to set up scopes so that TC can push the images it builds to a registry somewhere so we can use them.

edunham:
Waiting till I know enough to be certain this is right has hung it up for embarrassingly long. Let's merge and fix things if it breaks them. 

@bors-servo r+

Issue #327, opened 2016-04-19
Make both the Linux and Darwin `servo_home` variables available from Jinja


aneeshusa:
Opening this so I don't forget about this; see #324 for details.

edunham:
See also #209

PR #772, opened 2018-01-08
{{ common.servo_home }} seems to be wrong on Mac

SimonSapin:
https://github.com/servo/servo/pull/19395#issuecomment-356103258

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/772)


aneeshusa:
@SimonSapin this is because the `common.servo_home` variable will always reflect the value for the current OS; because the Buildbot master runs on Linux, it will always return the Linux value (Salt doesn't know that you're trying to write a build config for macOS).

#327 tracks exposing all the different possible values so they can be referred to explicitly in the Buildbot configurations.

Issue #496, opened 2016-09-29
Make each sls file individually `state.apply`able


aneeshusa:
This mainly involves making sure each sls file `include`s any other sls files it depends on, i.e. that contain states it references. I had been unsure whether this was necessary, but this now seems like a good idea to be able to run "partial" highstates in the case that some modules are broken, as seems to be the case with our OS X builders at the moment. (Everything works except a full state.highstate; being able to state.apply would let us roll out SSH keys, at the least.)

aneeshusa:
Some progress is made in #577, enabling at least `servo-build-dependencies` to run standalone.

Manishearth:
More progress in https://github.com/servo/saltfs/pull/883 adding common deps where needed

PR #782, opened 2018-01-25
Import module for missing user requirement.

jdm:
This fixes a longstanding problem when highstating on servo-mac4, since the user requirement is defined in common/init.sls.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/782)

### DIFF
--- a/buildbot/slave/init.sls
+++ b/buildbot/slave/init.sls
@@ -1,6 +1,7 @@
 {% from 'common/map.jinja' import common %}
 
 include:
+  - common
   - python
 
 buildbot-slave-dependencies:


jdm:
r? @aneeshusa 
I promise this is a tiny one.

aneeshusa:
@bors-servo r+
This shouldn't be required for full highstates since we always include the `common` SLS on not-Windows: https://github.com/servo/saltfs/blob/f50214b8fa012e03616ecae1ef2913e6fe9044da/top.sls#L4-L7.
However, this is still an improvement in the case that someone uses `state.sls` to selectively run the `buildbot.slave` SLS, so it's good to have (helps with #496).

aneeshusa:
Seemed to failed around starting the buildbot-slave service on OS X, so possibly related, but I think this is a Travis CI flake.
@bors-servo retry

PR #783, opened 2018-01-26
Provide credentials to nightly WPT sync job

jdm:
This will give us more useful information by actually creating a PR with the results of syncing with upstream. We can choose to disregard it without any issue.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/783)

### DIFF
--- a/buildbot/master/files/config/environments.py
+++ b/buildbot/master/files/config/environments.py
@@ -1,6 +1,7 @@
 from buildbot.plugins import util
 from passwords import GITHUB_DOC_TOKEN, GITHUB_HOMEBREW_TOKEN
 from passwords import S3_UPLOAD_ACCESS_KEY_ID, S3_UPLOAD_SECRET_ACCESS_KEY
+from passwords import WPT_SYNC_PR_CREATION_TOKEN
 
 
 class Environment(dict):
@@ -152,4 +153,8 @@ upload_nightly = Environment({
     'AWS_ACCESS_KEY_ID': S3_UPLOAD_ACCESS_KEY_ID,
     'AWS_SECRET_ACCESS_KEY': S3_UPLOAD_SECRET_ACCESS_KEY,
     'GITHUB_HOMEBREW_TOKEN': GITHUB_HOMEBREW_TOKEN,
+})
+
+sync_wpt = Environment({
+    'WPT_SYNC_TOKEN': WPT_SYNC_PR_CREATION_TOKEN,
 })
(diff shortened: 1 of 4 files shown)


buildbot/master/files/config/environments.py
```diff
@@ -152,4 +153,5 @@ def without(self, to_unset):
     'AWS_ACCESS_KEY_ID': S3_UPLOAD_ACCESS_KEY_ID,
     'AWS_SECRET_ACCESS_KEY': S3_UPLOAD_SECRET_ACCESS_KEY,
     'GITHUB_HOMEBREW_TOKEN': GITHUB_HOMEBREW_TOKEN,
+    'WPT_SYNC_TOKEN': WPT_SYNC_PR_CREATION_TOKEN,
```
aneeshusa: Why are we putting this in the `upload_nightly` environment? It sounds like this is being run as a separate step, so we should use a separate environment (e.g. I find it unlikely this needs AWS access), and use the environment combining feature with `+` if necessary.

buildbot/master/init.sls
```diff
@@ -38,6 +38,7 @@ deploy-{{ common.servo_home }}/buildbot/master:
     - context:
         common: {{ common }}
         buildbot_credentials: {{ pillar['buildbot']['credentials'] }}
+        wpt_sync_credentials: {{ pillar['wpt-sync'] }}
```
aneeshusa: I'd prefer if these credentials got moved under `buildbot:credentials` so this line is not needed. The test credential would be removed from the `.travis/test_pillars/wpt-sync.sls` file and moved into the `.travis/test_pillars/buildbot/master.sls`.

jdm: It's also used by https://github.com/servo/saltfs/blob/master/upstream-wpt-webhook/files/config.json#L2 so it won't actually be removed.

aneeshusa: I didn't realize that, keeping the current setup makes sense to me then.

buildbot/master/files/config/environments.py
jdm: Fixed.

aneeshusa:
Thanks!
@bors-servo r+

aneeshusa:
Looks like there is a lint error:
```
[ FAIL ] Trailing whitespace found on files and lines:
         ./buildbot/master/files/config/environments.py:156:    'WPT_SYNC_TOKEN': WPT_SYNC_PR_CREATION_TOKEN,----
```

PR #786, opened 2018-01-28
Add Salt testing root setup/cleanup scripts

aneeshusa:
I found versions of these on `servo-master1`
and thought they were a great idea,
so I've pulled them into Salt so we always have them.

I've also update the documentation,
and added a safety around a potential footgun inside Vagrant.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/786)


aneeshusa:
@bors-servo retry

- Looks like spurious/flaky package fetching issues.

PR #787, opened 2018-01-31
Share linux build env with doc build.

jdm:
This should fix https://github.com/servo/servo/issues/19844.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/787)


SimonSapin:
Filed https://github.com/servo/saltfs/issues/789 which doesn’t need to block this.

@bors-servo r+

Issue #657, opened 2017-05-05
Use Salt publisher ACL system to allow Salt usage without root access


aneeshusa:
Salt allows somewhat pluggable authentication systems. One of these is their publisher ACL system (docs: https://docs.saltstack.com/en/2016.3/ref/publisheracl.html), which allows local users to execute Salt commands without having root access. I think implementing this would help us lock down our systems - I think we could likely give out Salt-only access instead of root access via SSH (should help with https://github.com/servo/saltfs/pull/628).

It seems to allow restricting which Salt functions are available, which is also nice.
I haven't used this myself yet, so likely would like to play around with it first.

cc @edunham @larsbergstrom

larsbergstrom:
Oh, this is very interesting! I think that there are a few things that you could easily have people do (mainly pushing out updates, paving the "build" directory, and rebooting the machine) that would get us like 99% of the admin stuff that we need. Awesome!

Maybe restarting buildbot/homu, too :-)

larsbergstrom:
cc @metajack

aneeshusa:
Rebooting is `system.reboot`, and I can write a custom module function to blow away build dirs pretty quickly. What did you mean by "pushing out updates" - running `state.highstate`?

Restarting Homu should just be `service.restart` - I don't remember any other special steps needed.
Restarting Buildbot can be wrapped up into a custom module function.

metajack:
+1 from me

aneeshusa:
I tried this on my own infrastructure, and I don't think Salt's `publisher_acl` will be useful for us right now; restricting the commands users can run via the `sudoers` file is likely more fruitful for right now.

Details:
I thought the salt binary would simply make an RPC to a running salt-master daemon process, which would handle user authentication (e.g. `publisher_acl` via Unix socket permissions). However, it looks like the salt command actually does most of the work (logging, publishing a command to minions, etc.) itself, and only listens to the running daemon for responses as they come in from minions. This means that our user accounts still need access to various Salt bits like the config, the pillar, the log, etc. (confirmed in [the docs](https://docs.saltstack.com/en/2016.3/ref/publisheracl.html#permission-issues)), at which point they might as well have root access anyways via sudo.
I do want to extend Salt to make it do the kind of authentication I was expecting, but that might take a while, and we'll need to upgrade our Salt in order to use it anyways.

PR #791, opened 2018-02-05
Run IRC bot for salt interactions

jdm:
Example of the single command that `saltbot` understands right now:
```
14:22 <jdm> disk usage
14:22 <saltbot> servo-linux1: 68%
14:22 <saltbot> servo-linux5: 47%
14:22 <saltbot> servo-linux4: 70%
14:22 <saltbot> servo-linux6: 37%
14:22 <saltbot> servo-linux3: 68%
14:22 <saltbot> servo-mac2: 75%
14:22 <saltbot> servo-mac3: 35%
14:22 <saltbot> servo-mac8: 31%
14:22 <saltbot> servo-mac6: 32%
14:22 <saltbot> servo-mac7: 9%
14:22 <saltbot> servo-mac4: 37%
14:22 <saltbot> servo-mac5: 32%
14:22 <saltbot> servo-mac1: 16%
14:22 <saltbot> servo-linux2: 63%
```
Since salt commands can only be run as root (or they can't write to logs in /var/logs), I couldn't figure out how to run this bot as non-root. The code for the bot is at https://github.com/jdm/saltbot. @aneeshusa, what are your thoughts on this?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/791)


aneeshusa:
@jdm, thanks for working on this! I'm always glad to see progress on [chatops for servo](https://github.com/servo/servo/wiki/Student-projects).
A couple notes:
- Instead of using `cmd.run 'df -h'`, use Salt's [built in `disk.usage` execution function](https://docs.saltstack.com/en/2016.3/ref/modules/all/salt.modules.disk.html#salt.modules.disk.usage), which will give you structured data, and should work on Windows too. Might be nice to also have the IRC interface reflect this and be `disk.usage` instead of `disk usage` (or accept both possibly).
- It looks like the current code is in Javascript; I'd prefer to use Salt's [Python client API](https://docs.saltstack.com/en/2016.3/ref/clients/index.html#client-apis), which gives you structured data and is a bit easier to use IMO. In my experience trying to run NodeJS in production causes operational sadness. Otherwise, you'll probably be interested in using the JSON outputter, which will cause Salt to output (machine-readable) JSON.
- Thanks for trying to run this as non-root; I looked into this previously as part of #657 and there's currently not a great way to do this. A few options:
  - I'm OK with running this as root, but we'll need to be careful about what access we provide via IRC.
  - Another (more involved) option is setting up their [CherryPy REST API](https://docs.saltstack.com/en/2016.3/ref/netapi/all/salt.netapi.rest_cherrypy.html), and then hitting it directly or using the [Pepper Python Client](https://github.com/saltstack/pepper) to use that API. This uses Salt's pluggable [External Auth](https://docs.saltstack.com/en/2016.3/topics/eauth/index.html#acl-eauth) system, which will allow us to create a Salt user just for the IRC bot and restrict which functions and arguments it can use. If we do this, we can run just the CherryPy API on the Salt master, set up TLS, and then run the IRC bot on a separate machine for added safety.
- If you decide to keep invoking Salt via the CLI binary, then you should run one Salt command with a target of all builders instead of invoking Salt once for each builder; Salt should stream output for you as results come back. `salt-key -L` is useful to figure out the set of current minions, instead of hardcoding the number of Linux/macOS minions.
- Some form of rate-limiting would be good in the bot to avoid being DOSed. Similarly, we should probably check that incoming messages are to the bot.
- I can see this being extended to not-fully-Salt related things (e.g. which saltfs PRs are outstanding/yet-to-be-deployed? This actually is a GitHub call), so maybe pick a different name while it's easy.

Hooking up IRC to Salt is a bit scary, so I'll try to take a deeper look later on.

PR #795, opened 2018-02-07
Add my SSH key

jonleighton:
In https://github.com/servo/servo/pull/19928 @metajack suggested that I could use one of the builders to work on Mac-specific code and @jdm requested I submit my key in a PR here.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/795)


jdm:
Oops, I just noticed that we need an entry in https://github.com/servo/saltfs/blob/master/admin/map.jinja as well per https://github.com/servo/saltfs/blob/18dd9b1d66f7f1b74e367c4bd297041c48ee9e42/docs/admin.md#L5.

PR #798, opened 2018-02-09
Note down who created tmp saltfs clone and when

aneeshusa:
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/798)


jdm:
@bors-servo r+
Nice solution :)

PR #802, opened 2018-02-12
Add paulrouget to the list of reviewers

paulrouget:
This is mostly to be able to do @bors-servo r=XXX. Often I get a simple r=me, without the @bors-servo snippet.

Also, happy to review anything about embedding and ports.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/802)


paulrouget:
Thank you.

PR #818, opened 2018-03-14
Add HTTPS for build.servo.org

jdm:
This is yolo-deployed so it can be observed to work: http://build.servo.org/grid should load https://build.servo.org/grid.


jdm:
I originally had an HTTPS redirect for any HTTP URL, but that broke our tools like `./mach filter-intermittent` that became confused by a 307 response. We'll need to do this in two stages - first, turn on HTTPS. Once we've migrated everything to use HTTPS URLs, we can add the HTTP->HTTPS redirect. This should be ready to merge and deploy in its current state.

nginx/init.sls
```diff
@@ -19,3 +19,20 @@ nginx:
...
+  pkg.installed:
+    - pkgs:
+      - certbot
+      - python-certbot-nginx
+
+certbot renew:
+  cron.present:
+    - identifier: build-cert-renew
+    - user: root
+    - minute: 0
```
aneeshusa: In the interest of being a good internet citizen, let's not use midnight directly but a more "random" time as the base - I'm sure they already get way too many requests right at midnight. Additionally, we can use `sleep` to add some more jitter: https://stackoverflow.com/questions/9049460/cron-jobs-and-random-times-within-given-hours/16289693#16289693 for example.

nginx/init.sls
```diff
@@ -19,3 +19,20 @@ nginx:
...
+    - hour: 0
+    - daymonth: 1
```
aneeshusa: Instead of running once a month, it would be nice to run once a week (ish) so we can catch problems faster. Also, we should figure out a story for getting notifications of failed jobs.

jdm:
@aneeshusa https://github.com/servo/servo/pull/20267#issuecomment-373164398

jdm:
I've addressed both comments. My best idea for notifying when the job fails is to file a github issue in this repository.

edunham:
Looks good to me. I agree that failures should report to somewhere, perhaps a GitHub issue, but I would rather not block landing this on setting that up. 

@bors-servo r+

jdm:
I mean, the changes in this PR include filing a github issue when there's a failure, so... :)

PR #821, opened 2018-03-18
Make list of repos more configurable

### LINKED ISSUE #803: Remove travis gating from webrender repo.
Removing travis integration altogether would be fine, if that won't break any of the current setup (additional reviewers etc).

jdm:
This fixes #803 and makes it easier for other repositories to be ungated and enable taskcluster support in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/821)

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -110,8 +110,11 @@ secret = "{{ secrets['web-secret'] }}"
         "extra_reviewers": [ "nical", "jrmuizel", "Gankro" ],
         "taskcluster": True,
         "appveyor": True,
+        "travis": False,
     },
-    "webrender_traits": {},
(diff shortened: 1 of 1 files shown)


homu/files/cfg.toml
```diff
@@ -94,8 +108,13 @@ secret = "{{ secrets['web-secret'] }}"
     "unicode-script": {},
     "webrender": {
         "extra_reviewers": [ "nical", "jrmuizel", "Gankro" ],
+        "taskcluster": True,
+        "travis": False,
+        "webrender": True,
```
aneeshusa: I think this is meant to be `"appveyor": True,`?

jdm: Oops. Awkward.

PR #832, opened 2018-05-07
Add gw3583 to reviewers list.

gw3583:


gw3583:
Through an unfortunate sequence of events, I'm currently unable to log in to my normal @glennw account (for at least the next two weeks). So, I'm adding a new account to use in the interim.

r? @edunham

edunham:
I'm verbally confirming with the real-world instance of Glenn that this change is correct before Salting it out :)

PR #836, opened 2018-05-08
Replace eijebong's key with a smaller one

Eijebong:
Apparently the builders don't like 16k keys

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/836)


edunham:
@bors-servo r+

Thanks!

PR #837, opened 2018-05-11
Don't ignore existing pip packages on linux

jdm:
When we constantly reinstall pip packages on ubuntu, we run into https://github.com/servo/servo/issues/20774#issuecomment-388512785 which breaks every linux build machine.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/837)


jdm:
@bors-servo r-
This shouldn't be necessary.

PR #841, opened 2018-05-24
Require v4.0 of LLVM and friends

asajeffrey:
Bindgen needs LLVM 3.9 or higher.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/841)


aneeshusa:
Besides applying to a builder that's taken out of rotation, another way to test this might be to try to build a docker image (see docs/docker.md) and build servo in there; my memory is a bit fuzzy but IIRC the docker images can build servo, but can't quite run some of the tests (e.g. reftests that need xvfb).

asajeffrey:
I applied these changes, and ran vagrant locally. With these changes `./mach build -d` succeeds, both with servo master, and with bindgenup.

PR #849, opened 2018-07-05
Run TC more often

edunham:
* Switch to servo-docker-worker to match Servo repo and facilitate manual rebuilds on taskcluster for debugging when desired
* Build pull requests as well because I'm using the Salt repo to verify TC changes and also because more testing > less testing

r? @aneeshusa

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/849)

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -1,12 +1,15 @@
 version: 0
 tasks:
   - provisionerId: '{{ taskcluster.docker.provisionerId }}'
-    workerType: '{{ taskcluster.docker.workerType }}'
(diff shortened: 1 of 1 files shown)


aneeshusa:
@edunham This is r=me, thanks!
For my own edification, can you explain what setting the `workerType` to `servo-docker-worker`does?
Building on PRs seems like a good idea, and should be safe since [TC defaults to only building PRs for project collaborators](https://docs.taskcluster.net/docs/reference/integrations/taskcluster-github/docs/taskcluster-yml-v0#who-can-trigger-jobs-).

PR #850, opened 2018-07-08
Add linux-kvm builder type

SimonSapin:
CC https://github.com/servo/servo/issues/21124

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/850)

PR #852, opened 2018-07-11
Add the android-x86 build job 

SimonSapin:
CC https://github.com/servo/servo/issues/21124

Should not be deployed before https://github.com/servo/servo/pull/21162 lands, as that would likely break all builds.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/852)


SimonSapin:
Compared to https://github.com/servo/saltfs/pull/850#issuecomment-403890493, running `salt-run fileserver.clear_cache` helped `ID: /home/servo/buildbot/slave Function: file.recurse` succeed, but the other task still failed on the new server

```
          ID: buildbot-slave
    Function: service.running
      Result: False
     Comment: Failed to start the service
     Started: 07:02:11.274890
    Duration: 57.718 ms
     Changes:   
              ----------
              buildbot-slave:
                  False

SimonSapin:
`/etc/init/buildbot-slave.conf` exists on `servo-linux-kvm1`, but `service buildbot-slave start` says: `Failed to start buildbot-slave.service: Unit buildbot-slave.service not found.` Maybe this is because Ubuntu switched to systemd between 14.04 and 16.04?

PR #855, opened 2018-07-16
Install git inside Taskcluster

aneeshusa:
We need this for bootstrapping.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/855)


aneeshusa:
Haven't tested so not sure this works :)
Based on https://tools.taskcluster.net/groups/dHqfk75ETaOBJIA_fOoDww/tasks/dHqfk75ETaOBJIA_fOoDww/details.
r? @edunham

PR #856, opened 2018-07-16
SDK update

paulrouget:
This works locally, but I haven't tested extensively. `vagrand provision` runs without any error.

This is necessary for: https://github.com/servo/servo/pull/20912

Any recommendation for how to land this?

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/856)

### DIFF
--- a/servo-build-dependencies/android.sls
+++ b/servo-build-dependencies/android.sls
@@ -48,3 +48,9 @@ android-dependencies:
     - refresh: True
     - require:
       - pkgrepo: openjdk
+
+java-8-alternative:
+  cmd.run:
+    - name: update-java-alternatives -s java-1.8.0-openjdk-amd64
+    - require:
+      - pkg: android-dependencies


paulrouget:
Might be this: https://github.com/saltstack/salt/issues/22356

paulrouget:
We need Java 8 for the sdkmanager to work. By default, openjdk 7 is present. We install openjdk 8.

2 options: remove openjdk7 or set openjdk8 as the first alternative (via `update-alternatives`).

paulrouget:
r? @SimonSapin 

Any recommendation on how to land this?

Is it ok to land this PR then not land anything until https://github.com/servo/servo/pull/20912 lands (ready to land, just need the saltfs update)

SimonSapin:
Would PRs start failing if we deploy this without landing https://github.com/servo/servo/pull/20912 ?

I think we have a mechanism to actually close the tree until a given PR lands, but if it’s ready it’s probably fine to just r+ it when appropriate, possibly with a priority. There isn’t a lot of PR traffic at the moment.

By the way, maybe https://github.com/servo/servo/pull/21185 makes this unnecessary? But no need to block if this is ready, let’s do whichever you think is easiest.

paulrouget:
> Would PRs start failing

I haven't tested master with the new build tools and the new SDK. I'd be surprised if it works. 

>  it’s probably fine to just r+ it when appropriate, possibly with a priority

Ok.

> By the way, maybe servo/servo#21185 makes this unnecessary?

Using the new bootstrap mechanism is great, but not enough. We also need to update the JRE alternative.

paulrouget:
```
==> Installing saltstack
==> Downloading https://homebrew.bintray.com/bottles/saltstack-2016.3.3.sierra.b
==> Pouring saltstack-2016.3.3.sierra.bottle.tar.gz
Error: undefined method `undent' for #<String:0x0000000103a09750>
```

SimonSapin:
It looks like we’re installing an old version of the formula: https://github.com/servo/saltfs/blob/5f6a0f0a8f3adf1de1917811e517462952cd460d/.travis/install_salt.sh#L50

paulrouget:
We're trying to install our own version of the formula. See https://github.com/servo/saltfs/pull/859

jdm:
@paulrouget Want to rebase this and remove everything except the java update?

PR #863, opened 2018-07-23
Bump Salt to 2018.3.2

Manishearth:
Homebrew is no longer compatible with the old version of salt which we
pin, update to a new one.

Previously: https://github.com/servo/saltfs/commit/87d97417bf54ed6745867e1711d41c83e0ab4c91

saltfs-migration: Upgrade Salt on all machines. Steps:
- Stop all salt-master and salt-minion services. Because we're upgrading
  Salt by two major versions at once, there are no guarantees that the
  masters and minions with different versions will be able to
  communicate.
- Use the install script to re-install Salt on each machine, the master
  first then the minion.
- Restart the salt-master and salt-minion services; pay attention to the
  system logs in case of startup failure.
- Do NOT run a highstate until all minions and masters have been updated
  to Salt 2018.3.2

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/863)


Manishearth:
It's unable to install ttf-mscorefonts-installer , hmm

Manishearth:
So the core problem here is that in late 2016 Salt's archive.extracted started verifying that files got extracted

Which causes two problems:

 - It takes _forever_ to do this for the bajillion files in the NDK/SDK archives (solution: we actually don't need salt to do this anyway)
 - For some reason it can't find libglut.so.3 in the https://servo-rust.s3.amazonaws.com/ARM/armhf-trusty-libs/v2/armhf-trusty-libs-v2.tgz archive. That file is present in the archive, but when extracted on the server libglut.so.3 is owned by `servo` (and all other files by `root`) and I don't know why (this may be a symptom of not being able to find the file, not a cause?)

Manishearth:
ohhh we use a forked archive state, which causes some diversion here

see 11d4ac13e417438c59597bb5c6202975c8d7809c

unforking since this has probably been fixed and either way we don't fetch zips anymore (?)

Manishearth:
Okay, this is ready. r? @jdm @edunham @aneeshusa 

Should I deploy this upgrade first before landing so that we know it works?

.travis/install_salt.sh
```diff
@@ -44,12 +44,12 @@ install_salt() {
         printf "\n"
         # Unlink allows switching versions,
         # I wish Homebrew had an atomic operation for pinned upgrades
-        if brew list | grep 'saltstack' >/dev/null; then
+        if brew list | grep 'salt' >/dev/null; then
             brew unlink saltstack
         fi
-        brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/9e3a66b6b7ca978bfea86897dcc3391c37f9f0ef/Formula/saltstack.rb
+        brew install --force https://raw.githubusercontent.com/Homebrew/homebrew-core/master/Formula/salt.rb
```
aneeshusa: Please pin to a specific version, otherwise we'll end with up macOS builders all having different versions depending on what's in master when they come up.

aneeshusa:
Looks good other than that, good luck :)

.travis/install_salt.sh
Manishearth: oh oops I _was_ doing this but I probably copied the wrong URL at the end

Manishearth:
We still need the android reference, we just don't need the NDK bits.

Manishearth:
@bors-servo r=jdm,aneeshusa

deployed, except on servo-mac 1-3 and servo-linux3 (these are down for now, I'll upgrade them later)

PR #874, opened 2018-08-09
Remove require: user: servo

Manishearth:
This only works if we have a  `servo: user.present` state somewhere; but
we don't, so we get the following error:

```
     Comment: The following requisites were not found:
                                 require:
                                     user: servo
```

We assume the user exists elsewhere anyway, though perhaps we could add
the user.present in case (but when i tried i wasn't able to make that
work)

r? @aneeshusa

cc @jdm

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/874)

### DIFF
--- a/buildbot/master/init.sls
+++ b/buildbot/master/init.sls
@@ -21,7 +21,6 @@ buildbot-master:
     # Buildbot must be restarted manually! See 'Buildbot administration' on the
     # wiki and https://github.com/servo/saltfs/issues/304.
     - require:
-      - user: servo
       - pip: buildbot-master
       - file: ownership-{{ common.servo_home }}/buildbot/master
       - file: /etc/init/buildbot-master.conf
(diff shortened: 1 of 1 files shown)


Manishearth:
Any highstate on servo-master.


```
require:
  user: servo
```

does not require there to be a user named `servo`, it requires there to be a _state_ named `servo` with a `user.present` in it.

We depend on a servo user existing all over the place without doing this.

We could add a servo state somewhere if necessary.

aneeshusa:
I took a look and the issue is that none of the sls files that the master is using which depend on the servo user are pulling in the common.sls file. The better fix here is to make progress on #496 and make sure the states using the servo user - either for ownership or its home directory - a) have an include for `common` ([example](https://github.com/servo/saltfs/pull/782)) and b) ideally the relevant states depend on the state that sets up the user ([example](https://github.com/servo/saltfs/blob/d37e2c677a4448333ed611432e2051d40dbabe43/buildbot/slave/init.sls#L32-L33)).

Manishearth:
Ah, that explains why it's somewhat intermittent, it depends on the order.

Opened https://github.com/servo/saltfs/pull/883 to fix this.

Issue #423, opened 2016-07-01
Install Letsencrypt and automate certificate renewal/regeneration


edunham:
We want to turn on mandatory HTTPS  on servo.org with HSTS, see https://github.com/servo/download.servo.org/issues/1. 

To do this, all servo.org subdomains must have valid TLS certs. 

We're mostly ready, save for 2 things: 
- build.servo.org needs a cert, see https://github.com/servo/saltfs/issues/196. 
- In 90 days, the cert in use by CloudFront for downloads will expire and need to be renewed. 
- [ ] Salt state for  `pip install letsencrypt letsencrypt-s3front`, on buildmaster
- [ ] every 60 days, run the following in an environment with `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` set from the pillar: 

```
letsencrypt --agree-tos -a letsencrypt-s3front:auth \
--letsencrypt-s3front:auth-s3-bucket servo-builds \
--letsencrypt-s3front:auth-s3-region us-west-2 \
-i letsencrypt-s3front:installer \
--letsencrypt-s3front:installer-cf-distribution-id E146MJP5GJ9PPL \
-d dl.servo.org
letsencrypt --agree-tos -a letsencrypt-s3front:auth \
--letsencrypt-s3front:auth-s3-bucket servo-builds \
--letsencrypt-s3front:auth-s3-region us-west-2 \
-i letsencrypt-s3front:installer \
--letsencrypt-s3front:installer-cf-distribution-id E146MJP5GJ9PPL \
-d download.servo.org
```
- [ ] Use letsencrypt to create and renew cert for buildbot
- [ ] After all of the above are definitely working correctly, [flip the HSTS switch in Cloudfront](https://blog.cloudflare.com/enforce-web-policy-with-hypertext-strict-transport-security-hsts/) so that browsers will refuse to ever load *.servo.org pages over HTTP

aneeshusa:
@edunham The Cloudfront certs will expire in two weeks, we should renew them.

aneeshusa:
@edunham @larsbergstrom The certs have expired, please renew them manually ASAP, as I don't have time to investigate Salting this at the moment.

aneeshusa:
@frewsxcv mentioned using [Amazon ACM](https://aws.amazon.com/certificate-manager), which also auto-renews, is free, and has CloudFront integration, and I think is another good option for CloudFront sites like download.servo.org (no need for cron!).

[...]

PR #454, opened 2016-08-10
Check Travis build status via GitHub instead of Travis hooks

larsbergstrom:
r? @Manishearth 

Fixes https://github.com/servo/servo/issues/12739

cc: @metajack @edunham

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/454)

PR #906, opened 2018-10-31
Switch to https://build.servo.org

Darkspirit:
@jdm 
I don't want to just ping you, so I make this pull request.
This should help with https://github.com/servo/saltfs/issues/423#issuecomment-429399125, but is not tested with Vagrant.

- A next step would be to change it to a permanent http-to-https redirection and to activate HSTS.
- Current TLS config is [bad](https://www.hardenize.com/report/build.servo.org/1541002698#www_tls), therefore only modern ciphersuites should be allowed.
- http://build.servo.org:54856/travis might already not exist anymore?
- Please feel free to edit this pull request.

(Background: servo.org enforces [HSTS](https://www.hardenize.com/report/servo.org/1540988336#www_hsts) since [bug 1482947](https://bugzilla.mozilla.org/show_bug.cgi?id=1482947) and is ready to be HSTS-preloaded.)

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/906)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,6 +98,6 @@ matrix:
 script: .travis/dispatch.sh
 
 notifications:
-  webhooks: http://build.servo.org:54856/travis
+  webhooks: https://build.servo.org/homu/travis
   irc: "irc.mozilla.org#servo-bots"
 
--- a/buildbot/master/files/config/master.cfg
+++ b/buildbot/master/files/config/master.cfg
@@ -232,7 +232,7 @@ c['builders'] = [
 
 c['status'] = [
     status_push.HttpStatusPush(
-        serverUrl='http://build.servo.org:54856/buildbot',
+        serverUrl='https://build.servo.org/homu/buildbot',
         extra_post_params={'secret': HOMU_BUILDBOT_SECRET},
     ),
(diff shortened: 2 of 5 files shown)


Darkspirit:
@jdm
There is a [long list of travis webhooks](https://github.com/search?q=org%3Aservo+%22http%3A%2F%2Fbuild.servo.org%22&type=Code) trying to connect to http://build.servo.org:54856/travis which is already 404 Not Found when accessing with curl.
Since #454 Homu is checking build statuses directly at GitHub
https://github.com/servo/saltfs/blob/5dcb05a3d75430f3f8a2d78c4485bf0c2494059c/homu/files/cfg.toml#L268-L271 and doesn't rely on Travis webhook notifications, if I'm not wrong: https://github.com/servo/saltfs/blob/5dcb05a3d75430f3f8a2d78c4485bf0c2494059c/homu/files/cfg.toml#L100

Shall I create PRs removing each travis webhook? Or do you want to ignore them?

jdm:
Removing those unneeded webhook notifications would be useful.

Darkspirit:
I'll split this to multiple PRs as well.

PR #908, opened 2018-11-05
Add magicleap sdk to mac build machines

asajeffrey:
Download the magicleap SDK from s3 on the mac build machines.

PR #910, opened 2018-11-05
Add magicleap builders

asajeffrey:
Add magicleap builders to homu and buildbot.


asajeffrey:
Depends on https://github.com/servo/servo/pull/22117 and https://github.com/servo/saltfs/pull/908.

PR #913, opened 2018-11-07
Improve TLS config

Darkspirit:
https://github.com/servo/saltfs/pull/906#issuecomment-436313032

Problem:
https://www.hardenize.com/report/build.servo.org#www_tls
https://www.ssllabs.com/ssltest/analyze.html?d=build.servo.org&hideResults=on

* Let's first switch all clients to https before we disable TLS 1.0. Just to be sure.
* [`server_tokens off;`](https://nginx.org/en/docs/http/ngx_http_core_module.html#server_tokens) makes nginx return `Server: nginx` without any version number.
* ssl_session_cache: [official recommendation](https://nginx.org/en/docs/http/configuring_https_servers.html#optimization), [docs](https://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_session_cache)

An nginx restart is needed: [`service nginx restart`](https://github.com/servo/servo/wiki/Buildbot-administration)
You can test with `nginx -t` before restarting.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/913)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,6 +98,5 @@ matrix:
 script: .travis/dispatch.sh
 
 notifications:
-  webhooks: http://build.servo.org:54856/travis
   irc: "irc.mozilla.org#servo-bots"
 
--- a/nginx/default
+++ b/nginx/default
@@ -1,15 +1,18 @@
 server {
   listen 80 default_server;
+  listen 443 ssl;
   server_name build.servo.org;
+  server_tokens off;
+  ssl_protocols TLSv1.2 TLSv1;
+  ssl_ciphers ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-SHA;
+  ssl_prefer_server_ciphers on;
+  ssl_session_cache shared:SSL:10m;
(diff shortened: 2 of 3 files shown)


Darkspirit:
> One thing to be aware of - I changed a webhook target for servo/servo from http://build.servo.org:9010 to https://build.servo.org:9010 and it broke everything.

Oh dear. Of course, there is no starttls on an http port.
I created another commit to introduce https://build.servo.org/github-buildbot/.
Is this folder name okay or should we change it?

In the future, when all those plaintext services are behind nginx, they can be changed to only listen on 127.0.0.1. For github_buildbot.py it would be
* either blocking port 9010 with iptables for the public
* or changing
https://github.com/servo/saltfs/blob/90036502c8fac2291fba372b3309bafef605f5b2/buildbot/master/files/github_buildbot.py#L306
to `reactor.listenTCP(options.port, site, interface='127.0.0.1')` ([source 1](https://www.twistedmatrix.com/pipermail/twisted-commits/2003-January/005534.html), [source 2](https://www.programcreek.com/python/example/64153/twisted.internet.reactor.listenTCP)).

Darkspirit:
@jdm Should further changes be made, e.g. would you like to have a different name for the github_buildbot.py webhook, or is this ready to be merged? Thanks

PR #928, opened 2018-12-17
Give mandreyel try access.

emilio:
They're implementing parallel CSS parsing (see https://github.com/servo/servo/pull/22478), and I bet this will be useful!

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/928)


emilio:
r? @jdm or anyone else?

I guess the description should point to https://github.com/servo/servo/pull/22478

mandreyel:
Yaaaaay :)

PR #949, opened 2019-03-27
Add NTP service that relies on AWS NTP servers to avoid clock drift.

jdm:
I have done this manually on the linux cross builders which were exhibiting the same problem previously, and it went away. Based on https://www.allcloud.io/how-to/how-to-fix-amazon-s3-requesttimetooskewed/.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/949)


jdm:
servo-linux2 is >15m ahead of the AWS servers, for example.

PR #954, opened 2019-04-05
Add jq as required util for upstream wpt sync

CYBAI:
servo/servo#23162 should be merged after this one

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/954)

### DIFF
--- a/upstream-wpt-webhook/init.sls
+++ b/upstream-wpt-webhook/init.sls
@@ -26,6 +26,7 @@ upstream-wpt-webhook:
   pkg.installed:
     - pkgs:
       - patchutils
+      - jq
   pip.installed:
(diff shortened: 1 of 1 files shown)


jdm:
Ah, shoot. I just remembered that this change will only affect servo-master1, which runs the webhook. We actually want to change https://github.com/servo/saltfs/blob/c7007645cd58edc2c53fc4143195530d8b3b3bf9/servo-build-dependencies/init.sls#L26-L42 which controls the packages installed on our mac build machines that actually run the nightly script.

CYBAI:
Oops, let me send another PR to remove this one and add `jq` to the above block!

PR #973, opened 2019-07-31
Redirect http://build.servo.org/ to HTTPS

SimonSapin:
This is untested, but matches what https://ssl-config.mozilla.org/ suggests. (This could be a good time to apply other suggestions from there too.)

### DIFF
--- a/nginx/default
+++ b/nginx/default
@@ -1,10 +1,14 @@
 server {
   listen 80 default_server;
+  return 301 https://$host$request_uri;
+}
+
+server {
+  listen 443 ssl;
   server_name build.servo.org;
+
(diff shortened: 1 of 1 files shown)


Darkspirit:
Looks good, please merge! ^^ #913 was ignored, so I closed it in a moment of weakness. xD
https://www.hardenize.com/report/build.servo.org/1564587122#summary

jdm:
iirc last time I made this change I broke our buildbot CI completely. This will need to wait until I see convincing evidence that that won't happen again.

jdm:
In particular, I believe that there are some python-based HTTP requests for servo.org that see a 301 response and raise an exception.

PR #977, opened 2019-08-16
Move jrmuizel from 'extra_reviewers' into 'reviewers'

jrmuizel:
I'm not sure what the process is for becoming a reviewer but not having this has been a hassle enough times and I'm an extra_reviewer for enough things that's it's probably not too controversial.

---
This change is [Reviewable](https://reviewable.io/reviews/servo/saltfs/977)


jdm:
I am inclined to agree.
@bors-servo r+

PR #983, opened 2019-11-05
Gate servo/servo PRs on the `community-tc` deployment of Taskcluster

SimonSapin:
… in addition to the legacy deployment, for now. That one will be removed after the transition is finished.

CC https://bugzilla.mozilla.org/show_bug.cgi?id=1574648

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -278,6 +278,9 @@ context = 'Taskcluster (push)'
 
 # Specialized per-repo configs
 
+[repo.servo.status.community-tc]
+context = 'Community-TC (push)'
+
 [repo.glutin.branch]
 master = "servo"
 


SimonSapin:
This should only be deployed after https://github.com/servo/servo/pull/24662 lands.

SimonSapin:
@jdm Could you help with deploying this? https://github.com/servo/servo/pull/24662 has landed. I ran Salt with `test=True` on `servo-master` but didn’t dare do more since it showed a number of unrelated changes.

SimonSapin:
Alright, after filtering it down (`salt 'servo-master1' state.apply homu`) and enabling `/tmp/salt-testing-root` this looked better. The only unexpected change was:

```
Python package git+https://github.com/servo/homu@78dbd0be5dc1739787eaa2b9df858d736aa7aa4b is set to be installed
```

… which is hopefully harmless enough. It is still marked as to be done when running again with `test=True`, so I assume that Salt doesn’t realize this is already up to date.

PR #984, opened 2019-11-07
Add community TC integration for webrender.

jdm:


SimonSapin:
Note that this makes Homu wait for a Status from *both* TC deployments. It’ll still be stuck if tasks are not running on `taskcluster.net` anymore.

PR #986, opened 2019-11-07
Stop gating servo/servo on taskcluster.net (only Community-TC)

SimonSapin:


SimonSapin:
This should only be deployed after https://github.com/servo/servo/pull/24689 lands.

PR #990, opened 2019-12-09
Add marmeladema to try users.

jdm:
They're doing useful work on upgrading from python2->python3 in https://github.com/servo/servo/pull/24575 and that ends up requiring a bunch of tinkering with CI, unsurprisingly.


jdm:
This has been deployed - @marmeladema should now be able to run homu try commands.

PR #999, opened 2020-04-15
Rename `saneyuki` to `tetsuharuohzeki` in homu's reviewer list.

tetsuharuohzeki:
I renamed my github id to new one and update it to keep make secure this list.

I think I might have a chance to contribute servo for the future but I'm not sure about it. If you feel it's better to drop my id once from here, I'll change this patch.


tetsuharuohzeki:
@jdm Thank you!

PR #1001, opened 2020-05-13
Add font-kit and pathfinder to homu.

jdm:


jdm:
@bors-servo r+
Deployed.

PR #1002, opened 2020-08-02
Add kunalmohan to try users

kunalmohan:
I will be working on WebGPU CTS and having "try" permissions will be quite helpful.

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -218,6 +218,7 @@ secret = "{{ secrets['web-secret'] }}"
     "izgzhen",
     "jonleighton",
     "jyc",
+    "kunalmohan",
     "magni-",
     "mandreyel",
     "marmeladema",


jdm:
Filed https://github.com/servo/saltfs/issues/1003 about the CI failures.

jdm:
Try changing https://github.com/servo/saltfs/blob/90036502c8fac2291fba372b3309bafef605f5b2/salt/common.sls#L6-L8 as well.

kunalmohan:
Some success, but still 2 failures

kunalmohan:
I fail to understand why some tasks return `gpg: no valid OpenPGP data found.` and others `Ok`

jdm:
Ok, I figured out why those two jobs are failing - the URLs for the archived salt versions don't include the patch release. I'm going to try upgrading us to a non-archived version instead in #1004.

jdm:
This change has been deployed; you should be able to use try commands now.

PR #1005, opened 2020-08-05
Add crash reporter service

paulrouget:
Flask: https://github.com/paulrouget/crash-reporter - this will need a review, and be migrated under github.com/servo. Tested and appears to work as expected.

Saltfs configuration is blindly copy/pasted from the intermittent-tracker code.

@jdm can you give a quick a look at the flask code? And do we have any mechanism in place to prevent abuse of the different build.servo.org services?


jdm:
In general our build.servo.org services rely on secrets that need to be provided as part of the submission process (eg: https://github.com/servo/standups/blob/415c2fe2a399575740cd9ce585eeff0b11fafec1/standups/flask_server.py#L90). The crash reporter would need to embed the secret in the binary, which is unavoidable, but it's Good Enough for casual attempts to abuse the server.

crash-reporter/init.sls
```diff
@@ -0,0 +1,55 @@
...
+      - pkg: python3
+      - pip: virtualenv
+  pip.installed:
+    - pkgs:
+      - git+https://github.com/servo/crash-reporter@{{ tracker.rev }}
+    - bin_env: /home/servo/crash-reporter/_venv
+    - upgrade: True
+    - require:
+      - virtualenv: crash-reporter
+  {% if grains.get('virtual_subtype', '') != 'Docker' %}
```
jdm, line 24: This conditional (and the corresponding endif) can be removed like in https://github.com/servo/saltfs/pull/1006.

crash-reporter/init.sls
```diff
@@ -0,0 +1,55 @@
...
+      - pip: crash-reporter
+    - watch:
+      - file: /home/servo/crash-reporter/config.json
+      - file: /lib/systemd/system/tracker.service
+      - pip: crash-reporter
+  {% endif %}
+
+/home/servo/crash-reporter/config.json:
+  file.managed:
+    - source: salt://{{ tpldir }}/files/config.json
```
jdm, line 38: For the shared secret, we will want to add:
```
- context:
    secret: {{ pillars['crash-reporter']['secret'] }}
```
We will also need to add /srv/pillar/crash-reporter.sls on the main machine, and add the following fake crash-reporter.sls to .travis/test_pillars/:
```
'crash-reporter':
  'secret': 'TEST-CRASH-REPORTER-SECRET'
```

crash-reporter/files/config.json
```diff
@@ -0,0 +1 @@
+{"port": 5004, "crash_dir": "./crashes/"}
```
jdm, line 1: This will need a `"secret": "{{ pillar['crash-reporter']['secret'] }}"` entry as well.

crash-reporter/init.sls
```diff
@@ -0,0 +1,55 @@
...
+    - template: jinja
+    - user: servo
+    - group: servo
+    - mode: 644
+
+/lib/systemd/system/tracker.service:
+  file.managed:
+    - source: salt://{{ tpldir }}/files/tracker.service
```
jdm, line 46: We'll need a different name for this file or it will overwrite the existing tracker.service. Let's call it `crash-reporter.service` instead.

crash-reporter/init.sls
```diff
@@ -0,0 +1,55 @@
...
+      - virtualenv: crash-reporter
+  {% if grains.get('virtual_subtype', '') != 'Docker' %}
+  service.running:
+    - enable: True
+    - name: tracker
+    - require:
+      - pip: crash-reporter
+    - watch:
+      - file: /home/servo/crash-reporter/config.json
+      - file: /lib/systemd/system/tracker.service
```
jdm, line 32: This will need to change if we change the .service file name.

paulrouget:
Note to self: [this](https://github.com/servo/saltfs/pull/1005/files#r467124883) hasn't been addressed yet,

PR #1007, opened 2020-10-30
Add mstange to reviewers.

mstange:
I would like to be able to review changes to the various macOS and OpenGL crates that WebRender depends on.

r? @jdm

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -38,7 +38,9 @@ secret = "{{ secrets['web-secret'] }}"
     "app_units": {},
     "cgl-rs": {},
     "cocoa-rs": {},
-    "core-foundation-rs": {},
+    "core-foundation-rs": {
+        "extra_reviewers": [ "mstange" ],
+    },
     "core-graphics-rs": {},
     "core-text-rs": {},
     "devices": {},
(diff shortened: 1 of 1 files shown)


jdm:
@mstange Is that https://github.com/servo/core-foundation-rs/ and https://github.com/servo/gleam/, or are there any others? It would be cleaner to add you to the reviewers list for the specific crates.

mstange:
Yes, those two, and core-graphics-rs + core-text-rs.

jdm:
Ok, if you could:
* remove your name from the global reviewers list
* add a `extra_reviewers` list with your name in https://github.com/servo/saltfs/blob/da8e2c6d9f020e33d0fa783a70f65f60ffef7b1e/homu/files/cfg.toml#L41 and https://github.com/servo/saltfs/blob/da8e2c6d9f020e33d0fa783a70f65f60ffef7b1e/homu/files/cfg.toml#L60
then we should be ok to merge.

jdm:
(homu only cares about github repositories, and all the macOS crates live in a workspace in the core-foundation-rs repo now)

mstange:
Ah, I see. I'll make that change.

mstange:
I've updated the PR.

PR #1033, opened 2023-01-05
Add try users @delan + @Loirooriol + @mrego + @mrobinson

delan:
It looks like being a reviewer [doesn’t imply](https://github.com/servo/servo/pull/29201#issuecomment-1371582879) being a try user.


mrego (approved):

jdm:
Reviewer does imply try user so I suspect the change wasn't actually deployed.

PR #1045, opened 2023-05-19
Slim down the list of users with permissions

mrego:
Remove inactive users from the list of users with permissions.


mrego:
I guess it makes sense to reduce the list of users here.

Maybe I'm removing too many people here, please let me know if I removed someone that I shouldn't by mistake.

mrobinson (approved):

homu/files/cfg.toml
```diff
@@ -220,104 +220,36 @@ secret = "{{ secrets['web-secret'] }}"
...
     "jdm",
-    "jgraham",
     "jrmuizel",
-    "KiChjang",
-    "kmcallister",
     "kvark",
-    "larsbergstrom",
     "Loirooriol",
     "Manishearth",
-    "mbrubeck",
```
jdm, line 248: mbrubeck approves PRs in https://github.com/servo/rust-smallvec/

jdm (approved):

PR #1048, opened 2023-07-11
Re-add additional reviewers for rust-url

lucacasonato:
Hey, during the [recent Servo TSC Github permission cleanup](https://github.com/servo/rust-url/pull/839#issuecomment-1571352449), @valenting and I lost access to `rust-url` to do reviews. We have been maintaining it for the past year and a bit, and it would be great to have review access again. Is this the right way?

### DIFF
--- a/homu/files/cfg.toml
+++ b/homu/files/cfg.toml
@@ -169,7 +169,7 @@ secret = "{{ secrets['web-secret'] }}"
     },
     "rust-stb-image": {},
     "rust-url": {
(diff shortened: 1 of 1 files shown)


lucacasonato:
My main issue is the lack of my ability to do "Request reviews" now.

lucacasonato (commented):
I can't tell whether this is actually needed.

lucacasonato:
Great, thanks

PR #1050, opened 2023-07-25
Don't run the upstream-wpt-webhook service

mrobinson:


mukilan:
Will we be removing the remaining references to  upstream-wpt-webhook in a follow up PR after we deploy this change to disable it and validate it?

PR #1054, opened 2024-08-04
Increase client_max_body_size for intermittent-tracker

### LINKED ISSUE #31845: Handle HTTP Error 413: Request Entity Too Large more gracefully.
WPT import currently fails because of it: https://github.com/servo/servo/actions/runs/8405500156/job/23018617126#step:8:14419 and I've hit this when doing webgpu tests too. Ideally if too large we should split it into multiple request.

sagudev:
Should fix https://github.com/servo/servo/issues/31845, but it might be still to low.

### DIFF
--- a/nginx/default
+++ b/nginx/default
@@ -17,6 +17,7 @@ server {
     proxy_pass http://localhost:54857/;
   }
   location /intermittent-tracker/ {
+    client_max_body_size 10M; # For https://github.com/servo/servo/issues/31845, default value is 1M
     proxy_pass http://localhost:5000/;
   }
   location /upstream-wpt-webhook/ {


mrobinson (approved):

mrobinson (approved):

sagudev:
Is this deployed, so we can test?

mrobinson:
I believe this is deployed now, but I'm always a bit mystified by saltfs, so let me know if there are any issues.

sagudev:
I manually tested with curl, and it seems that it is deployed.

Test WPT import: https://github.com/servo/servo/actions/runs/10281279358

sagudev:
It worked: https://github.com/servo/servo/pull/32959 🎉

